PR #27, opened 2019-09-13
Ensure table name length is consistent across CPUs

### LINKED ISSUE #25: TBL services using platform-scope values in table header definition
**Describe the bug**
The `CFE_TBL_MAX_FULL_NAME_LEN` size definition, which is used as part of the `CFE_TBL_FileDef_t` structure, is partly based on the `OS_MAX_API_NAME` value.

The problem is that this value is platform scope and thereby configurable on a per-cpu basis.  Therefore, the `OS_MAX_API_NAME` value can be (validly) different on each CPU in a multi-cpu deployment.  This, in turn, means the table definition headers may be different on each CPU.

Tools like `elf2cfetbl` and `tblCRCTool` do not account for this possibility; when reading or writing the table header, they use a single definition.  The definition that they use may or may not even match what the flight code uses at all, depending on how the mission is configured.

**To Reproduce**
Configure a mission with two CPUs, and specify a different osconfig.h for each one (i.e. set TGT2_PLATFORM to something other than default).  Change the value of `OS_MAX_API_NAME` for the second CPU, and build all software.

Observe that the `sizeof(CFE_TBL_FileDef_t)` is now also different on CPU2 vs. CPU1.  Tables generated by `elf2cfetbl` and the CRC computed by `tblCRCTool` only work in CPU1.  CPU2 will not be able to load tables because of the incompatible header.

**Expected behavior**
The `sizeof(CFE_TBL_FileDef_t)` should remain consistent regardless of the platform-scope values used, because this is an external format shared between (at least) between ground systems and flight code.

The `CFE_MISSION_MAX_API_LEN` macro is a better choice here.  It was introduced in a previous version of CFE for telemetry packets, but table files are effectively the same issue.

**Code snips**
The problem definition is:
```
#define CFE_TBL_MAX_FULL_NAME_LEN_COMP (CFE_MISSION_TBL_MAX_NAME_LENGTH + OS_MAX_API_LEN + 2)
```

Furthermore, as this definition is only really for the header struct, it should be in cfe_tbl_filedef.h, not cfe_tbl.h.  This allows external tools to include this definition without having to pull in the rest of CFE headers implicitly.

**System observed on:**
Ubuntu 18.04 (64-bit), kernel 5.0.0-23-generic

**Reporter Info**

jphickey:
**Describe the contribution**
To ensure consistency in the size of the CFE_TBL_FileHdr_t struct, this should be constrained to use only mission-scope definitions.
    
Also simplifies the structure definitions by _NOT_ padding strings out to 32 bit multiples.  The default size of the strings are already 32 bit multiples so this is just unnecessary complexity.  There is also no major issue if not 32 bit aligned, as the compiler will add it automatically where needed.

Add a note in the related cfe_mission_cfg.h descriptions to affected values, that these should be kept as a multiple of 4 to maintain alignment.
    
Fixes #25

**Testing performed**
Built CFE using default/sample config
Verified no build issues, CFE and unit tests all run as expected (no change).

**Expected behavior changes**
No impact to behavior, but this may change the size of telemetry packets in some cases depending on the mission config.

**System(s) tested on:**
 Ubuntu 18.04 (64 bit)

**Contributor Info**
Full name and company/organization of all contributors (required for acceptance)

**Community contributors**

### DIFF
--- a/cmake/sample_defs/sample_mission_cfg.h
+++ b/cmake/sample_defs/sample_mission_cfg.h
@@ -300,7 +300,9 @@
 **
 **
 **  \par Limits
-**      Not Applicable
+**       This value should be kept as a multiple of 4, to maintain alignment of
+**       any possible neighboring fields without implicit padding.
+**
 */
 #define CFE_MISSION_ES_CDS_MAX_NAME_LENGTH         16
 
@@ -310,7 +312,7 @@
 **  \cfeevscfg Maximum Event Message Length
 **
 **  \par Description:
-**      Indicates the maximum length (in characers) of the formatted text
+**      Indicates the maximum length (in characters) of the formatted text
 **      string portion of an event message
 **
 **  \par Limits
@@ -349,7 +351,8 @@
 **      form: "ApplicationName.TblName"
 **
 **  \par Limits
-**      Not Applicable
+**       This value should be kept as a multiple of 4, to maintain alignment of
+**       any possible neighboring fields without implicit padding.
 */
 #define CFE_MISSION_TBL_MAX_NAME_LENGTH         16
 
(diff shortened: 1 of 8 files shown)


skliper:
Switched base branch to rc-6.7.0 until it gets merged so all those changes don't clutter this change.

jphickey:
Commit 8fe60c3 contains the requested modifications

Issue #360, opened 2019-09-30
Application List in 'sample_defs'


skliper:
**Is your feature request related to a problem? Please describe.**
The `CMakeLists.txt` file in each application that utilizes tables (HK, SCH, TO, etc.) requires mission specific applications to be added to the `include_directories` statement.

**Describe the solution you'd like**
I suggest we use a single `APPLICATION_PLATFORM_INC_LIST` variable declared in `sample_defs/targets.cmake` that can be used for multiple target app lists and in the applications as well so that the mission specific code and tables remain separate.

**Describe alternatives you've considered**
None

**Additional context**
None

**Requester Info**
jplucas from babelfish

skliper:
Was ticket 329 on babelfish

skliper:
Not clear what problem this is solving.  Re-scoping app include directories to a top level complicates app portability model.  The goal is to not have to change top level settings based on app changes (which is why the include list is specified in the app CMakeList.txt file).

skliper:
Got it.  In the framework it relates to:

https://github.com/nasa/sch_lab/blob/e226d55d76621a8247286ba150e20e2edd51111c/CMakeLists.txt#L6-L8

Where this list could be passed in from targets.cmake.  Likely a mission scope list to cover all the apps for all the targets (the app lists could be different), such that each target could have it's own table built without having to change the SCH (or whatever) CMakeList.txt file.

We will likely discuss this in the CCB today and come back with a response.

skliper:
See solution suggested in #365 comment thread.

PR #365, opened 2019-10-02
Update `cmake/sample_defs/targets.cmake` to include APPLICATION_LIST;

jlucas9:
**Describe the contribution**
Solves issue #360.

**Expected behavior changes**
Enables missions to slowly roll out new applications without updating include lists in each table provided application each time a new one is included.

**Contributor Info**
John P. Lucas - NASA GSFC

### DIFF
--- a/cmake/sample_defs/targets.cmake
+++ b/cmake/sample_defs/targets.cmake
@@ -83,18 +83,29 @@ SET(UI_INSTALL_SUBDIR "host/ui")
 # be copied during the install process.
 SET(FT_INSTALL_SUBDIR "host/functional-test")
 
-# Each target board can have its own HW arch selection and set of included apps
+# APPLICATION_LIST defines a set of baseline applications to be used across architectures
+SET(APPLICATION_LIST sample_app ci_lab to_lab sch_lab) 
+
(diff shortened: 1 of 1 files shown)


skliper:
A goal is to reduce user modifications required within the open source apps, especially when it's just another app that is being added or removed.  The SCH table is easily managed at the mission level, so it sounds like an approach to manage the include list (or paths to elf2cfetbl, or whatever mechanism you'd like to propose for consideration) at the mission level or automate based on the included apps supports this goal.  

My guess is it's just the table build step (prior to elf2cfetbl conversion) that requires it.

skliper:
For SCH_Lab, all the table build actually needs is the MID's.  If these were available/managed at a mission level all the app include directories wouldn't be needed.

skliper:
CCB - 2019/10/09: could update table building rule.  Or manage MID's at mission level (as defined in the app they are just suggestions).  Would be helpful to update related documentation.

Brought up command codes as potentially another concept to address, but these probably aren't needed (not used by scheduler).

skliper:
Pull request not accepted for incorporation into framework.  Open source framework preference is to manage MID's at the mission level (*_defs/*_msgids.h).

Issue #291, opened 2019-09-30
FS - externalize cfe_fs_decompress


skliper:
The decompression code should be removed and compile-time hooks to connect in an external decompression library should be provided.

skliper:
_Imported from trac issue 260.  Created by cdknight on 2018-12-11T16:57:03, last modified: 2019-07-03T12:32:27_

skliper:
_Trac comment by cdknight on 2018-12-11 16:58:45:_

Much discussion on the mailing list early December 2018, initiated by a user finding a bug in the decompression code. My recommendation is as the ticket describes, externalize the code.

One suggestion is to use the "ExternalProject" capability in the latest CMake:

https://github.com/abseil/googletest/blob/master/googletest/README.md#incorporating-into-an-existing-cmake-project

skliper:
_Trac comment by jhageman on 2019-07-03 12:25:14:_

From @acudmore:

LZMA compression SDK:
https://www.7-zip.org/sdk.html

We used this on MMS/RTEMS, and Dellingr/FreeRTOS for the elf loader we created. The development host tool optionally compresses the code and data segments, then the flight loader decompresses the segments. 
On Dellingr, the original ACS static load file (.slf) was 103Kbytes. Using the lzma compression it was 19Kbytes. This was key for our ability to upload new apps on Dellingr. 

The embedded decompress code consists of a single header and single C file. The license is public domain and should be compatible with Apache 2.0.

skliper:
_Trac comment by jhageman on 2019-07-03 12:32:27:_

Or zlib or similar.

Preference from cFE would be to provide an interface where one or more of these open source solutions could be easily plugged in.  I suggest not building anything in by default.

skliper:
Externalization question @acudmore @CDKnightNASA @jphickey:

There is functionality that doesn't link to any requirements within ES, where an app file or library file that ends in ".gz" is uncompressed before loading that depends on this code.  In context of externalizing decompress, what approach should we take?  Do we take out this functionality in ES so that decompress can be moved to a standard cFS library (distributions could still 2-step loading apps that are compressed using FM to decompress and then ES to load, but the core set would need to be uncompressed)?  Any other externalization approach I can think of just leads to more complexity, and/or configurations that will fail (adding an unimplemented "module" option will just fail, so it seems like the core needs to include at least one functional decompress if the unrequired capability needs to become fully supported), or dependency on non-flight libraries.  Definitely open to suggestions.

[...]

PR #367, opened 2019-10-07
Fix #298, fix infinite loop when gz file is truncated

### LINKED ISSUE #298: CFE_FS_Decompress infinite loop on truncated gzip file
If CFS_FS_Decompress is given a truncated gzip file, it will enter an infinite loop in which it attempts to read more data from the file, gets nothing, and tries again. Discovered by accidentally attempting to have ES load a new compressed application (which we didn't know had been truncated), which led to ES getting stuck and an eventual watchdog reset.

The problem seems to be that running out of bytes in a gzipped file before decompression is finished is not considered an error. I can see how this might be intentional if it is expected that the file handle might be a stream that could present data after being emptied, but for the normal file use case I think it is a bug. Suggested fix by combining the two checks at line 309 of cfe_fs_decompress.c:

if ( State->insize == 0 ) return EOF;

if ( len == OS_FS_ERROR ) 
{
    State->Error = CFE_FS_GZIP_READ_ERROR;
    return EOF;
}

into one: 

if ((State->insize == 0) !|| (len == OS_FS_ERROR)) {
    State->Error = CFE_FS_GZIP_READ_ERROR;
    return EOF;
}

Recommend investigation if insize might temporarily hit 0 during a normal decompression.

Reported via email from Mike Stewart, mike@capellaspace.com

avan989:
**Describe the contribution**
Fix issue #298, when cfe_fs_decompress goes into infinite loop when gz file is truncated. 

**Testing performed**
1. Build 
2. Pass in truncated file as parameter to cfe_fs_decompress.
[corgi2.jpg.gz](https://github.com/nasa/cFE/files/3698606/corgi2.jpg.gz)
3. Verify the return status:

```
STATUS CODE = -973078518
```

4. Gzip good file
66239143-f6736880-e6ad-11e9-98ac-ce321d6d495a

5. Pass as parameter and verify output. 

**System(s) tested on:**
 - Hardware
 - Ubuntu 18.04.03
 - cFE 6.7.0

**Contributor Info**
Anh Van, NASA Goddard

**Community contributors**

### DIFF
--- a/fsw/cfe-core/src/fs/cfe_fs_decompress.c
+++ b/fsw/cfe-core/src/fs/cfe_fs_decompress.c
@@ -284,16 +284,14 @@ int16 FS_gz_fill_inbuf_Reentrant( CFE_FS_Decompress_State_t *State )
    {
 		len = OS_read( State->srcFile_fd, (int8*)State->inbuf + State->insize, INBUFSIZ - State->insize );
 		
-		if ( len == 0 || len == EOF || len == OS_FS_ERROR ) break;
+		if ( len <= 0 ) break;
 		
 		State->insize += len;
 		
 	} while ( State->insize < INBUFSIZ );
 
 
-	if ( State->insize == 0 ) return EOF;
-
-	if ( len == OS_FS_ERROR ) 
+	if ( State->insize == 0 || len < 0) 
    {
 		State->Error = CFE_FS_GZIP_READ_ERROR;
 		return EOF;


jphickey:
Has this been confirmed to correctly handle a **NON**-truncated file?

jphickey:
Another general comment, if we are touching this code I would _like_ to see a couple more changes.

The checks of `len` in the function should all be simplified to be just:

At line 287:
```
if (len <= 0) break;
```

And At line 294:
```
if (State->insize == 0 || len < 0) 
...
```

Rather than specifically testing for `OS_ERROR` only.  Any negative value constitutes an error, and it will corrupt the state if not caught (i.e. insize could go negative).

Also note that the `EOF` macro is not defined by OSAL nor defined anywhere else in CFE.  It just happens to not trigger a compiler error because it must be getting the value implicitly via `stdio.h` , and probably has the value -1.  It might be clearer/better to just use -1 here.

skliper:
Preference is to avoid "magic numbers" like -1.  Is there an OSAL/CFE define that is appropriate?

jphickey:
> Preference is to avoid "magic numbers" like -1. Is there an OSAL/CFE define that is appropriate?

Not particularly.  The function is supposed to return the first character in the buffer, but in this case the buffer is empty, so there is nothing to return.

Another reasonable option might be the NUL char (0).  But that just as magic as -1.  Returning 0 won't corrupt the shift register that the values go into, whereas -1 does.

Either way though, this decompress code still violates coding standards in every conceivable way.  I only mention it because it might cause a compile failure if things change and EOF no longer happens to be defined.  Same with the length check (testing <= 0 vs. testing a specific value).

The real fix for this should be done in #291

skliper:
> Either way though, this decompress code still violates coding standards in every conceivable way. I only mention it because it might cause a compile failure if things change and EOF no longer happens to be defined. Same with the length check (testing <= 0 vs. testing a specific value).

Understood.  Let's fix the <=0 now.  The EOF builds today and the "-1" solution isn't acceptable, so lets just leave as is for now.  Or if you really want to, feel free to propose a solution that doesn't include the use of a "magic number" (create an appropriate #define, identify an appropriate one that already exists, or whatever).  But I don't think the EOF change is worth our time right now due to #291.

> The real fix for this should be done in #291

Agreed, which is why I'm OK with leaving the EOF as-is, but I'd like to get this fix in first.

skliper:
Since it's all local to cfe_fs_decompress.c, why not just:

#define CFE_FS_EOF -1

or something? and change EOF to CFE_FS_EOF?

PR #369, opened 2019-10-08
Fix #199, Update obj creation failure requirement

### LINKED ISSUE #199: cFE cES1515.1 Requirement Failure
Requirements cES1515 and cES1515.1 state:

"cES1515 Upon a Processor Reset, the cFE shall create all operating system objects required by the
cFE."

"cES1515.1 If the creation of the operating system object fails, the cFE shall perform a power on
reset." 

To satisify this requirement, the CFE_ES_CreateObjects function makes a call to the CFE_PSP_Panic function in the case where the return from OS_TaskCreate != OS_SUCCESS.  The CFE_PSP_Panic function however, does not perform a power on reset.  It performs an exit(-1).  To correctly satisfy this requirement, ES needs to make a call to CFE_PSP_Restart with an input argument of CFE_PSP_RST_TYPE_POWERON to specify a power on reset.  This may not be the best solution.  It is recommended to reconsider this requirement.  It may be the best/required course of action to perform a power on reset following a processor reset if the creation of one of the cFE core application objects fails.  This may not be the best/required course of action if the creation of one of the applications listed in the ES startup script fails.

It should be noted that this requirement has not been properly handled in past releases of the cFE.  The call to the CFE_PSP_Panic function to satisfy this requirement has been in place since the initial release of the cFE.  It was recently found that the test to verify this requirement was faulty producing a false positive.

skliper:
**Describe the contribution**
Fixes #199, updates obj creation failure requirement to platform defined response

**Testing performed**
None - document update

**Expected behavior changes**
None

**System(s) tested on:**
None

**Additional context**
Platform defined response now matches cFS PSP model.

Note - Requirements transitioned to new tracking system with minor typo fixes throughout and new output format which changes every line.  "Version tracking" is really accomplished in the requirements management tool, not the *.md file.

**Contributor Info**
Jacob Hageman - NASA/GSFC


skliper:
It was the pandoc gfm (github flavored markdown) setting.

skliper:
I tried the standard markdown (--to markdown) and it didn't render the tables in a readable format on GitHub.  Happy to supply it in whatever format people want, it's just an output from the tool and pandoc conversions.

jphickey:
The version I see now is loaded with HTML tags, it is not markdown at all.  

Markdown is definitely preferred as it keeps things easily readable and editable with a standard text editor.  Github display on the website is a secondary goal -- most people won't look at it that way.  Rather, they will clone the git repo and open their local copy in an editor.

That being said, I believe there is a way to do tables in github-flavored markdown, even if an automatic conversion isn't possible (can always write a one-off perl script to fix it).

Even if not, I would rather see it as a simple (non-tabular) list but keep it as plain text.  Right now I can't even see what the actual change was in this pull request, because the entire file is converted to HTML.

skliper:
Uploaded for consideration - *.csv (directly from tool), *.html (from tool with branding removed), *.gfm (*.html converted with pandoc -gfm), *.md (*.html converted with pandoc -markdown).

skliper:
Updated per CCB 2019/10/09 review (csv format chosen)

PR #376, opened 2019-10-12
Fix #373: Correct mutex around log file write

### LINKED ISSUE #373: CFE_EVS_WriteLogFileCmd requires recursive locking
**Describe the bug**
The function `CFE_EVS_WriteLogFileCmd` in cfe_evs_log.c locks `CFE_EVS_GlobalData.EVS_SharedDataMutexID` for the entire duration of the function. If an error is encountered, it will call `EVS_SendEvent`, which also tries to lock the mutex. If the OS/OSAL do not support recursive locking on mutexes, this will lead to a deadlock.  

**To Reproduce**
1. Cause an error during the execution of `CFE_EVS_WriteLogFileCmd` (by e.g. giving it an invalid filename).

**Expected behavior**
This may be expected behavior, if cFE requires its mutexes to be recursively lockable. If so, this isn't a bug and I'll have to modify our OSAL. But if that is not the expectation, then I would this function should relinquish its lock before calling `EVS_SendEvent`.

More specifically, it looks like the this can be accomplished by moving the line
```
   OS_MutSemGive(CFE_EVS_GlobalData.EVS_SharedDataMutexID);
```
to just before each instance of `EVS_SendEvent()` in the function. By each of these points, this function is done accessing sensitive EVS data.

**Code snips**
* [CFE_EVS_WriteLogFileCmd's release of the lock](https://github.com/nasa/cFE/blob/6aaf5a3244e02af43c8ebb893ef413916dff5ef2/fsw/cfe-core/src/evs/cfe_evs_log.c#L251)

**System observed on:**
 - Capella flight computer
 - OS: FreeRTOS 10.1.1
 - Versions: cFE 6.5.0, OSAL 4.2.0 (plus in-house FreeRTOS port)

**Additional context**
I found this while testing our fix for #372.

**Reporter Info**
Mike Stewart, Capella Space.

jphickey:
The mutex for the log file write during the WriteLogFileCmd
excution was held longer than it should have been.

This lock must *NOT* be held during EVS_SendEvent, as this
will cause deadlock.

This moves the mutex to protect only the area that actually
accesses the log data, and it moves all send event calls
to be AFTER the mutex is released.

**Testing performed**
Built all code using default config and start CFE
Externally issue the WriteLogDataFile command with a valid output file
- Confirmed that the log data file is generated successfully
Externally issued the WriteLogDataFile command with a invalid output file (where OS_creat fails)
- Confirmed that the expected error event is generated with no deadlock or issues

**Expected behavior changes**
No longer deadlocks in the case of an error during the file operations

**System(s) tested on:**
Ubuntu 18.04.2 LTS 64-bit

**Contributor Info**

### DIFF
--- a/fsw/cfe-core/src/evs/cfe_evs_log.c
+++ b/fsw/cfe-core/src/evs/cfe_evs_log.c
@@ -167,9 +167,6 @@ int32 CFE_EVS_WriteLogDataFileCmd(const CFE_EVS_WriteLogDataFile_t *data)
     }
     else
     {
-        /* Serialize access to event log control variables */
-        OS_MutSemTake(CFE_EVS_GlobalData.EVS_SharedDataMutexID);
-
         /* Copy the commanded filename into local buffer to ensure size limitation and to allow for modification */
         CFE_SB_MessageStringGet(LogFilename, (const char *)CmdPtr->LogFilename, CFE_PLATFORM_EVS_DEFAULT_LOG_FILE,
                 OS_MAX_PATH_LEN, sizeof(CmdPtr->LogFilename));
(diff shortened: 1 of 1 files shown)


stashakk:
MSFC approved

CDKnightNASA:
ARC approved

skliper:
CCB 20191023 - approved

PR #401, opened 2019-11-08
fix-288-remove_pipename changes

CDKnightNASA:
**Describe the contribution**
Removes the PipeName field from the CFE_SB_PipeD_t struct.

**Testing performed**
make install
make test

**Expected behavior changes**
A clear and concise description of how this contribution will change behavior and level of impact.
 - API Change: internal change to CFE_SB_GetPipeName
 - Behavior Change: reduced memory impact of the CFE_SB_PipeD_t table

**System(s) tested on:**
 - Hardware: VM
 - OS: Linux
 - Versions: "master"

**Contributor Info**
Chris Knight, NASA Ames Research Center

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -180,8 +180,6 @@ int32  CFE_SB_CreatePipe(CFE_SB_PipeId_t *PipeIdPtr, uint16  Depth, const char *
     CFE_SB.PipeTbl[PipeTblIdx].CurrentBuff = NULL;
     CFE_SB.PipeTbl[PipeTblIdx].ToTrashBuff = NULL;
     strcpy(&CFE_SB.PipeTbl[PipeTblIdx].AppName[0],&AppName[0]);
-    strncpy(&CFE_SB.PipeTbl[PipeTblIdx].PipeName[0],PipeName,OS_MAX_API_NAME);
-    CFE_SB.PipeTbl[PipeTblIdx].PipeName[OS_MAX_API_NAME-1]='\0';
 
     /* Increment the Pipes in use ctr and if it's > the high water mark,*/
     /* adjust the high water mark */
@@ -207,8 +205,7 @@ int32  CFE_SB_CreatePipe(CFE_SB_PipeId_t *PipeIdPtr, uint16  Depth, const char *
     /* send debug event */
     CFE_EVS_SendEventWithAppID(CFE_SB_PIPE_ADDED_EID,CFE_EVS_EventType_DEBUG,CFE_SB.AppId,
           "Pipe Created:name %s,id %d,app %s",
-          CFE_SB_GetPipeName(CFE_SB.PipeTbl[PipeTblIdx].PipeId),
-          (int)CFE_SB.PipeTbl[PipeTblIdx].PipeId,
+          PipeName, (int)CFE_SB.PipeTbl[PipeTblIdx].PipeId,
           CFE_SB_GetAppTskName(TskId,FullName));
 
     return CFE_SUCCESS;
@@ -375,7 +372,6 @@ int32 CFE_SB_DeletePipeFull(CFE_SB_PipeId_t PipeId,uint32 AppId)
     CFE_SB.PipeTbl[PipeTblIdx].SysQueueId    = CFE_SB_UNUSED_QUEUE;
     CFE_SB.PipeTbl[PipeTblIdx].PipeId        = CFE_SB_INVALID_PIPE;
     CFE_SB.PipeTbl[PipeTblIdx].CurrentBuff   = NULL;
-    memset(&CFE_SB.PipeTbl[PipeTblIdx].PipeName[0],0,OS_MAX_API_NAME);
 
     /* zero out the pipe depth stats */
     if (PipeTblIdx < CFE_SB_TLM_PIPEDEPTHSTATS_SIZE)
@@ -687,6 +683,9 @@ int32  CFE_SB_SubscribeFull(CFE_SB_MsgId_t   MsgId,
     uint8  PipeIdx;
     CFE_SB_DestinationD_t *DestBlkPtr = NULL;
     char   FullName[(OS_MAX_API_NAME * 2)];
+    char   PipeName[OS_MAX_API_NAME] = {'\0'};
+
+    CFE_SB_GetPipeName(PipeId, PipeName);
 
     /* take semaphore to prevent a task switch during this call */
     CFE_SB_LockSharedData(__func__,__LINE__);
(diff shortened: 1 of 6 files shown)


skliper:
CCB 20191123 - Discussed the string copies vs passing pointers, or use the get properties but could result in fixing implementation (os_queues dependency would get built in).

This implementation saves memory, but is a performance hit with the string copies.  Not in the standard get/send path (just for error processing), and in the create/delete so not really a significant real time operation impact.

skliper:
CCB 20191123 - approved for integration candidate, under additional review by jphickey and acudmore

fsw/cfe-core/src/sb/cfe_sb_priv.c
```diff
@@ -497,24 +497,33 @@ CFE_SB_RouteEntry_t* CFE_SB_GetRoutePtrFromIdx(CFE_SB_MsgRouteIdx_t RouteIdx)
...
 
     if(PipeId >= CFE_PLATFORM_SB_MAX_PIPES){
-        return &PipeName4ErrCase[0];
+        Status = CFE_SB_FAILED;
     }else{
-        return &CFE_SB.PipeTbl[PipeId].PipeName[0];
+        if (OS_QueueGetInfo(CFE_SB.PipeTbl[PipeId].SysQueueId, &queue_prop)
+            == OS_SUCCESS)
+        {
+            strcpy(queue_prop.name, PipeNameBuf);
```
acudmore: Should you use strncpy here?

fsw/cfe-core/src/sb/cfe_sb_priv.c
```diff
@@ -497,24 +497,33 @@ CFE_SB_RouteEntry_t* CFE_SB_GetRoutePtrFromIdx(CFE_SB_MsgRouteIdx_t RouteIdx)
...
-**    is in range. Otherwise this function returns a pointer to the
-**    PipeName4ErrCase[0], which is initialized with a null terminator.
+**    Will return CFE_SUCCESS and populate PipeNameBuf with the name
+**    of the pipe on success, otherwise returns CFE_SB_FAILED on error.
 **
 */
-char *CFE_SB_GetPipeName(CFE_SB_PipeId_t PipeId){
-
-    static char PipeName4ErrCase[1] = {'\0'};
+int32 CFE_SB_GetPipeName(CFE_SB_PipeId_t PipeId, char *PipeNameBuf){
```
acudmore: Is it worth checking to see if PipeNameBuf is non-NULL? If it's only used internally, then it might not be necessary.

fsw/cfe-core/src/sb/cfe_sb_api.c
```diff
@@ -1047,10 +1046,13 @@ int32 CFE_SB_UnsubscribeFull(CFE_SB_MsgId_t MsgId,CFE_SB_PipeId_t PipeId,
 
     /* if there are no subscriptions for this message id... */
     if(!CFE_SB_IsValidRouteIdx(RouteIdx)){
+        char    PipeName[OS_MAX_API_NAME] = {'\0'};
+
+        CFE_SB_GetPipeName(PipeId, PipeName);
         CFE_SB_UnlockSharedData(__func__,__LINE__);
```
jphickey: In this case the CFE_SB_GetPipeName is called while the SB global lock is still held.  This should probably be switched.  In all other cases it looks like CFE_SB_GetPipeName is invoked outside the lock, which is probably preferable, as it avoids a double lock with OSAL.  OSAL already has protections in case a delete occurs in close proximity to this call.

fsw/cfe-core/src/sb/cfe_sb_priv.c
jphickey: I would also prefer to see this have a third parameter, which is the size of the output buffer, rather than assuming it is always exactly OS_MAX_API_NAME.  There are several reasons for this, but mainly that OS_MAX_API_NAME is an OSAL definition, not a CFE platform definition, and theoretically a single build of CFE could be linked with a different build of OSAL without necessarily having the same definition.  

Also note that the `CFE_ES_GetAppName()` function, which is similar in concept/purpose, has the following prototype:
`int32 CFE_ES_GetAppName(char *AppName, uint32 AppId, uint32 BufferLength)`

The output/destination buffer pointer is first, which mirrors the way many C library string functions work.

fsw/cfe-core/src/sb/cfe_sb_priv.c
```diff
@@ -497,24 +497,31 @@ CFE_SB_RouteEntry_t* CFE_SB_GetRoutePtrFromIdx(CFE_SB_MsgRouteIdx_t RouteIdx)
...
 
-    if(PipeId >= CFE_PLATFORM_SB_MAX_PIPES){
-        return &PipeName4ErrCase[0];
+    if(PipeNameBuf == NULL || PipeId >= CFE_PLATFORM_SB_MAX_PIPES){
+        Status = CFE_SB_FAILED;
     }else{
-        return &CFE_SB.PipeTbl[PipeId].PipeName[0];
+        if (OS_QueueGetInfo(CFE_SB.PipeTbl[PipeId].SysQueueId, &queue_prop)
+            == OS_SUCCESS){
+            strncpy(queue_prop.name, PipeNameBuf, OS_MAX_API_NAME-1);
```
jphickey: It looks like the source/destination is reversed here

fsw/cfe-core/src/sb/cfe_sb_priv.c
```diff
@@ -497,24 +497,31 @@ CFE_SB_RouteEntry_t* CFE_SB_GetRoutePtrFromIdx(CFE_SB_MsgRouteIdx_t RouteIdx)
...
+        }
+        else{
+            Status = CFE_SB_FAILED;
+        }/* end if */
     }/* end if */
 
+    return Status;
```
jphickey: I would also like to see this guaranteed to return a null-terminated string so long as a valid buffer was passed.  We also did this for `CFE_ES_GetAppName()` too -- although it returns a status as to whether or not it was successful, nobody checked the status, and it looks like this is also the case with this function.  If it fails, return an empty string.

CDKnightNASA:
> I would like to see some updates to the CFE_SB_GetPipeName() to always fill the buffer with a null terminated string (i.e. an empty string) even in cases of error,

Generally I prefer the minimum of change on error. Is there "style guidance" we should adopt and that should be documented in the coding style guide?

The SB code is not particularly consistent:

CreatePipe does this--initializes *PipeIdPtr to an invalid ID at the start of the function.

GetPipeOpts does NOT do this for *OptsPtr.

CFE_SB_GetAppTskName ...sorta...does this, by adding a null at the end of the buffer (no matter how long the actual name is.) But that, of course, does not handle the intervening bytes.

Remember, also, that CFE_SB_GetPipeName is a "private" function, although I could easily see exposing this. (In fact, there's one unit test already, which is kinda odd for a private function.)

> to also pass in a separate size,

Again this is not what CFE_SB_GetAppTskName does, a lot of functions assume (and are documented to assume) that the buffer is OS_MAX_API_NAME in length. If we want to change the "guidance" and start recommending array buffers be passed with size, I am good with that.

> and also reverse one case where it was called before releasing the SB lock.

Ok I will evaluate the locking logic per this and your other comment.

fsw/cfe-core/src/sb/cfe_sb_api.c
CDKnightNASA: Fixed!

fsw/cfe-core/src/sb/cfe_sb_priv.c
CDKnightNASA: GAH! I can't believe I did that. Fixed this.

fsw/cfe-core/src/sb/cfe_sb_priv.c
CDKnightNASA: fixed

skliper:
CCB - 12/4/2019: Consider adding an OSAL API which just retrieves the name of a resource, simplifies this issue and the other's related

CDKnightNASA:
OBE - pull request 404

Issue #415, opened 2019-11-20
Build system needs to better separate the UT stub function from the UT test cases


jphickey:
**Describe the bug**
Currently the "unit-test" subdirectory generates libraries that provide stubs for all public API calls defined in the CFE core apps, as well as test cases and executables for unit testing the CFE itself.

Having these combined like this causes a few problems:
- the two products have different scopes. The UT stubs apply globally across the entire mission and can/should be used for all UT (platform-independent), but the test cases do need to be tuned to the platform, so a separate set of test cases needs to be built per-platform.
- As a result, the UT stubs themselves are also built per-platform.  This is wasteful but more importantly this makes linking to the CFE stubs from application UT code difficult.
- It is also undesirable just from a general code organization standpoint.  The directory structure would be cleaner if these were in separate subdirectories.

**To Reproduce**
Build code with ENABLE_UNIT_TESTS=TRUE and observe that the CFE "stub" library is named "libut_cfe_core_default_cpu1_stubs.a" (i.e. build specifically for the default/cpu1 platform config) and it resides in the same directory as the UT executables/test cases.

**Expected behavior**
The stub library should be named only "libut_cfe_core_stubs.a" and it should be in a separate subdirectory from the unit test cases themselves.

**System observed on:**
Ubuntu 18.04 LTS 64 bit

**Reporter Info**
Joseph Hickey, Vantage Systems, Inc.

PR #417, opened 2019-11-21
Separate the CFE stubs from UT test cases

jphickey:
**Describe the contribution**
Fix issue #415

Build the CFE UT stub library separately from the test cases.
This moves the stub files into a separate ut-stubs directory,
and the library is now called "ut_cfe_core_stubs"

Note: Minor cleanup also done as part of moving.  Two stub
files were in the source tree but not being built or used
by any framework test config.  These were:
  ut_arinc653_stubs.c
  ut_configdata_stubs.c

These are now removed.  Also cleaned up some old macros/ifdefs
that were never enabled or used in the current tests.

**Testing performed**
Build CFE using default configuration with ENABLE_UNIT_TEST=TRUE.  Confirm CFE core executes normally.  Confirm all CFE unit tests build and run as normal.

**Expected behavior changes**
No impact to behavior.  CFE FSW code is unchanged, this only affects UT build/link procedure.

**System(s) tested on:**
Ubuntu 18.04 LTS 64-bit

**Contributor Info**

**Community contributors**


jphickey:
Also note: I noticed that the stub for the `CFE_TBL_GetInfo` function was missing, so this also adds it.

skliper:
fsw to me means "flight software".  Could the stubs and unit tests be moved out of the flight software directory as part of this change?  I'd expect the fsw directory to only contain flight code.

jphickey:
> fsw to me means "flight software". Could the stubs and unit tests be moved out of the flight software directory as part of this change? I'd expect the fsw directory to only contain flight code.

Yes, I concur that it doesn't make sense for this to be under "fsw".

But the difficulty for CFE stuff is that the "cfe-core" dir is considered the top level for all source code (i.e. this is where ${cfe-core_MISSION_DIR} points to) which is unfortunately a subdir of "fsw".

Perhaps the easiest solution would be swapping the "fsw" and "cfe-core" subdirs, so the path to the FSW would become:
`cfe-core/fsw/src/<module>`
rather than:
`fsw/cfe-core/src/<module`
as it is today.  That way the unit test stubs can reside at:
`cfe-core/ut-stubs`

Such a directory change seems like it should be a separate ticket/pull request though.

skliper:
> Such a directory change seems like it should be a separate ticket/pull request though.

I concur, a bigger change as suggested is worth a separate ticket.

skliper:
CCB 20191218 - Reviewed and approved for IC

PR #419, opened 2019-11-27
Fix #101, Always increment sequence counter

### LINKED ISSUE #101: SB Only Increments Message Sequence Count Where There are Subscribers
In the current implementation of the Software Bus, sequence counter management is handled in the routing table.  The routing table only includes entries that have subscribed messages.  This results in the sequence counter being incremented only if there are subscribers to a message.  If a message is unsubscribed to and then resubscribed, the sequence counter will be reset to zero.

A message sequence count should reflect the production of the message vs. the receipt of a message.  One use case for this implementation is to support the filtering of a message via the sequence count.

APL/SPP project implemented a solution to increment the sequence count upon production of the message regardless of whether or not there are any subscribers.  The solution involved removing the sequence count table from the routing table.  This exact solution may not support the new/expanded APID/message ID name space.  Expanding the message ID name space will need to be considered when implementing the overall solution to increment the sequence count to reflect message production.

CDKnightNASA:
**Describe the contribution**
Fixes issue #101 . Note that this means route entries will never be removed and unsubscribe messages will never be sent.

**Testing performed**
Built and ran the SB unit tests. All passed, with changes included in this pull req.

**Expected behavior changes**
Sequence ID's of messages should increment even when there's (currently) no subscribers to that message ID.

**System(s) tested on:**
Debian VM

**Contributor Info**
Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_sb_events.h
+++ b/fsw/cfe-core/src/inc/cfe_sb_events.h
@@ -430,20 +430,6 @@
 **/
 #define CFE_SB_SUBSCRIPTION_RPT_EID     22
 
-
-/** \brief <tt> 'Sending Unsubscription Report Msg=0x\%x,Pipe=\%d,Stat=0x\%x' </tt>
-**  \event <tt> 'Sending Unsubscription Report Msg=0x\%x,Pipe=\%d,Stat=0x\%x' </tt>
-**
-**  \par Type: DEBUG
-**
-**  \par Cause:
-**
-**  This debug event message is issued when SB subscription reporting is enabled,
-**  (which is disabled by default) and a subscription is successfully received.
-**/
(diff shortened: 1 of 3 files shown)


CDKnightNASA:
> Remove CFE_SB_UNSUBSCRIPTION_RPT_EID (I think this was the CCB decision?)

will do

CDKnightNASA:
I believe I've properly squashed the branch, removed the unused EID, added unit tests...anything else?

FYI for me what worked:

% git checkout <branch>
% git log # look for the first SHA before the branch
% git reset --soft <SHA>
% git commit -m 'new commit msg'
% git push origin +<branch>

skliper:
CCB 20200115 - Reviewed and approved for IC pending the changes already addressed above

skliper:
Updated commit message (no code change)

PR #450, opened 2020-01-03
Avoid alignment warnings on some CPUs

### LINKED ISSUE #313: Memory alignment issues in Software Bus
The CFE code has some problem areas for CPUs that have strict memory alignment requirements (e.g. SPARC).  Casting from a type with smaller requirements to a type with larger requirements generates a compile time warning and may induce a runtime exception if the memory actually is not aligned.

The following errors exist in the SB subsystem:

{{{
/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c: In function 'CFE_SB_ZeroCopyGetPtr':
/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1853:12: warning: cast increases required alignment of target type [-Wcast-align]
     return (CFE_SB_Msg_t *)address;
            ^
/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c: In function 'CFE_SB_ZeroCopyReleasePtr':
/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1896:35: warning: cast increases required alignment of target type [-Wcast-align]
                                   (uint32 *) (((uint8 *)Ptr2Release) - sizeof(CFE_SB_BufferD_t)));
                                   ^
/cfe/fsw/cfe-core/src/sb/cfe_sb_buf.c: In function 'CFE_SB_GetBufferFromCaller':
/cfe/fsw/cfe-core/src/sb/cfe_sb_buf.c:107:30: warning: cast increases required alignment of target type [-Wcast-align]
    CFE_SB_BufferD_t    *bd = (CFE_SB_BufferD_t *)(((uint8 *)Address) - sizeof(CFE_SB_BufferD_t));
                              ^
/cfe/fsw/cfe-core/src/sb/cfe_sb_util.c: In function 'CFE_SB_GetMsgTime':
/cfe/fsw/cfe-core/src/sb/cfe_sb_util.c:280:21: warning: cast increases required alignment of target type [-Wcast-align]
         TlmHdrPtr = (CFE_SB_TlmHdr_t *)MsgPtr;
                     ^
/cfe/fsw/cfe-core/src/sb/cfe_sb_util.c: In function 'CFE_SB_SetMsgTime':
/cfe/fsw/cfe-core/src/sb/cfe_sb_util.c:345:21: warning: cast increases required alignment of target type [-Wcast-align]
         TlmHdrPtr = (CFE_SB_TlmHdr_t *) MsgPtr;
                     ^
}}}

NOTE: This is related to PSP ticket [cfs_psp:41]

### LINKED ISSUE #437: Alignment warnings when casting char* pointers
**Describe the bug**
When compiling on architectures with strict alignment requirements, casts between `char*` (as used in some pointer arithmetic to calculate addresses) to other types triggers a warning about alignment.

**To Reproduce**
Build for an architecture with strict alignment requirements (SPARC, MIPS, etc).
Example compiler warning:
```
/home/joe/code/cfecfs/github/cfe/fsw/cfe-core/src/sb/cfe_sb_buf.c:119:30: warning: cast increases required alignment of target type [-Wcast-align]
    CFE_SB_BufferD_t    *bd = (CFE_SB_BufferD_t *)(((uint8 *)Address) - sizeof(CFE_SB_BufferD_t));
```

**Expected behavior**
Should build cleanly with no warnings

**System observed on:**
- Ubuntu 18.04 LTS build host
- MIPS cross compiler (mips-poky-linux-gcc version 8.2.0)

**Additional context**
Many of these are actually safe because CFE ensures that the `sizeof(CFE_SB_BufferD_t)` as well as the base/pool addresses are aligned for the largest data types.  However, because the pointer arithmetic is done as a `unsigned char*` the compiler sees the cast from an `unsigned char*` to a larger type as an issue.

**Reporter Info**

jphickey:
**Describe the contribution**

Fix #437 and partially address #313 (overlapping issue)

On CPUs with strict alignment requirements, some CFE code that uses a char-type pointer (e.g. uint8*) to compute memory addresses triggers an alignment warning when it gets cast back to the actual data type.

In the mempool implementation, the pointer should be sufficiently aligned already, because the address computation already takes CPU alignment requirements into account when calculating the addresses/offsets.  

- For the CFE_SB pool buffers, using the `cpuaddr` type, which is integer in nature, avoids the warning.
- For the CFE_TBL internal table pointer, use a `void*` internally to store the buffer pointer, rather than a `uint8_t*`.  This changes the casting needs elsewhere.

**Testing performed**
Build CFE with ENABLE_UNIT_TESTS=TRUE
Confirm all unit tests passing
Perform sanity test on CFE (normal startup, send commands from console)
Build for MIPS64 and ensure that (some) alignment warnings are fixed

**Expected behavior changes**
No change to behavior.  Fixes build warnings only.

**System(s) tested on:**
Ubuntu 18.04 LTS  64-bit

**Additional context**
There are still some remaining alignment cast warnings regarding the message types, where a local message buffer is cast to a `CFE_SB_Msg_t*` which has a higher alignment requirement.  This is a little harder to fix as it requires changing the local buffer definition.

**Contributor Info**

**Community contributors**

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -1802,7 +1802,7 @@ CFE_SB_Msg_t  *CFE_SB_ZeroCopyGetPtr(uint16 MsgSize,
 {
    int32                stat1;
    uint32               AppId = 0xFFFFFFFF;
-   uint8               *address = NULL;
+   cpuaddr              address = 0;
    CFE_SB_ZeroCopyD_t  *zcd = NULL;
    CFE_SB_BufferD_t    *bd = NULL;
 
@@ -1849,7 +1849,7 @@ CFE_SB_Msg_t  *CFE_SB_ZeroCopyGetPtr(uint16 MsgSize,
     }/* end if */
 
     /* first set ptr to actual msg buffer the same as ptr to descriptor */
-    address = (uint8 *)bd;
+    address = (cpuaddr)bd;
 
     /* increment actual msg buffer ptr beyond the descriptor */
     address += sizeof(CFE_SB_BufferD_t);
@@ -1914,6 +1914,7 @@ int32 CFE_SB_ZeroCopyReleasePtr(CFE_SB_Msg_t  *Ptr2Release,
 {
     int32    Status;
     int32    Stat2;
+    cpuaddr  Addr = (cpuaddr)Ptr2Release;
 
     Status = CFE_SB_ZeroCopyReleaseDesc(Ptr2Release, BufferHandle);
 
@@ -1922,7 +1923,7 @@ int32 CFE_SB_ZeroCopyReleasePtr(CFE_SB_Msg_t  *Ptr2Release,
     if(Status == CFE_SUCCESS){
         /* give the buffer back to the buffer pool */
         Stat2 = CFE_ES_PutPoolBuf(CFE_SB.Mem.PoolHdl,
-                                  (uint32 *) (((uint8 *)Ptr2Release) - sizeof(CFE_SB_BufferD_t)));
+                                  (uint32 *) (Addr - sizeof(CFE_SB_BufferD_t)));
         if(Stat2 > 0){
              /* Substract the size of the actual buffer from the Memory in use ctr */
             CFE_SB.StatTlmMsg.Payload.MemInUse-=Stat2;
--- a/fsw/cfe-core/src/sb/cfe_sb_buf.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_buf.c
@@ -116,7 +116,7 @@ CFE_SB_BufferD_t * CFE_SB_GetBufferFromPool(CFE_SB_MsgId_t MsgId, uint16 Size) {
 
 CFE_SB_BufferD_t * CFE_SB_GetBufferFromCaller(CFE_SB_MsgId_t MsgId,
                                               void *Address) {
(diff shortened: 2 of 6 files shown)


jphickey:
Update: changed the description to reflect that this will fix the `char*` casts cited in #437 but the older issue #313 also mentions issues related to the message types (`CFE_SB_Msg_t*` casts) that this does not fully address.

There will need to be an additional pull request to fix those ones.

jphickey:
>  does this mean however that the pointer may not be aligned? I don't know if there are any CPU's that cannot interact with unaligned data but it will affect performance on many platforms. Should we add code to align?

SPARC and MIPS processors (at least) *will* throw an exception when attempting to access a data structure that is not properly aligned.  I believe ARM and x86 will auto-correct but there is a performance penalty to pay.  Not sure about PowerPC.

For the items specifically in this pull request (the pool address calculation) the buffer descriptor is already a union including large types, such that `sizeof(BD_t)` will be a multiple of the system alignment required for the large.  In turn all pool buffers are also already allocated to be a multiple of this size.

So all the pointers are already aligned sufficiently, the problem is merely that it was using `uint8*` intermediate pointer to perform the address calculations, and then casting a `uint8*` back to the intended type.  This final step is what triggered a warning, just because the final type has a stricter alignment requirement than `uint8*` does.

By doing the intermediate address calculation as an integer (cpuaddr) the result is exactly the same but gcc doesn't warn about it.

skliper:
CCB 20190108 - Reviewed and approved for integration candidate

PR #452, opened 2020-01-06
Fix #308, Improve SB create pipe error reporting

### LINKED ISSUE #308: CFE_SB_CreatePipe: Improve message when OS_QueueCreate fails
Currently if `OS_QueueCreate` fails in `CFE_SB_CreatePipe`, the message is just `"CreatePipeErr:OS_QueueCreate returned %d,app %s"`. It would be valuable to show the name of the error rather than (or in addition to) the numeric error code. The `OS_GetErrorName` function may help here if it's available in all of the OSAL implementations.

It may be valuable to provide other details depending on the error. For example, the message in the case of `OS_ERR_NAME_TAKEN` could indicate the name of the pipe that was requested (and is already in use).

CDKnightNASA:
**Describe the contribution**
Improves error reporting for CFE_SB_CreatePipe()

**Testing performed**
Standard build process, ran SB unit tests.

**Expected behavior changes**
Improvement in error reporting when using a pipe name that is already in use, or when the queue limit has been reached.

**System(s) tested on:**
Debian 9

**Additional context**

**Contributor Info**
Christopher.D.Knight@nasa.gov

**Community contributors**
N/A

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -160,13 +160,27 @@ int32  CFE_SB_CreatePipe(CFE_SB_PipeId_t *PipeIdPtr, uint16  Depth, const char *
 
         /* if OS_QueueCreate() failed because the pipe name passed in was already in use... */
         /* let's make sure we don't alter the user's pipe ID data */
-        if (Status == CFE_OS_ERR_NAME_TAKEN){
-            *PipeIdPtr = OriginalPipeIdParamValue;
-        }
+        switch(Status) {
+            case OS_ERR_NAME_TAKEN:
+                CFE_EVS_SendEventWithAppID(CFE_SB_CR_PIPE_NAME_TAKEN_EID,CFE_EVS_EventType_ERROR,CFE_SB.AppId,
+                    "CreatePipeErr:OS_QueueCreate failed, name taken (app=%s, name=%s)",
+                    CFE_SB_GetAppTskName(TskId,FullName), PipeName);
 
-        CFE_EVS_SendEventWithAppID(CFE_SB_CR_PIPE_ERR_EID,CFE_EVS_EventType_ERROR,CFE_SB.AppId,
+                *PipeIdPtr = OriginalPipeIdParamValue;
+
+                break;
+            case OS_ERR_NO_FREE_IDS:
+                CFE_EVS_SendEventWithAppID(CFE_SB_CR_PIPE_NO_FREE_EID,CFE_EVS_EventType_ERROR,CFE_SB.AppId,
+                    "CreatePipeErr:OS_QueueCreate failed, no free id's (app=%s)",
+                    CFE_SB_GetAppTskName(TskId,FullName));
+
+                break;
+            default:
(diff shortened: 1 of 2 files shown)


fsw/cfe-core/src/sb/cfe_sb_api.c
```diff
@@ -1,5 +1,5 @@
 /*
-**  GSC-18128-1, "Core Flight Executive Version 6.7"
+**  GSC-18128-1, "Core Flight Executive Version 6.6"
```
skliper: Should be left at 6.7

CDKnightNASA: fixed

skliper:
Squashed, ready for review.

fsw/cfe-core/src/sb/cfe_sb_api.c
```diff
@@ -160,13 +160,27 @@ int32  CFE_SB_CreatePipe(CFE_SB_PipeId_t *PipeIdPtr, uint16  Depth, const char *
...
+
+                break;
+            case OS_ERR_NO_FREE_IDS:
+                CFE_EVS_SendEventWithAppID(CFE_SB_CR_PIPE_ERR_EID,CFE_EVS_EventType_ERROR,CFE_SB.AppId,
+                    "CreatePipeErr:OS_QueueCreate failed, no free id's (app=%s)",
+                    CFE_SB_GetAppTskName(TskId,FullName));
+
+                break;
+            default:
+                CFE_EVS_SendEventWithAppID(CFE_SB_CR_PIPE_ERR_EID,CFE_EVS_EventType_ERROR,CFE_SB.AppId,
```
jphickey: This should either both use the OSAL-defined code (OS_ERR_NAME_TAKEN, OS_ERR_NO_FREE_IDS) or the CFE-defined equivalent (CFE_OS_ERR_NAME_TAKEN, CFE_OS_ERR_NO_FREE_IDS) but not one of each.

My preference is to use the actual OSAL return codes, the CFE wrapper ones don't add much value.

CDKnightNASA:
I’m good with this, but is this the first time cFE sends OSAL EID’s upstream? I was under the impression that cFE always remapped OSAL EIDs to cFE EIDs.

jphickey:
This is not sending OSAL return codes in an event - it is sending the CFE-defined event based on the OSAL return code.

Because the actual call (`OS_QueueCreate()` here) is returning e.g. OS_ERR_NAME_TAKEN and the value is not "rewritten" - it is returned directly from the encapsulating function - then that is what should be checked for here, as opposed CFE_OS_ERR_NAME_TAKEN.

CDKnightNASA:
I stand corrected, I’m good with this.

CDKnightNASA:
> Just to translate to a request - add unique EID's for each error (per non-open-sourced coding standard).

copy will implement by next week's CCB

CDKnightNASA:
> Just to translate to a request - add unique EID's for each error (per non-open-sourced coding standard).

resolved

CDKnightNASA:
> Confirm current unit tests cover these new cases, needs squash

squashed

skliper:
CCB 20200122 - Reviewed and approved for CI pending confirmation of test coverage

PR #457, opened 2020-01-07
Fix #24 #355, Add compile flag abstractions

### LINKED ISSUE #24: Sample makefile should use strict compiler flags and treat warnings as errors
**Describe the bug**
The default makefiles / CMake scripts do not enforce any strict compiler warning flags and do not treat warnings as errors.  This means issues like #22 can more easily slip past.

**To Reproduce**
Building the `rc-6.7.0` branch as a "release" (-O3) using the default settings/sample config/makefile wrapper by preparing as:

`make SIMULATION=native BUILDTYPE=release prep`

Then run make and you do get some warnings, at least when using gcc 7.x and above:

```
[ 56%] Building C object cfe_core_default_cpu1/CMakeFiles/cfe_core_default_cpu1.dir/src/es/cfe_es_task.c.o
/home/joe/code/cfecfs/github/cfe/fsw/cfe-core/src/es/cfe_es_task.c: In function ‘CFE_ES_TaskInit’:
/home/joe/code/cfecfs/github/cfe/fsw/cfe-core/src/es/cfe_es_task.c:374:64: error: array subscript is below array bounds [-Werror=array-bounds]
        strncat(EventBuffer, VersionBuffer, sizeof(EventBuffer)-strlen(EventBuffer-1));
                                                                ^~~~~~~~~~~~~~~~~~~~~
/home/joe/code/cfecfs/github/cfe/fsw/cfe-core/src/es/cfe_es_task.c:380:64: error: array subscript is below array bounds [-Werror=array-bounds]
        strncat(EventBuffer, VersionBuffer, sizeof(EventBuffer)-strlen(EventBuffer-1));
                                                                ^~~~~~~~~~~~~~~~~~~~~
```

However, the build continues and completes the process with no error results.

**Expected behavior**
The build should stop, because warnings are problems that need to be resolved.  If the build does not stop then it is very easy to not notice these issues.

**System observed on:**
Ubuntu 18.04 (64-bit), kernel 5.0.0-23-generic, gcc 7.4.0

**Reporter Info**

### LINKED ISSUE #355: Add flag at prep stage to OMIT_DEPRECATED
Need an easy way for users (and the CI system) to omit all deprecated elements.  Something like

make OMIT_DEPRECATED=true prep

which would then internally add all the -DCFE_OMIT_DEPRECATED -DOSAL_OMIT_DEPRECATED and so on flags when building everything.

skliper:
**Describe the contribution**
Fixes #24, Fixes #355 

Adds STRICT_NO_WARNINGS and OMIT_DEPRECATED
prep options for CI and as example build

**Testing performed**
Built with STRICT_NO_WARNINGS=true, false, and undefined
Built with OMIT_DEPRECATED=true, false, and undefined (note sample_app fails)

Confirmed flags get globally applied (not great scoping, but ensures flag coverage)

**Expected behavior changes**
Simplifies/enables CI based on abstracted flags, see https://github.com/nasa/cFS/issues/39

**System(s) tested on:**
 - cFS Dev Server 2
 - OS: Ubuntu 18.04
 - Versions: Master bundle with this change

**Contributor Info**
Jacob Hageman - NASA/GSFC

### DIFF
--- a/cmake/sample_defs/targets.cmake
+++ b/cmake/sample_defs/targets.cmake
@@ -98,3 +98,21 @@ SET(TGT1_FILELIST cfe_es_startup.scr)
 #SET(TGT3_APPLIST sample_app ci_lab to_lab sch_lab)
 #SET(TGT3_FILELIST cfe_es_startup.scr)
 
+# Set host strict warnings, gets applied to all host compiled elements
+# Note target strict warnings are defined in associated toolchain file
+SET(HOST_STRICT_WARNINGS "-Wall -Werror -std=c99 -D_XOPEN_SOURCE=600 -pedantic -Wstrict-prototypes -Wcast-align -Wwrite-strings")
+
+# Cache related environment variables
+# NOTE - requires distclean and re-"make prep" to change
+set(OMIT_DEPRECATED $ENV{OMIT_DEPRECATED} CACHE STRING "Omit deperecated elements")
+
+# Helper definition to abstract omitting deprecated elements,
+# allows for one setting in CI and used by projects
+if (OMIT_DEPRECATED)
+  message (STATUS "OMIT_DEPRECATED=true: Not including deprecated elements in build")
+  add_definitions(-DCFE_OMIT_DEPRECATED_6_6 -DOSAL_OMIT_DEPRECATED)
+else()
+  message (STATUS "OMIT_DEPRECATED=false: Deprecated elements included in build")
+endif (OMIT_DEPRECATED)
+
+
--- a/cmake/sample_defs/toolchain-cpu1.cmake
+++ b/cmake/sample_defs/toolchain-cpu1.cmake
@@ -23,3 +23,5 @@ SET(CFE_SYSTEM_PSPNAME      "pc-linux")
 SET(OSAL_SYSTEM_BSPNAME     "pc-linux")
 SET(OSAL_SYSTEM_OSTYPE      "posix")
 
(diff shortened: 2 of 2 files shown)


skliper:
Preference is for complexity to be shifted to the build system vs CI to simplify user's burden to execute similar tests locally (and adapt the sample as needed).  Our CI has external dependencies that the build system doesn't have, if a user want's to try out the examples it's trivial and obvious when provided as part of the build system (just flags to prep).

jphickey:
Understood, but CI needs to be more extensive anyway, because it currently does not sufficiently cover the compile-time options available.  This has been discussed in the past and also I believe there are some related tickets to test with different configs.

My suggestion is to make more than one "defs" directory here, so the `sample_defs` (which the guides suggest new users clone as their starting point) can remain simple and straightforward.

The other "defs" directories can be used during CI as well as examples of more complex configs.  

But the main issue is the logic being proposed is something very CI-specific, and absolutely **NOT** a good example of the right way to do things.

skliper:
CCB - 20200108: discussed compiler flags in depth, scoping for mission/host vs target would improve things, Joe will submit an update

skliper:
Took a shot at better scoping, and applied by default.  Note this just defines HOST_STRICT_COMPILE and TARGET_STRICT_COMPILE, the add_definitions would need to go in the appropriate locations.

skliper:
Rejected in favor of #462 and #463

PR #485, opened 2020-01-21
Fix #274, NumOfChildTasks not decremented

### LINKED ISSUE #274: ES AppTable.TaskInfo.NumOfChildTasks incremented but never decremented

dmknutsen:
**Describe the contribution**
Fixes #274, NumOfChildTasks not decremented

**Testing performed**
1. Nominal build process.
2. Ran unit tests.

**Expected behavior changes**
 NumOfChildTasks will now be decremented when a child task is deleted.

**System(s) tested on**
Oracle VM VirtualBox
OS: ubuntu-19.10
Version: cFE 6.7.3.0; OSAL 5.0.3.0; PSP 1.4.1.0

Dan Knutsen
GSFC/NASA

### DIFF
--- a/fsw/cfe-core/unit-test/es_UT.c
+++ b/fsw/cfe-core/unit-test/es_UT.c
@@ -4306,8 +4306,8 @@ void TestAPI(void)
     char CDSName[CFE_MISSION_ES_CDS_MAX_NAME_LENGTH + 2];
     int i;
     uint32 ExceptionContext = 0;
-    int32 Return;
-    uint8 Data[12];
+    int32  Return;
+    uint8  Data[12];
     uint32 ResetType;
     uint32 *ResetTypePtr;
     uint32 AppId;
@@ -4318,6 +4318,7 @@ void TestAPI(void)
     uint32 CounterCount;
     CFE_ES_CDSHandle_t CDSHandle;
     CFE_ES_TaskInfo_t TaskInfo;
+    CFE_ES_AppInfo_t AppInfo;
 
 #ifdef UT_VERBOSE
     UT_Text("Begin Test API\n");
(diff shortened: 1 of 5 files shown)


skliper:
CCB 20200122 - Discussed, preference is to calculate on request and remove the counter.

fsw/cfe-core/unit-test/es_UT.c
```diff
@@ -4793,14 +4795,38 @@ void TestAPI(void)
...
+    CFE_ES_Global.TaskTable[18].AppId = 15;
+    CFE_ES_Global.AppTable[15].AppState = CFE_ES_AppState_RUNNING;
+    CFE_ES_Global.AppTable[15].TaskInfo.MainTaskId = 25;
+
+    OS_TaskCreate(&CFE_ES_Global.TaskTable[16].TaskId, NULL, NULL, NULL,
                   0, 0, 0);
+   
+    CFE_ES_GetAppInfo(&AppInfo,CFE_ES_Global.TaskTable[1].AppId);    
+    uint32 NumORegTasksTemp = CFE_ES_Global.RegisteredTasks;
+    uint32 NumOChildTastTemp1 = AppInfo.NumOfChildTasks;
```
jphickey: Wouldn't the "NumOfChildTasks" here be a predictable/consistent number based on the previous TaskTable config?  That is, I think we should know exactly how many child task entries are associated with the AppId because it was just set up earlier in the test sequence.  If we do have a specific number which is the correct answer based on an explicitly set config, then we should assert for that specific number.

Also as a general style thing I think we try to avoid instantiating local variables in the middle of the function.

fsw/cfe-core/unit-test/es_UT.c
```diff
@@ -4797,10 +4801,17 @@ void TestAPI(void)
...
-                  0, 0, 0);
+                  0, 0, 0);    
+    DeleteChildResult = CFE_ES_DeleteChildTask(CFE_ES_Global.TaskTable[1].TaskId);    
+    CFE_ES_GetAppInfo(&AppInfo,CFE_ES_Global.TaskTable[1].AppId);    
+
+    if( DeleteChildResult == CFE_SUCCESS && CFE_ES_Global.RegisteredTasks == 15
+        && AppInfo.NumOfChildTasks == 0 )
+    {
+       DeleteChildTestResult = CFE_SUCCESS;
+    }
```
jphickey: Each of these conditions should be a separate assert.  When combined into a single assert then it is not possible to see what went wrong from a log file, making it difficult to diagnose.

Please use the `UtAssert_True()` macro to test each individual condition you are looking for, and also output the actual values too (not just a generic text) for debug.

The general pattern should be something like this:

```
Return = CFE_ES_DeleteChildTask(CFE_ES_Global.TaskTable[1].TaskId);
UtAssert_True(Return == CFE_SUCCESS, 
         "DeleteChildResult() return=%x", (unsigned int)Return);
Return = CFE_ES_GetAppInfo(&AppInfo,CFE_ES_Global.TaskTable[1].AppId);
UtAssert_True(Return == CFE_SUCCESS, 
         "CFE_ES_GetAppInfo() return=%x", (unsigned int)Return);
UtAssert_True(AppInfo.NumOfChildTasks == 0, 
         "AppInfo.NumOfChildTaskss == %u", (unsigned int)AppInfo.NumOfChildTasks);
```
And so on....

It also is not clear to me why we verify `CFE_ES_Global.RegisteredTasks == 15` as part of this test case.  I expected to see a call to CFE_ES_GetAppInfo before and after the call to CFE_ES_DeleteChildTask, to verify that the NumOfChildTasks output was the expected number.

jphickey: Also Note - asserting on each condition separately should also remove the need for extra local variables.

fsw/cfe-core/unit-test/es_UT.c
```diff
@@ -4797,12 +4798,22 @@ void TestAPI(void)
...
+         "CFE_ES_GetAppInfo() return=%x", (unsigned int)Return); 
+    UtAssert_True(AppInfo.NumOfChildTasks == 1, 
+         "AppInfo.NumOfChildTaskss == %u", (unsigned int)AppInfo.NumOfChildTasks); 
+    Return = CFE_ES_DeleteChildTask(CFE_ES_Global.TaskTable[1].TaskId);  
+    UtAssert_True(Return == CFE_SUCCESS, 
+         "DeleteChildResult() return=%x", (unsigned int)Return);
+    Return = CFE_ES_GetAppInfo(&AppInfo,CFE_ES_Global.TaskTable[1].AppId);
+    UtAssert_True(Return == CFE_SUCCESS, 
+         "CFE_ES_GetAppInfo() return=%x", (unsigned int)Return); 
+    UT_Report(__FILE__, __LINE__,
```
jphickey: All good up through here.  But why is this last case going back to using UT_Report?  It should look the same as line 4805, using UtAssert_True, just testing for a different expected value (0 instead of 1).

(UT_Report was just a compatibility wrapper for old test cases, and it doesn't allow context info to be logged)

skliper:
Reset soft to master and checked in again.

fsw/cfe-core/unit-test/es_UT.c
astrogeco: Resolving based on @jphickey comment below.

astrogeco:
CCB 20200212 - Approved

PR #533, opened 2020-02-26
Fix #532, Link CFE core with OSAL BSP

### LINKED ISSUE #532: CFE core build should link with the OSAL BSP library
The CFE PSP has historically duplicated some logic that was also in the OSAL BSP.  In an effort to consolidate this and have only one basic BSP implementation, the CFE PSP should be an extension of the OSAL BSP rather than duplicating the logic.

The first step to accomplish this is to have the CFE final link also include the `osal_bsp` library in the final link.  

**Additional context**
Including `osal_bsp` should not affect existing builds as long as it is included _after_ the PSP, so the linker will not pull in duplicate symbols.  It should be effectively a no-op/ignored in the current build but enables a future transition to using the osal-provided BSP for startup.

**Requester Info**

jphickey:
**Describe the contribution**
Link the CFE core target with the OSAL BSP library by including this in the final link library list.

Fixes #532

**Testing performed**
Build and run unit tests on native (x86-64) and PC-RTEMS.  Confirm CFE runs normally, no change.

**Expected behavior changes**
No behavior changes.  This change only enables future transition to using the OSAL BSP rather than
duplicating startup logic in the CFE PSP.  Initially the linker should drop/ignore the content of this library.

**System(s) tested on**
 - Ubuntu 18.04 LTS, 64 bit (native)
 - PC-RTEMS using i686-rtems4.11 cross build

**Additional context**
This depends on the related change in nasa/osal#261 which defines the separate `osal_bsp` target.  It should be merged at the same time (or after) the OSAL change is merged.


astrogeco:
CCB 2020-03-25: Approved

jphickey:
Please hold off on merging this one for now, it has a relationship to nasa/osal#261 and on second look it might be easier to handle it via `target_link_libraries` on OSAL.

astrogeco:
> Please hold off on merging this one for now, it has a relationship to [nasa/osal#261](https://github.com/nasa/osal/issues/261) and on second look it might be easier to handle it via `target_link_libraries` on OSAL.

I merged it into the IC but I won't merge IC into master. I see that the nasa/osal#261 doesn't have an attached PR yet does it make sense to delay this fix if the new solution doesn't exist yet? @skliper thoughts?

skliper:
I'd back it out of the IC branch if @jphickey doesn't think this is the way to go, or at least hold off on merge until confirmed.

astrogeco:
> I'd back it out of the IC branch if @jphickey doesn't think this is the way to go, or at least hold off on merge until confirmed.

I'll bring it up at CCB

astrogeco:
CCB 2020-04-01 @jphickey working on alternate implementation. Remove from integration-candidate

PR #542, opened 2020-03-04
Fix #539, build failure in UT stubs with multiple CPUs

### LINKED ISSUE #539: Unit test build fails when multiple CPUs are defined
**Describe the bug**
After updating some projects to the latest baseline, getting a new build failure:

```
[ 73%] Building C object ut_cfe_core_stubs/CMakeFiles/ut_cfe-core_stubs.dir/ut_es_stubs.c.o
cfe/fsw/cfe-core/ut-stubs/ut_es_stubs.c:37:10: fatal error: cfe_platform_cfg.h: No such file or directory
 #include "cfe_platform_cfg.h"
          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

**To Reproduce**
The configuration of this project uses multiple CPUs that all share the same architecture (PPC) but different platform configs (cfe_platform_cfg.h).

**Expected behavior**
Build should succeed

**System observed on:**
 - Ubuntu 18.04 LTS 64-bit (build host)
 - Cross Compiling for PPC (embedded Linux)
 - Latest "master" baseline (2020-02-26)

**Additional context**
This is related to a recent change that separated the UT stubs from the CFE.  It works fine with only a single CPU/config, but if multiple CPUs are defined then this fails.  Unfortunately only testing this now - the CI/CD build only uses a single CPU/config.

Need to remove UT dependencies on a specific "cfe_platform_cfg.h" to fix this.  (stubs should not need this file, it is not a real implementation, it is just stubs).

**Reporter Info**

jphickey:
**Describe the contribution**

Fix #539 

Removes dependencies on `cfe_platform_cfg.h` from the unit test stub code.  The behavior of the stubs should not be dependent on the FSW runtime config.

This also fixes a duplicate name collision where different configs exist within the same architecture, causing a conflict on the UT target names.

**Testing performed**
Build with ENABLE_UNIT_TESTS=TRUE for both native and cross environments, with both a default config (single CPU/config) and a more complex two CPU config.  

Execute unit tests and confirm all passing in all configs.

**Expected behavior changes**
The two CPU config originally fails and now builds and runs UT with this change.
No changes to FSW, only UT.

**System(s) tested on**
Ubuntu 18.04 LTS 64 bit
Cross build for 32-bit architecture

Joseph Hickey, Vantage Systems, Inc. (CLA on file)

### DIFF
--- a/fsw/cfe-core/unit-test/CMakeLists.txt
+++ b/fsw/cfe-core/unit-test/CMakeLists.txt
@@ -32,42 +32,47 @@ add_library(ut_${CFE_CORE_TARGET}_support STATIC
 # UT version of the real module (compiled with coverage flags)
 foreach(MODULE ${CFE_CORE_MODULES})
 
+  # The "UT_TARGET_NAME" is a concatenation of the configuration name with the current module
+  # This avoids target name duplication in case more than one configuration is used
+  # (All  targets should be based on this name)
+  set(UT_TARGET_NAME "${CFE_CORE_TARGET}_${MODULE}")
+  
   set(CFE_MODULE_FILES)
   aux_source_directory(${cfe-core_MISSION_DIR}/src/${MODULE} CFE_MODULE_FILES)
   
   # Compile the unit(s) under test as an object library
   # this allows easy configuration of special flags and include paths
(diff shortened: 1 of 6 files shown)


astrogeco:
CCB 20200311 - Approve

PR #584, opened 2020-04-02
Fix #335: Shell unsigned pkt length bug

### LINKED ISSUE #335: CFE_MISSION_ES_MAX_SHELL_PKT Causes ES Error When Unsigned
When CFE_MISSION_ES_MAX_SHELL_PKT is defined as an unsigned value (2000u, for example) and a shell command is executed that results in a shell output file of less than CFE_MISSION_ES_MAX_SHELL_PKT, ES gets stuck in a loop of sending millions of ES Shell Telemetry packets.

This error/loop is in the int32 CFE_ES_ShellOutputCommand(const char * CmdString, const char *Filename) function. I think it is caused by the FileSize variable (int32) getting converted to an uint32 and under-flowing.

A workaround to this is to keep CFE_MISSION_ES_MAX_SHELL_PKT defined as a signed integer, but I think that the fix should be for ES to fix this under-flow. The CFE_MISSION_ES_MAX_SHELL_PKT should be able to be thought of/defined as an unsigned value.

skliper:
**Describe the contribution**
Fix #335 
Really just avoids problems if users mistakenly define CFE_MISSION_ES_MAX_SHELL_PKT as an unsigned value (2000u, for example)

**Testing performed**
1. CI

Suggest requester (@krmoore) confirm it works as requested

**Expected behavior changes**
None, except if someone defines this value as unsigned it won't break...

**System(s) tested on**
 - Hardware: AMD
 - OS: Ubuntu 18.04
 - Versions: bundle + this change

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_shell.c
+++ b/fsw/cfe-core/src/es/cfe_es_shell.c
@@ -170,7 +170,7 @@ int32 CFE_ES_ShellOutputCommand(const char * CmdString, const char *Filename)
                 /* start processing the chunks. We want to have one packet left so we are sure this for loop
                 * won't run over */
         
-                for (CurrFilePtr=0; CurrFilePtr < (FileSize - CFE_MISSION_ES_MAX_SHELL_PKT); CurrFilePtr += CFE_MISSION_ES_MAX_SHELL_PKT)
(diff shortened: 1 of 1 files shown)


astrogeco:
CCB 20200415 - APPROVED

PR #585, opened 2020-04-02
Fix #581, Propagate the OSAL compile definitions to CFE build

### LINKED ISSUE #581: CFE needs to compile with the interface properties provided by OSAL
After the changes introduced by nasa/osal#312, the OSAL CMake script will no longer force setting of `CMAKE_C_FLAGS` directly in the CFE build by overwriting this value (which is good).  

Instead, any required compile definitions and related information will be conveyed in the more appropriate manner, through the `INTERFACE_COMPILE_DEFINITIONS` and `INTERFACE_INCLUDE_DIRECTORIES` on the "osal" library target.

To work with this change, the CFE needs to explicitly check these properties on the osal target and use the values.  By setting a directory-scope property at the top level, it will apply to all code.

CMake does this automatically so long as the executable directly links with OSAL.  However, this doesn't apply to the apps which are built as a `MODULE`.   This is why it is easier to set the directory property so its all-inclusive.

**Requester Info**

jphickey:
**Describe the contribution**
Use the `INTERFACE_COMPILE_DEFINITIONS` and `INTERFACE_INCLUDE_DIRECTORIES` properties from the osal target and apply them to the entire CFE build as a directory-scope property.

At this time, the OSAL library build does not use/export these properties so this is effectively a no-op for the CFE build and can be merged with no effect.  However, in a future version (specifically after nasa/osal#312 gets fixed), the OSAL library will export these interface properties and this will become important.

Fixes #581 

**Testing performed**
Built for all three test platforms (ppc-vxworks6.9, i686-rtems4.11, native/x86-64 Linux).  Sanity Check CFE build boots and runs, unit tests run.

**Expected behavior changes**
No impact to behavior - build is identical because these properties are currently empty/not set in the current OSAL build.

**System(s) tested on**
Ubuntu 18.04 LTS 64 bit.

**Additional context**
Also locally tested against a version of OSAL+PSP that includes all the latest proposed changes and confirm that this still works.

### DIFF
--- a/cmake/arch_build.cmake
+++ b/cmake/arch_build.cmake
@@ -339,6 +339,20 @@ function(process_arch SYSVAR)
   include_directories(${MISSION_SOURCE_DIR}/cfe/fsw/cfe-core/src/inc)
   include_directories(${MISSION_SOURCE_DIR}/cfe/cmake/target/inc)
     
+  # propagate any OSAL interface compile definitions and include directories to this build
+  # This is set as a directory property here at the top level so it will apply to all code.
+  # This includes MODULE libraries that do not directly/statically link with OSAL but still
+  # should be compiled with these flags.
+  get_target_property(OSAL_COMPILE_DEFINITIONS osal INTERFACE_COMPILE_DEFINITIONS)
+  get_target_property(OSAL_INCLUDE_DIRECTORIES osal INTERFACE_INCLUDE_DIRECTORIES)
+
+  if (OSAL_COMPILE_DEFINITIONS)
+    set_directory_properties(PROPERTIES COMPILE_DEFINITIONS "${OSAL_COMPILE_DEFINITIONS}")
+  endif (OSAL_COMPILE_DEFINITIONS)
+  if (OSAL_INCLUDE_DIRECTORIES)
+    include_directories(${OSAL_INCLUDE_DIRECTORIES})
+  endif (OSAL_INCLUDE_DIRECTORIES)
+
   # Append the PSP and OSAL selections to the Doxyfile so it will be included
   # in the generated documentation automatically.
   # Also extract the "-D" options within CFLAGS and inform Doxygen about these


astrogeco:
CCB 20200408 - APPROVED

skliper:
Rebased on master so it can be tested with the bundle

cmake/arch_build.cmake
```diff
@@ -339,6 +339,20 @@ function(process_arch SYSVAR)
   include_directories(${MISSION_SOURCE_DIR}/cfe/fsw/cfe-core/src/inc)
   include_directories(${MISSION_SOURCE_DIR}/cfe/cmake/target/inc)
     
+  # propagate any OSAL interface compile definitions and include directories to this build
+  # This is set as a directory property here at the top level so it will apply to all code.
+  # This includes MODULE libraries that do not directly/statically link with OSAL but still
+  # should be compiled with these flags.
+  get_target_property(OSAL_COMPILE_DEFINITIONS osal INTERFACE_COMPILE_DEFINITIONS)
```
stanislaw, line 346: I am not sure if I understand the intention of this right but it looks like you could simply control this by specifying in the `osal` target's CMake file:

```
target_include_directories(osal
  PUBLIC
    ...
  INTERFACE
    ...
  PRIVATE
    ...
)
```

and the same with `target_compile_definitions`.

stanislaw: `target_*` with `PUBLIC` and `INTERFACE` specifiers are designed to propagate corresponding definitions to their user targets.

jphickey: That's exactly what the OSAL CMakeLists.txt file is doing in the upcoming update.  However, it only works "implicitly" for targets which directly link with OSAL. For shared library/module targets that do not directly link with OSAL but yet still need to use the headers and a compatible set of definitions, this is why it needs to be done explicitly.

Unless there is another way of doing this for MODULE libraries?

stanislaw: After your explanation I understand it better now.

There are also interface libraries: you could collect the needed flags and definitions of osal and attach them to an interface library `osal-interface`. The library/module targets could then link to the interface library instead of linking to `osal` directly.

But I am not sure if this will simplify existing setup.

https://cmake.org/cmake/help/v3.0/command/add_library.html (`INTERFACE`)

jphickey: Interface libraries are an interesting option... this might actually be a possibility for simplifying the way OSAL manages the flags internally, then we can just put this inside add_cfe_app().  Will not have time to revisit this in the near term however, so hopefully this approach is good enough for now.

skliper: See #626 to track

Issue #263, opened 2019-09-30
expose CFE_SB_IsValidMsgId()


skliper:
CFE apps would benefit from a publicly-available IsValidMsgId() function (or perhaps an expanded "is valid message" function, that would check all header fields?)

For example, SCH has a function to validate its message table entries and has its own logic for determining whether a message ID is valid or not.

This would particularly facilitate CCSDS_VER_2 transitions.

skliper:
_Imported from trac issue 232.  Created by cdknight on 2018-01-16T18:26:34, last modified: 2019-07-03T12:48:08_

skliper:
_Trac comment by jphickey on 2018-05-14 14:35:14:_

Upvote this for next cFE release.

Particularly with the recent addition of extended CCSDS header, it is not necessarily a trivial check to determine if a MsgId value is valid or not. Apps should not be making assumptions about this value, and should only treat it as an opaque value.

Note that #245, if implemented, will turn any comparison test of MsgId values into a compiler error (a good thing).  This would necessitate having a function to call in its place.

skliper:
_Trac comment by dcmccom3 on 2018-11-15 09:19:31:_

In addition to the run-time functional change, we should create a compile-time solution.  Multiple apps use an undefined MID macro for default table values. I looked at three apps(DS, HK, and SCH) and they each define a local undefined MID macro. They're also inconsistent in their definitions. DS & HK use 0 and SCH uses 0x1897.

Some preliminary thoughts:
1. The EDS build tool chain may be the long term solution
2. We need to check whether an app's code makes assumptions about the local definition. For example, simply checking whether a MID is great than the undefined value because the undefined value is assumed to be zero.

skliper:
_Trac comment by jphickey on 2018-11-15 09:38:37:_

Can you elaborate on "compile-time solution"?  There are two possibilities:

- A value to use as a placeholder/initializer for msgids within data structures 
- A method to check if a given msgid is valid.

We also need to be sure apps differentiate between a "MsgId" and "TopicId".  Many places that are checking MIDs would be better served by checking a topic ID instead and I think this will solve a lot of the issue.  

[...]

PR #592, opened 2020-04-08
Fix #245, Opaque CFE_SB_MsgId_t values

### LINKED ISSUE #245: Type safety and improved handling of CFE_SB_MsgId_t values
In 6.6, as we move to supporting MsgId's, MsgKey's, RouteIdx, and other types, we should move away from using native C types and wrapping the types in a struct to prevent accidentally using the wrong type in assignments and function calls.

This will, of course, require re-tooling any existing code that expects the type to be a simple type...

For example, instead of:

```
typedef uint16 CFE_SB_MsgId_t;

uint32 SomeFunction(CFE_SB_MsgId_t MsgId) {
    /* ... */
    return MsgId; /* this will work fine, even though it's bad behavior! */
}
```

Use:

```
typedef uint16 CFE_SB_MsgId_Atom_t;
typedef struct
{
    CFE_SB_MsgId_Atom_t __MsgId; /* private, do not touch */
} CFE_SB_MsgId_t;

uint32 SomeFunction(CFE_SB_MsgId_t MsgId) {
    /* ... */
    return MsgId; /* will now give a compiler error */
}
```

jphickey:
**Describe the contribution**

Ensures that all CFE core apps outside of SB itself treat the `CFE_SB_MsgId_t` type as an opaque/abstract value and do not assume it is an integer.

This pull request is primarily a concept/preview of this idea in action for CCB discussion.  It does work though and could be merged as-is although I expect some change requests. 

Currently the type safe enforcement is enabled based on a compile definition `CFE_SB_OPAQUE_MSGIDS`.  This could be moved into the "omit deprecated" switch instead rather than a new switch.

The type-safe enforcement switches CFE_SB_MsgId_t from being a simple integer value to a structure wrapper, so it cannot be implicitly exchanged with other integers, nor can it be involved in any arithmetic calculation.  However, this is a breaking API change due to the way most apps specify their msgid values as integers via `#define` macros.

Fixes #245 

**Testing performed**
Build CFE for SIMULATION=native (64 bit linux)
Execute all unit tests
Sanity check CFE by booting and sending commands with cmdUtil
In particular check all CFE SB routing info, map info commands and compare file output before and after change to make sure data is the same.

**Expected behavior changes**
No externally-visible changes (i.e. CMD/TLM, files, tables, etc).
API changes when building with enforcing mode:
- All CFE SB API operations that accept or return a CFE_SB_MsgId_t (Subscribe, etc) must be supplied with an instance of `CFE_SB_MsgId_t` only, not an integer.  Passing or assigning to a bare integer triggers a compiler error.
- Macros/conversions are provided to bridge the gap, and allow printing as integer for syslog/events as well as converting from an integer
- No longer can `switch()` on a MsgId -- must use `CFE_SB_MsgId_Equal` and a nested if.
- This also exposes CFE_SB_IsMsgIdValid() in the public API.

**System(s) tested on**
 - Ubuntu 18.04 LTS 64 bit


skliper:
Please separate exposing CFE_SB_IsValidMsgId changes into a different pull request.  It's fixing #263 and a priority for this round of certification.

skliper:
@jwilmot  @acudmore - Requesting a full design review of the entire concept (much bigger than this change) w/ stakeholder involvement, defined requirements, schedule, and committed resources.

astrogeco:
CCB 20200408 - Push to later date, maybe Fall 2020 or later? Requires a longer design discussion involving more stakeholders.

skliper:
Just to document - I'm ok with switching cFE to use the appropriate wrappers/macros (where possible and it makes sense) that already exist in the code.  I'd prefer `CFE_SB_MSGID_LITERAL` or other approaches get discussed at a design level first (and consider phasing vs just going directly to a more permanent solution.)

skliper:
How do we want to handle this one?  Close for now and resubmit when the design has been reviewed and approved by our stakeholders?  My sense is the design will evolve as part of the review, so keeping this pull request open doesn't seem all that beneficial.

jphickey:
The pull request can be closed, but the issue ticket should remain open.

Issue #587, opened 2020-04-07
Add requirements to cover IsValidMsgId and GetPktType


skliper:
**Is your feature request related to a problem? Please describe.**
Exposing CFE_SB_IsValidMsgId and CFE_SB_GetPktType needs to flow from a requirement given the current cFE requirements pattern.

**Describe the solution you'd like**
Add the requirements

**Describe alternatives you've considered**
None

**Additional context**
#263 #543

**Requester Info**
Jacob Hageman - NASA/GSFC

PR #593, opened 2020-04-08
Fix# 543, Expose CFE_SB_GetPktType

### LINKED ISSUE #543: expose CFE_SB_GetPktType (and add stub)
Requests to expose packet type as an API (helps SBN, testing)

Add requirement, expose in API header, add stub, etc.

N/A

**Additional context**
Slightly related to #263

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
**Describe the contribution**
Fix #543 
Note this only exposes and adds the stub.  Requirement (#587)/functional test/coverage test still in work.

**Testing performed**
1. Built with SIMULATION=native and ENABLE_UNIT_TESTS=true
1. Tests passed
1. Ran with cFS-GroundStation and observed reported sequence counter that wasn't reporting actual sequence count
1. Note it does pass the sequence count test, which is what SB uses GetPktType for

**Expected behavior changes**
None

**System(s) tested on**
 - Hardware: cFS Dev Server 3
 - OS: Ubuntu 18.04
 - Versions: Bundle w/ this change

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_sb.h
+++ b/fsw/cfe-core/src/inc/cfe_sb.h
@@ -55,6 +55,15 @@
 
 #define CFE_SB_INVALID_MSG_ID           0xFFFF /**< \brief Initializer for #CFE_SB_MsgId_t values that will not match any real MsgId */
 
+/**
+ * \defgroup CFESBPktTypeDefs cFE SB Packet Type Defines
+ * \{
+ */
+#define CFE_SB_PKTTYPE_INVALID          0      /**< \brief #CFE_SB_GetPktType response if message type can not be determined */
+#define CFE_SB_PKTTYPE_CMD              1      /**< \brief #CFE_SB_GetPktType response for command packets */
+#define CFE_SB_PKTTYPE_TLM              2      /**< \brief #CFE_SB_GetPktType response for telemetry packets */
+/** \} */
+
 /*
 ** Macro Definitions
 */
@@ -1358,6 +1367,20 @@ static inline CFE_SB_MsgId_t CFE_SB_ValueToMsgId(CFE_SB_MsgId_Atom_t MsgIdValue)
 {
     return MsgIdValue;
 }
+
+/*****************************************************************************/
+/**
+ * \brief Identifies packet type given message ID
+ *
+ * Provides the packet type associated with the given message ID
+ *
+ * \return Packet type
+ * \retval #CFE_SB_PKTTYPE_CMD     Command packet type
+ * \retval #CFE_SB_PKTTYPE_TLM     Telemetry packet type
+ * \retval #CFE_SB_PKTTYPE_INVALID Invalid/unknown packet type
+ */
+uint32 CFE_SB_GetPktType(CFE_SB_MsgId_t MsgId);
+
 /**@}*/
 
 #endif  /* _cfe_sb_ */
--- a/fsw/cfe-core/src/sb/cfe_sb_msg_id_util.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_msg_id_util.c
@@ -33,7 +33,9 @@
 **      combination of bits from the primary header SID (StreamId) and the secondary header APID Qualifiers
 **      
 **      Implementation is based on CCSDS Space Packet Protocol 133.0.B-1 with Technical Corrigendum 2, September 2012
-**      The extended secondary header is expected in an upcoming revision of 133.0.B-1
+**      Multi-mission Interoperable extended secondary headers should be registered in Space
+**      Assigned Numbers Authority (SANA). The process for SANA registration is documented in
+**      133.0.B-2. Mission specific headers need not be registered
 **
 **      For  MESSAGE_FORMAT_IS_CCSDS_VER_2 the default setup will combine:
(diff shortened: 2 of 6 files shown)


astrogeco:
CCB 20200408 - Approved with small changes
Change invalid to be zero and valid to be nonzeros.

fsw/cfe-core/src/inc/cfe_sb.h
```diff
@@ -55,6 +55,15 @@
 
 #define CFE_SB_INVALID_MSG_ID           0xFFFF /**< \brief Initializer for #CFE_SB_MsgId_t values that will not match any real MsgId */
 
+/**
+ * \defgroup CFESBPktTypeDefs cFE SB Packet Type Defines
+ * \{
+ */
+#define CFE_SB_PKTTYPE_CMD              1      /**< \brief #CFE_SB_GetPktType response for command packets */
+#define CFE_SB_PKTTYPE_TLM              0      /**< \brief #CFE_SB_GetPktType response for telemetry packets */
+#define CFE_SB_PKTTYPE_INVALID          0xFF   /**< \brief #CFE_SB_GetPktType response if message type can not be determined */
```
jphickey, line 64: As mentioned in CCB - For enumerated values like this we should keep zero as the reserved/invalid value (e.g. in case one of these values is stored in memory which is memset to zero, that should alias to `INVALID`, not `TLM`.

fsw/cfe-core/src/inc/cfe_sb.h
```diff
@@ -1358,6 +1367,21 @@ static inline CFE_SB_MsgId_t CFE_SB_ValueToMsgId(CFE_SB_MsgId_Atom_t MsgIdValue)
...
+ *
+ * Provides the packet type associated with the given message ID
+ * as defined by #CFESBPktTypeDefs
+ *
+ * \return Packet type
+ * \retval #CFE_SB_PKTTYPE_CMD     Command packet type
+ * \retval #CFE_SB_PKTTYPE_TLM     Telemetry packet type
+ * \retval #CFE_SB_PKTTYPE_INVALID Invalid/unknown packet type
+ */
+uint8 CFE_SB_GetPktType(CFE_SB_MsgId_t MsgId);
```
jphickey, line 1383: Note there is no real benefit to using a smaller size than 32 bits in most cases, as we generally don't target CPUs under 32 bits.  

When comparing to a literal value the compiler will implicitly expand this return value to `int` when evaluating the equality test, so using the smaller `uint8` type may generate a few _more_ machine instructions when comparing this value (depends on arch and optimization).

fsw/cfe-core/src/sb/cfe_sb_msg_id_util.c
```diff
@@ -185,3 +185,27 @@ void CFE_SB_SetMsgId(CFE_SB_MsgPtr_t MsgPtr,
...
+uint8 CFE_SB_GetPktType(CFE_SB_MsgId_t MsgId)
+{
+
+  CFE_SB_MsgId_Atom_t Val = MsgId;
+  uint8 PktType;
+
+  if (CFE_SB_IsValidMsgId(MsgId))
+  {
+#ifndef MESSAGE_FORMAT_IS_CCSDS_VER_2
+    PktType = (CFE_TST(Val,12) == true) ? CFE_SB_PKTTYPE_CMD : CFE_SB_PKTTYPE_TLM;
```
jphickey, line 201: The extra `== true` here is redundant as the `CFE_TST()` macro already results in a boolean.  But it is potentially problematic if the definition of `CFE_TST()` macro changes as it only tests for one possible `true` value.  Just testing the value of `CFE_TST(Val,12)` directly is a little more future proof.

This is also (I believe) the only use of ternary conditional operator in our runtime codebase.  Not that I have a problem with it, but I think it was avoided out of concern that some compilers might not support it.  AFAIK it is standard since C99, but there is a (slight) risk that some compiler might have an issue with it.

skliper:
Updated per requests.  

Ping @tngo67

skliper:
Standby... screwed up the changes

skliper:
Now ready (last one changes didn't fit commit topic, now fixed).

skliper:
Nope... one more change pending...

skliper:
Fixed stub leftover cast.

skliper:
Fixed doxygen warning

astrogeco:
@skliper do we need to wait for @tngo67 to review?

tngo67:
> 
> 
> @skliper do we need to wait for @tngo67 to review?

I'm good with the changes.

jwilmot:
Looks good, but the comment "The extended secondary header is expected in an upcoming revision of 133.0.B-1" in cfe_sb_msg_id_util should be changed to "Multi-mission Interoperable extended secondary headers should be registered in Space Assigned Numbers Authority (SANA). The process for SANA registration is documented in 133.0.B-2. Mission specific headers need not be registered"

skliper:
Updated standards comment per @jwilmot request.  Should be ready for IC.

skliper:
Squashed

skliper:
More descriptive commit...

PR #595, opened 2020-04-09
Fix #456, Perf log threading and concurrency issues

### LINKED ISSUE #456: CFE_ES_PerfLogAdd needs better mutual exclusion
**Describe the bug**
The `CFE_ES_PerfLogAdd()` routine uses `OS_IntLock()` in an attempt to get exclusive access to a common global data structure to record performance metrics/state.  This is insufficient.

The `OS_IntLock` function is generally not implemented on SMP, and even if it is, it probably only affects the current core.  Either way, it will **not** provide exclusivity, because the other cores can still access the data even when interrupts are disabled.  

This function is also a no-op in the POSIX OSAL.

**To Reproduce**
Enable performance monitoring on a POSIX system and observe that occasionally samples occur in the log out of order or otherwise appear corrupted.  This is likely due to concurrent writes to the same entry related to insufficient locking.

**Expected behavior**
The function should use some form of primitive that actually does provide exclusivity between threads (such as a mutex/spinlock) and not an interrupt lock.

**Code snips**
https://github.com/nasa/cFE/blob/2b27dfc8f2a4e0627e6c4b646765269213584cda/fsw/cfe-core/src/es/cfe_es_perf.c#L439-L440

**System observed on:**
Ubuntu 18.04 LTS 64-bit

**Reporter Info**

jphickey:
**Describe the contribution**

Replaces the `OS_IntLock` with a standard OSAL mutex for protecting the shared/global perflog data structure.  This may introduce unexpected task switches when contention occurs, but it ensures proper exclusion with respect to the data structures.

Removes the temporary child worker task that was spawned for writing the log data to a file, and replace with a more generic CFE ES background task.  The background task is started at boot and
pends on a semaphore until there is work to do.

The background performance log dump is implemented as a state machine which is called repeatedly over time from the background job task.  This performs a limited amount of work on each invocation, and resumes where it left from the previous invocation.

Fixes #456 

**Testing performed**
Build CFE with ENABLE_UNIT_TESTS=TRUE
Confirm all unit tests pass
Confirm near 100% coverage on all newly added/modified code
Run CFE and send commands to start performance logging
Send other CFE commands to generate performance log activity
Send command to stop performance log and generate a dump file
Confirm validity of dump file by opening with Java tool.  No errors reported when opening file.

**Expected behavior changes**
No impact to behavior.  Previously the perf log dump file frequently contained errors due to out of order or otherwise corrupted entries, which is now fixed.

**System(s) tested on**
 - Ubuntu 18.04 LTS 64-bit (native)
- MIPS 32 target running in QEMU (big endian)

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_backgroundtask.c
+++ b/fsw/cfe-core/src/es/cfe_es_backgroundtask.c
@@ -1,0 +1,250 @@
+/*
+**  GSC-18128-1, "Core Flight Executive Version 6.7"
+**
+**  Copyright (c) 2006-2019 United States Government as represented by
+**  the Administrator of the National Aeronautics and Space Administration.
+**  All Rights Reserved.
+**
+**  Licensed under the Apache License, Version 2.0 (the "License");
+**  you may not use this file except in compliance with the License.
+**  You may obtain a copy of the License at
+**
+**    http://www.apache.org/licenses/LICENSE-2.0
+**
+**  Unless required by applicable law or agreed to in writing, software
+**  distributed under the License is distributed on an "AS IS" BASIS,
+**  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+**  See the License for the specific language governing permissions and
+**  limitations under the License.
+*/
+
+/*
+** File: cfe_es_backgroundtask.c
+**
+** Purpose: This file contains the implementation of the ES "background task"
+**
+** This task sits idle most of the time, but is woken by the ES application
+** for various maintenance duties that may take time to execute, such as
+** writing status/log files.
+**
+*/
+
+/*
+** Include Section
+*/
+
+#include <string.h>
+
+#include "osapi.h"
+#include "private/cfe_private.h"
+#include "cfe_es_perf.h"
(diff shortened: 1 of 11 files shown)


skliper:
@ejtimmon Is this design change an impact on the training slides?  Might be worth a state diagram and a couple bullets...

jphickey:
It should not change the perf log feature from a user standpoint.  One still sends the same command to start logging, set the trigger, and stop logging, and then the dump file appears.

The only thing that a user might care about is that `CFE_ES_PerfLogAdd()` _might_ cause a task switch because it now uses a mutex.  This makes it less deterministic.  Of course this is only a risk when using the perf log (i.e. no chance of this happening until you send the start command).

skliper:
I understand it's not a user impact, which is why I'm not requesting user's guide updates.  We are required to maintain current design documentation, and this is a significant design change.

astrogeco:
CCB Fast Track - Approved, @skliper should we open a ticket to address your documentation comment?

skliper:
cFE design documentation isn't currently in the repo, so more like an action item.  I'm suggesting a slide or two overview in the context of ES covering the new background task and work it performs.

PR #598, opened 2020-04-09
Fix #295, Resolve app table scanning race conditions

### LINKED ISSUE #295: CFE_ES_ScanAppTable() possible race conditions
The `CFE_ES_ScanAppTable()` is called from the ES message processing thread, but it does not lock the global ES data structure when reading/writing from the global data.

ES software bus command messages are safe because this function is called by the same thread that is processing those messages, therefore concurrency is not possible here.

But other functions, like `CFE_ES_DeleteApp()` and `CFE_ES_ReloadApp()` are part of the public API and these also update the same fields within the app state data structure that `CFE_ES_ScanAppTable()` is reading.  `CFE_ES_RunLoop()` can also modify fields within this structure and this is called by pretty much every app in the system.

This issue was noted while reviewing the fix for a similar issue in #279.

### LINKED ISSUE #480: Potential for app exit failure when not passing back RunStatus
**Describe the bug**
CFE_ES_ExitApp reports error when passed in ExitStatus is CFE_ES_RunStatus_APP_RUN.

Typical app pattern (see https://github.com/nasa/sample_app/blob/master/fsw/src/sample_app.c) is:
````
  while (CFE_ES_RunLoop(&Sample_AppData.RunStatus) == TRUE) {do stuff}
 
  CFE_ES_ExitApp(Sample_AppData.RunStatus);
````

But CFE_ES_RunLoop does not update RunStatus on internal request to stop:
https://github.com/nasa/cFE/blob/7251cbccc79574a897ebd7cae3ef1975cfaedcec/fsw/cfe-core/src/es/cfe_es_api.c#L545-L552

**To Reproduce**
1. Send RestartApp, will error and fail to restart.

**Expected behavior**
Set passed in RunStatus to the control request for the case above:
RunStatus = CFE_ES_Global.AppTable[AppID].ControlReq.AppControlRequest

Allows the App to take appropriate action.

**System observed on:**
 - cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: current dev (6.7.4)

**Additional context**
Fails build verification testing

**Reporter Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**

Fixes #295 
Fixes #480 

Create a new background job to handle the maintenance tasks that had been performed in the ES main task as part of the `CFE_ES_ScanAppTable()` routine.

All app state changes, including those invoked by messages, are now handled by this job.
    
This also slightly changes the semantics of CFE_ES_RunLoop and CFE_ES_ExitApp.  Now, the CFE_ES_RunLoop routine no longer requires a RunStatus buffer.  Instead, the only thing that matters is the RunStatus value that is eventually passed to CFE_ES_ExitApp after the shutdown is complete.  This should be mostly backward compatible, as the recommended app pattern would pass the same value to both functions.
    
This commit also fixes #480, as the value passed to CFE_ES_ExitApp will not override a request that was already pending.

**Testing performed**
Build with ENABLE_UNIT_TESTS=TRUE for native (x86-64 Linux) and RTEMS (i686-rtems4.11)
- Run unit tests and confirm all passing
- Check coverage report (lcov) and confirm no reduction in coverage for modified functions, and newly added functions are fully covered.
- Run CFE and send various restart/reload/delete requests for SAMPLE_APP
- Confirm that SAMPLE_APP exits itself and is restarted/reloaded/deleted correctly by ES.
- Confirm that if SAMPLE_APP does not exit itself, it is still forcibly deleted by ES
- Confirm same behavior on RTEMS

**Expected behavior changes**
No impact to CMD/TLM interface or API.  Internal changes only.

**System(s) tested on**
- Ubuntu 18.04 LTS 64 bit (native)
- QEMU (i686-rtems4.11)

**Additional context**
This reuses the background task concept introduced in pull #595.  Will rebase this changeset as soon as that pull request makes it into a stable baseline.

Joseph Hickey, Vantage Systems Inc.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_api.c
+++ b/fsw/cfe-core/src/es/cfe_es_api.c
@@ -19,10 +19,10 @@
 */
 
 /*
-**  File:  
+**  File:
 **    cfe_es_api.c
 **
-**  Purpose:  
+**  Purpose:
 **    This file implements the cFE Executive Services API functions.
 **
 **  References:
@@ -73,7 +73,7 @@ int32 CFE_ES_GetResetType(uint32 *ResetSubtypePtr)
 int32 CFE_ES_ResetCFE(uint32 ResetType)
 {
     int32 ReturnCode;
-    
+
     if ( ResetType == CFE_PSP_RST_TYPE_PROCESSOR )
     {
        /*
@@ -82,10 +82,10 @@ int32 CFE_ES_ResetCFE(uint32 ResetType)
        CFE_ES_ResetDataPtr->ResetVars.ProcessorResetCount++;
 
        /*
-       ** Before doing a Processor reset, check to see 
+       ** Before doing a Processor reset, check to see
        ** if the maximum number has been exceeded
        */
-       if ( CFE_ES_ResetDataPtr->ResetVars.ProcessorResetCount > 
+       if ( CFE_ES_ResetDataPtr->ResetVars.ProcessorResetCount >
             CFE_ES_ResetDataPtr->ResetVars.MaxProcessorResetCount )
        {
            CFE_ES_WriteToSysLog("POWER ON RESET due to max proc resets (Commanded).\n");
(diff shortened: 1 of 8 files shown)


jphickey:
Force-push: rebased to latest master branch

astrogeco:
CCB 20200415 - APPROVED

Issue #600, opened 2020-04-10
CFE_TBL_Load() and CFE_TBL_LoadCmd() overlap


CDKnightNASA:
**Is your feature request related to a problem? Please describe.**
There's code commonalities between CFE_TBL_LoadCmd() and CFE_TBL_Load().

**Describe the solution you'd like**
Code should be consolidated betwen CFE_TBL_Load() and CFE_TBL_LoadCmd().

**Requester Info**
Christopher.D.Knight@nasa.gov

CDKnightNASA:
Pref. just have LoadCmd() call Load()?

CDKnightNASA:
should wait for #503 to be merged into master.

CDKnightNASA:
> I'd like to revive/review this again for the Equuleus cycle. Part of the planned feature set for this release does potentially require an update to the `CFE_TBL_Load` function, and thus becomes important to break this function up into smaller and more manageable parts to reduce its complexity.

Do we want to make major API-breaking redesign changes as part of this? Perhaps we can create a top-level issue and link other issues that we want to implement/fix as part of this redesign?

PR #601, opened 2020-04-10
fix #503 - CFE_TBL_Load() errors - strawman for CCB consideration

### LINKED ISSUE #503: CFE_TBL_Load should produce clearer message when it has a file header mismatch
CFE_TBL_LoadFromFile compares the application's name + table name with the string in the header of the file. If you configure the cfe_es_startup.scr with a different application name than what is expected in the header, you get an obtuse "Fail to load Tbl '<app>.<tbl>' from '/cf/<file>.tbl'" not making it clear that it's purely a mismatch of the header.

At minimum it should indicate that the matter is a mis-match of the header and the expected header (which means that the file is there, for example). Better would be to report the actual contents of the header and the expected value. The EID for the error is unique so CFE_TBL_Load can produce an event with more specific text.

Could have CFE_TBL_LoadFromFile generate the event, but that's an "internal" fn so probably not.

**Additional context**
Add any other context about the feature request here.

**Requester Info**
Christopher.D.Knight@nasa.gov

CDKnightNASA:
**Describe the contribution**
Moved CFE_TBL_LoadFromFile() code into CFE_TBL_Load(), lots new EID's, some new status codes, removed SysLog calls from Load()

**Testing performed**
Builds and runs, just no UT code changes yet.

**Expected behavior changes**
CFE_TBL_Load() now produces events instead of writing to syslog (and generating a generic event). Also removed CFE_TBL_LoadFromFile() as it was only called by CFE_TBL_Load() (and the UT code, which will have to be adapted to call Load().)

**System(s) tested on**
Debian 9

Christopher.D.Knight@nasa.gov


fsw/cfe-core/src/tbl/lff
```diff
@@ -0,0 +1,90 @@
+    int32                Status = CFE_SUCCESS;
```
skliper, line 1: Looks like a stray file...

fsw/cfe-core/src/tbl/cfe_tbl_api.c
```diff
@@ -761,19 +766,157 @@ int32 CFE_TBL_Load( CFE_TBL_Handle_t TblHandle,
...
-                        /* Load the data from the file into the specified buffer */
-                        Status = CFE_TBL_LoadFromFile(WorkingBufferPtr, RegRecPtr, (const char *)SrcDataPtr);
+                        CFE_FS_Header_t      StdFileHeader;
+                        CFE_TBL_File_Hdr_t   TblFileHeader;
+                        int32                FileDescriptor;
+                        const char *         Filename = SrcDataPtr;
+                        size_t               FilenameLen = strlen(Filename);
+                        uint32               NumBytes;
+                        uint8                ExtraByte;
+
```
skliper, line 776: This is against our code standard (declare variables up front).

Curious about what the CCB thinks about this approach.  Coding standard suggest shorter functions if I recall correctly, vs 1 big function.

CDKnightNASA: It's 363 lines, far shorter than CFE_TBL_Register() [450 lines] and similar to CFE_SB_SendMsgFull() [321 lines]...It is possible that it will be split out again when merging the code with CFE_TBL_LoadCmd() (see #600 ). But I wanted to simplify code and tests to address this ticket, then approach LoadCmd() separately.

If there's objection, I can move back to using an external function. HOWEVER, I definitely do not like the logic in Load() and would keep this simpler, more direct logic.

The old code calls LoadFromFile() and checks the return code ( https://github.com/nasa/cFE/blob/60a5f6580d1e737f46efce0418aba5b8dac6df6a/fsw/cfe-core/src/tbl/cfe_tbl_api.c#L753 ) and generate a SysLog event sometimes, then at the end of Load() it generates some sort of event ( https://github.com/nasa/cFE/blob/60a5f6580d1e737f46efce0418aba5b8dac6df6a/fsw/cfe-core/src/tbl/cfe_tbl_api.c#L896 ). So event generation is about as far away from error detection as possible. :o

I'd like to also discuss with the CCB the validity of functional testing code that pokes at internal API's. We do it everywhere (CFE_SB_SendMsgFull(), for example) but it seems like not true black-box testing. I'd rather change the tbl_UT to only call CFE_TBL_Load() and not whatever internal function is behind it.

skliper: CFE_TBL_Register is a cyclomatic complexity of 49 CFE_SB_SendMsgFull is 32.  The recommendation for flight code is 10 (https://swehb.nasa.gov/display/SWEHBVC/SWE-134+-+Safety+Critical+Software+Design+Requirements).  Any new/refactored code should keep this number as a goal, and avoid raising complexity where possible.

Simpler logic is definitely preferred.

Note tbl_UT primary goal is white box testing (coverage and unit function).  Functional/black box/requirements verification is through the build verification tests and test apps (black box).

CDKnightNASA: Copy all, I'll close this pull request and refactor and resubmit.

PR #578, opened 2020-04-01
Fix #240, removes MESSAGE_FORMAT_IS_CCSDS ifdefs from code

dmknutsen:
**Describe the contribution**
Fixes #240, removes MESSAGE_FORMAT_IS_CCSDS ifdefs from CFS code

Testing performed
Ran unit tests

System(s) tested on
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cFE 6.7.10.0, OSAL 5.0.8.0, PSP 1.4.7.0,

Contributor Info
Dan Knutsen
NASA/Goddard

PR #602, opened 2020-04-10
Fix #263, expose IsValidMsgID and partially address #245 Consistent use of MsgId types

### LINKED ISSUE #263: expose CFE_SB_IsValidMsgId()

jphickey:
**Describe the contribution**

Fixes #263

This exposes the `CFE_SB_IsValidMsgId()` in the CFE SB public API, rather than confining it to the private/internal API.

Partially addresses #245 

This also includes a number of internal changes to make CFE core apps consistent in their use of the CFE_SB_MsgId type and with the CFE SB API.  This employs the appropriate CFE SB abstraction API whenever any of the following needs to happen:

- Use of a CFE_SB_MsgId_t value within a printf (event, syslog, etc).
- Initialization of a CFE_SB_MsgId_t from an integer value
- Comparison of two CFE_SB_MsgId_t values
- Checking if a CFE_SB_MsgId_t value is within the valid set

Notable exception to integer conversion rules is for the MID `#define` values, which are not changed for backward compatibility reasons.  These are still defined as integers, and used directly in the code.

Also worth noting - this replaces the switch/case constructs used in message dispatch with a nested if/else based on CFE_SB_MsgId_Equal(), as the switch statement can only be used with an integer.

A few new macros are introduced, mainly because the inline functions that already existed for this purpose cannot be used where evaluation must be done at compile time (e.g. constants, struct initialization).  These are initially just typecasts, but could become more interesting
in future revisions.

**Testing performed**
Build with SIMULATION=native ENABLE_UNIT_TESTS=TRUE
- Run all unit tests and confirm passing
- Run sanity check on CFE, boot up and issue commands to each core app using `cmdUtil`.  Ensure no changes to message routing/processing for every core app.

**Expected behavior changes**
This exposes the `CFE_SB_IsValidMsgId()` for application usage.
No other API or behavior changes.  All other changes are transparent.

**System(s) tested on**
Ubuntu 18.04 LTS 64 bit

**Additional context**
This is a step in the direction of related issue #245 but does not do anything that would affect backward compatibility with the way the MID integers are currently defined and used.  It does introduce a few of the macros that will become more relevant/necessary (i.e. so they can evolve beyond a simple typecast/passthrough).

*EDIT* added reference for MsgId handling

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_task.c
+++ b/fsw/cfe-core/src/es/cfe_es_task.c
@@ -251,23 +251,30 @@ int32 CFE_ES_TaskInit(void)
     /*
     ** Initialize housekeeping packet (clear user data area)
     */
-    CFE_SB_InitMsg(&CFE_ES_TaskData.HkPacket, CFE_ES_HK_TLM_MID, sizeof(CFE_ES_TaskData.HkPacket), true);
+    CFE_SB_InitMsg(&CFE_ES_TaskData.HkPacket,
+            CFE_SB_ValueToMsgId(CFE_ES_HK_TLM_MID),
+            sizeof(CFE_ES_TaskData.HkPacket), true);
 
     /*
     ** Initialize shell output packet (clear user data area)
     */
-    CFE_SB_InitMsg(&CFE_ES_TaskData.ShellPacket, CFE_ES_SHELL_TLM_MID, sizeof(CFE_ES_TaskData.ShellPacket), true);
+    CFE_SB_InitMsg(&CFE_ES_TaskData.ShellPacket,
+            CFE_SB_ValueToMsgId(CFE_ES_SHELL_TLM_MID),
+            sizeof(CFE_ES_TaskData.ShellPacket), true);
 
     /*
     ** Initialize single application telemetry packet
     */
-    CFE_SB_InitMsg(&CFE_ES_TaskData.OneAppPacket, CFE_ES_APP_TLM_MID, sizeof(CFE_ES_TaskData.OneAppPacket), true);
+    CFE_SB_InitMsg(&CFE_ES_TaskData.OneAppPacket,
+            CFE_SB_ValueToMsgId(CFE_ES_APP_TLM_MID),
+            sizeof(CFE_ES_TaskData.OneAppPacket), true);
 
     /*
     ** Initialize memory pool statistics telemetry packet
     */
-    CFE_SB_InitMsg(&CFE_ES_TaskData.MemStatsPacket, CFE_ES_MEMSTATS_TLM_MID,
-                   sizeof(CFE_ES_TaskData.MemStatsPacket), true);
+    CFE_SB_InitMsg(&CFE_ES_TaskData.MemStatsPacket,
+            CFE_SB_ValueToMsgId(CFE_ES_MEMSTATS_TLM_MID),
+            sizeof(CFE_ES_TaskData.MemStatsPacket), true);
 
     /*
     ** Create Software Bus message pipe
@@ -282,7 +289,7 @@ int32 CFE_ES_TaskInit(void)
     /*
     ** Subscribe to Housekeeping request commands
     */
-    Status = CFE_SB_SubscribeEx(CFE_ES_SEND_HK_MID, CFE_ES_TaskData.CmdPipe,
+    Status = CFE_SB_SubscribeEx(CFE_SB_ValueToMsgId(CFE_ES_SEND_HK_MID), CFE_ES_TaskData.CmdPipe,
                                 CFE_SB_Default_Qos, CFE_ES_TaskData.LimitHK);
     if ( Status != CFE_SUCCESS )
     {
@@ -293,8 +300,8 @@ int32 CFE_ES_TaskInit(void)
     /*
     ** Subscribe to ES task ground command packets
     */
-    Status = CFE_SB_SubscribeEx(CFE_ES_CMD_MID, CFE_ES_TaskData.CmdPipe,
-                       CFE_SB_Default_Qos, CFE_ES_TaskData.LimitCmd);
+    Status = CFE_SB_SubscribeEx(CFE_SB_ValueToMsgId(CFE_ES_CMD_MID), CFE_ES_TaskData.CmdPipe,
+                                CFE_SB_Default_Qos, CFE_ES_TaskData.LimitCmd);
     if ( Status != CFE_SUCCESS )
     {
         CFE_ES_WriteToSysLog("ES:Cannot Subscribe to ES ground commands, RC = 0x%08X\n", (unsigned int)Status);
@@ -427,7 +434,7 @@ void CFE_ES_TaskPipe(CFE_SB_MsgPtr_t Msg)
     uint16         CommandCode;
 
     MessageID = CFE_SB_GetMsgId(Msg);
-    switch (MessageID)
+    switch (CFE_SB_MsgIdToValue(MessageID))
     {
         /*
         ** Housekeeping telemetry request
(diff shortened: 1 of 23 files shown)


fsw/cfe-core/src/inc/cfe_sb.h
```diff
@@ -53,7 +53,54 @@
...
-#define CFE_SB_INVALID_MSG_ID           0xFFFF /**< \brief Initializer for #CFE_SB_MsgId_t values that will not match any real MsgId */
+/* ------------------------------------------------------ */
+/* Macro Constants for use with the CFE_SB_MsgId_t type   */
+/* ------------------------------------------------------ */
+
+/**
+ * \brief Translation macro to convert from MsgId integer values to opaque/abstract API values
+ *
+ * This conversion exists in macro form to allow compile-time evaluation for constants, and
+ * should not be used directly in application code.
```
skliper, line 64: I'd rather thse be in a separate file if it's not intended for application use, since this file is pulled into the API by doxygen.

jphickey: Yeah, the problem is the same logic is needed by CFE_SB_MsgIdToValue/ValueToMsgId and these are inline functions within the same file, so the macro needs to be available.

By "application code' really I mean runtime -- the reason these need to exist in macro form is because one cannot use the inline function when initializing data at compile-time.  I want to encourage use of the inline functions whenever possible because its more type safe, but occasionally one does need an expression that evaluates at compile-time, and its those times that the macro needs to be used instead.

skliper: Still could be in different file and just include, which provides a clear separation.

jphickey: But there is a possible valid case for using it externally, where one actually needs a value that evaluates at compile time.  

The _preference_ should always be to use the inline function, but there are some places (e.g. global variable init) where they don't work.

As there are no applications using this and the decision isn't made yet as to how MsgIds will be defined in the future, I suppose I could put this into a new file in the `private/` subdir if that helps, but that seems like overkill to make a new file for just this.

But No matter how we define MID's in the future, I would think we'd need something that can be evaluated at compile time....

jphickey: What if the documentation was clarified to indicate "application runtime code" rather than just "application code"?

skliper: Eh, from a naive user point of view I wouldn't know the difference.  No biggie, I just don't like huge header files to begin with, especially containing things that have different uses.  Wouldn't have to be in private, just prefer separation.

fsw/cfe-core/src/es/cfe_es_task.c
```diff
@@ -427,214 +427,210 @@ void CFE_ES_TaskPipe(CFE_SB_MsgPtr_t Msg)
     uint16         CommandCode;
 
     MessageID = CFE_SB_GetMsgId(Msg);
-    switch (MessageID)
+    if (CFE_SB_MsgId_Equal(MessageID, CFE_ES_SEND_HK_MID))
```
skliper, line 430: Why not leave the swtich, and just CFE_SB_MsgIdToValue?  I'd prefer a one line change vs the approach shown here.  Also if CFE_SB_MsgId_Equal ever gets more complex, the approach here seems like it could lead to inefficiencies, and doesn't really solve the bigger design issue of hardcoded values.  I recommend keeping changes simple/minimal until the full MsgId/SB updated design concepts get reviewed and approved by the major stakeholders.

jphickey: A previous version of this change did that, but that implies/requires that the MsgId is always/completely representable as a single integer, which is part of the problem this is trying to address.  Note that extended header fields already exceed 32 bits for the full address range, so even a 32-bit integer msgid is insufficient to fully support current standards.

Admittedly it makes a lot more sense if the literals are also type-correct.  They are still integers here.

jphickey: This was also a reason I asked this question: https://github.com/nasa/cFE/issues/263#issuecomment-612053991 before submitting this pull request.

FWIW, The CFE_SB_MsgIdToValue is also not good for the same reason, but its use should be confined to syslog/events for now.  Someday should consider replacing it with a custom print routine (MsgIdToString maybe) for this purpose.

skliper: Still sound like an attempt to solve a bigger issue prior to the necessary design reviews, trades, and stakeholder buy-in.  I don't see the issue you reference, MsgId may not always be completely represented as a single integer, documented as an issue and evaluated/discussed prior to this change.   I'd rather this just focus on type correctness (use the conversions where appropriate) for now.  Alternatively, split exposing IsValidMsgId from this change if it's going to require more discussion.

skliper: Sounds like this is going to be a bigger discussion.  Please split out exposing IsValidMsgId.

jphickey: OK, I can change that back to the other version.  Again, this was why I posed the question before even testing and submitting this pull, I thought we were OK with this.

TBH, Getting rid of the local dispatch routine in favor of a common approach (as noted in https://github.com/nasa/cFE/issues/263#issuecomment-612144320) is the best approach anyway, so neither of these is really a long term/recommended pattern anyway.

skliper: I should have been more clear "good to go" does not mean approved pull request and ready to merge to master.

jphickey: Understood on that, but this was a case of confirming what to include/not include prior to submission, not approval of a pull request before it was submitted.  

I had preemptively asked specifically about whether to include this switch() change as there is a time cost associated with re-doing and re-validating the pull request again.  I will change it back, but this back-and-forth is precisely what I was hoping to avoid by posing the question on content ahead of time.

skliper: If you are concerned about the time you are spending on the MsgId type safety work, then postpone it as I originally requested.  It is not a priority this cycle, and any real changes that happen prior to a full design review with stakeholder buy-in are at risk of being wasted work anyways.

jphickey:
NOTE: commit  67eaf73 is only for a minor cleanup in unit test file sb_UT.c.  Didn't realize this was re-reviewed since my last push, but it is largely the same thing.  I had just mistakenly pushed to my fork a little earlier than I should have.

At any rate, this version basically assumes that the `#define` MID values will _always_ be integers in nature, not necessarily `CFE_SB_MsgId_t` values.  This means that the old `switch` statements can stay with minor update, because the `case` labels will be integers.  However, it also means that the `CFE_SB_ValueToMsgId` function or `CFE_SB_MSGID_WRAP_VALUE` macro needs to be applied in a lot more places than it did in the previous proposal.  

Nonetheless, this should be (hopefully) good to go now.

fsw/cfe-core/src/es/cfe_es_task.c
jphickey: It is not big time-wise because that version of the change already (mostly) existed in a previous branch.  Its just the time to rebase/merge and re-test.  I do think its critical to get his moving in the right direction WRT MsgId values which is arguably very broken right now so I wouldn't want to drop this entirely.  My frustration was only because  on Friday (I thought) I confirmed that replacing the `switch()` was OK to include and today it was not acceptable.  Nonetheless I'll call this conversation resolved as the 67eaf73 should be OK.

skliper:
I recommend changing the issue reference to #245 (partial fix) related to the MsgId changes, and at minimum a separate commit for #263

jphickey:
Same basic changeset, in two commits now.

astrogeco:
CCB 20200415 - APPROVED

astrogeco:
@jphickey can you work with @dmknutsen and resolve the conflicts between this #578 and #635 ?

jphickey:
Manually merged PR #635 and #602, resolved conflicts, and pushed to integration-candidate

PR #605, opened 2020-04-13
Fix #494, Updates CFE_SB_GetLastSenderID to check if message has been sent on pipe

### LINKED ISSUE #494: CFE_SB_GetLastSenderID will crash if if called before message sent on pipe
CFE_SB_GetLastSenderID assumes that CFE_SB.PipeTbl[PipeId].CurrentBuff is not NULL. Upon CFE_SB_CreatePipe, this will be null. CurrentBuff is only set upon receiving a message in CFE_SB_RcvMsg(). So, if CFE_SB_GetLastSenderID is called before receiving a message the program will crash.

This is Steven Seeger from GSFC. Guess I have a personal account on github. :)

dmknutsen:
**Describe the contribution**
Updates CFE_SB_GetLastSenderID to check if a message has been sent on the associated pipe prior to setting the receivers pointer to the associated address of 'sender' in the buffer descriptor.

Includes associated unit test addition to cover the additional path added to CFE_SB_GetLastSenderID.

Includes an addition of a SB status code  and event ID.

Fix #494

**Testing performed**
1. Ran unit tests.
2. Reviewed coverage test results to ensure that additional path added is covered with the updated unit test.

**Expected behavior changes**
CFE_SB_GetLastSenderID will now detect if it is being called prior to a message being sent on a given pipe.

System(s) tested on
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cFE 6.7.11.0, OSAL 5.0.9.0, PSP 1.4.7.0

Contributor Info
Dan Knutsen
NASA/Goddard

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -1651,12 +1651,20 @@ uint32  CFE_SB_GetLastSenderId(CFE_SB_SenderId_t **Ptr,CFE_SB_PipeId_t  PipeId)
     /* Get ptr to buffer descriptor for the last msg received on the given pipe */
     Ptr2BufDescriptor = CFE_SB.PipeTbl[PipeId].CurrentBuff;
 
-    /* Set the receivers pointer to the adr of 'Sender' struct in buf descriptor */
-    *Ptr = (CFE_SB_SenderId_t *) &Ptr2BufDescriptor -> Sender;
-
-    CFE_SB_UnlockSharedData(__func__,__LINE__);
-
-    return CFE_SUCCESS;
+    if ( Ptr2BufDescriptor == NULL  )
+    {
+        *Ptr = NULL;
+        CFE_SB.PipeTbl[PipeId].LastSender = CFE_SB_INVALID_MSG_ID;
+        CFE_SB_UnlockSharedData(__func__,__LINE__);
+        return CFE_SB_NO_MSG_RECV; 
+    }
+    else
+    {
+        /* Set the receivers pointer to the adr of 'Sender' struct in buf descriptor */
+        *Ptr = (CFE_SB_SenderId_t *) &Ptr2BufDescriptor -> Sender;
+        CFE_SB_UnlockSharedData(__func__,__LINE__);
+        return CFE_SUCCESS;
+    }
 
 }/* end CFE_SB_GetLastSenderId */
 
--- a/fsw/cfe-core/src/inc/cfe_error.h
+++ b/fsw/cfe-core/src/inc/cfe_error.h
@@ -980,6 +980,16 @@
 
 
 /**
+ * @brief No Message Recieved
+ *
+ *  When trying to determine the last senders ID, this return 
+ *  value indicates that there was not a message recived on the pipe.
+ *
+ */
+#define CFE_SB_NO_MSG_RECV       ((int32)0xca00000f)
+
+
+/**
  * @brief Not Implemented
  *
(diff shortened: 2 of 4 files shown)


fsw/cfe-core/src/sb/cfe_sb_api.c
```diff
@@ -1651,12 +1651,23 @@ uint32  CFE_SB_GetLastSenderId(CFE_SB_SenderId_t **Ptr,CFE_SB_PipeId_t  PipeId)
...
+        CFE_SB_UnlockSharedData(__func__,__LINE__);
+        return CFE_SB_NO_MSG_RECV; 
+    }
+    else
+    {
+        /* Set the receivers pointer to the adr of 'Sender' struct in buf descriptor */
+        *Ptr = (CFE_SB_SenderId_t *) &Ptr2BufDescriptor -> Sender;
+        CFE_SB_UnlockSharedData(__func__,__LINE__);
+        return CFE_SUCCESS;
+    }
```
jphickey, line 1670: Actually, this API is fundamentally broken.  I had never really looked at it before.

The content of `CFE_SB.PipeTbl[PipeId].CurrentBuff` is shared, and needs to be locked while accessing it.  (which it is, within this function).  But then it is gettng a direct pointer to the data within the descriptor, unlocking, and returning that pointer for the application to access.  This is unsafe - that data can be overwritten at any time after the unlock.

Rather, this should take a pointer to the `CFE_SB_SenderId_t` buffer (not a pointer-to-pointer) and copy the data, then unlock.

Of course, this changes the API, and makes it incompatible.

One possible mitigation to make it backward compatible is to have a separate buffer in the PipeTbl struct holding another copy, then copy the data into that and return a pointer to it.  That way at least the returned pointer would remain valid until the next call to the same function on the same pipe (sort of like the way the posix `readdir` issue was mitigated).

fsw/cfe-core/src/inc/cfe_sb_events.h
```diff
@@ -933,6 +933,19 @@
...
+**
+**  \par Type: ERROR
+**
+**  \par Cause:
+**
+**  This error event message is issued when GetLastSender detects that the pntr to the buffer
+**  descriptor for the last msg received for a given pipe is NULL.  This usually occurs because
+**  CFE_SB_GetLastSenderId was called prior to recieving a msg.
+**/
+#define CFE_SB_NO_LAST_MESG_RECV_EID    64
```
jphickey, line 947: Does this really warrant an "error event"?  Seems like normal operational context - there is nothing inherently an "error" about a pipe that exists but hasn't been sent to yet.  

I can foresee this API being used to simply check if a pipe has been written to or not, where so long as the application handles the return code its not an error either way, and generating an "Error event" might be confusing.

fsw/cfe-core/src/sb/cfe_sb_api.c
skliper: See #608

fsw/cfe-core/src/inc/cfe_sb_events.h
dmknutsen: Just removed error event and pushed the updates.

fsw/cfe-core/src/sb/cfe_sb_api.c
dmknutsen: Agree with the general philosophy.  I just wanted to point out that the documentation does state that *Ptr is only valid until the next call to CFE_SB_RcvMsg for the same pipe.

astrogeco:
CCB 20200415 - APPROVED

PR #622, opened 2020-04-16
Fix #121, Add check for UsedFlag and sets APID to CFE_TBL_NOT_OWNED

### LINKED ISSUE #121: ES - Invalid Memory Handle When Restarting/Deleting an Application with Tables (GSFC DCR 14483)
When you delete an application that uses tables (e.g. HK) and then restart another task (e.g. SC) for a second time, the ES task writes to the system log that there are invalid memory handles.
 
Seems like when tables get unregistered is where the errors are happening. Message says it got a bad pointer for this table, not sure if the app in messed up. This problem is not isolated to RestartApp. It occurs in DeleteApp as well. What it looks like is that the linked list is not getting cleaned up properly when an app is deleted or restarted.
 
Further investigation in the CFS Lab narrows the problem down to the RemoveAccessLink function in cfe_tbl_internal.c. The errors are being generated on table handles from the deleted app. The buffer that is trying to be placed back into the pool is set to NULL because it has already been put back into the pool. The tables that were "cleaned up" still contain the AppID of the deleted app. When the subsequent app is restarted, its AppID becomes that of the deleted app and inherits the table handles from the previous app. For example, the HK app has 2 tables and the SC app has 73 tables. When HK is deleted, the 2 tables are removed and the entries still contain the AppID of HK. When SC is restarted, it becomes the AppID that HK was. The reason the errors occur on the 2nd restart is because on the first restart SC had a unique AppID. On the second restart, it has inherited HKs original AppID. In this case, you will see 2 sets of errors when SC is restarted. The SC application did not show any adverse functionality because of these errors. All that is happening is that the PutPoolBuf function is reporting an error when trying to return a NULL buffer to the pool.

dmknutsen:
Describe the contribution
Updates CFE_TBL_CleanUpApp such that it now checks the 'used flag' prior to calling CFE_TBL_RemoveAccessLink for a given TblHandle.  Also sets the AppId to CFE_TBL_NOT_OWNED after removing the access descriptor link from linked list.

Fix #121

Testing performed:
1. Modified cfe_es_runloop such that if AppSate == cfe_es_appstate_waiting, then it will not return false.  This was done to allow the restart command to work.
2. Created a test app that uses tables.
3. Modified table Noop command such that it prints out Table Access Descriptor data when received. 
4. Ran the following test - Note that I sent a noop after each step to review data (test results attached)
   a. Started CFE
   b. Deleted Test App
   c. Restarted Sample App
   d. Restarted Sample App

Ran test twice - Once to ensure I could recreate the error. Once to ensure changes corrected error.

5. Ran unit tests

Expected behavior changes
Will no longer receive 'CFE_TBL:RemoveAccessLink-PutPoolBuf[0] Fail' error.

System(s) tested on
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: 6.7.12.0, OSAL 5.0.9.0, PSP 1.4.7.0

Contributor Info
Dan Knutsen
NASA/Goddard

[issue_121_w_Fix.txt](https://github.com/nasa/cFE/files/4489244/issue_121_w_Fix.txt)
[issue_121_wo_Fix.txt](https://github.com/nasa/cFE/files/4489245/issue_121_wo_Fix.txt)

### DIFF
--- a/fsw/cfe-core/src/tbl/cfe_tbl_internal.c
+++ b/fsw/cfe-core/src/tbl/cfe_tbl_internal.c
@@ -1391,7 +1391,7 @@ int32 CFE_TBL_CleanUpApp(uint32 AppId)
     for (i=0; i<CFE_PLATFORM_TBL_MAX_NUM_HANDLES; i++)
     {
         /* Check to see if the Handle belongs to the Application being deleted */
-        if (CFE_TBL_TaskData.Handles[i].AppId == AppId)
+        if (CFE_TBL_TaskData.Handles[i].AppId == AppId && CFE_TBL_TaskData.Handles[i].UsedFlag == TRUE)
         {
             /* Delete the handle (and the table, if the App owned it) */
             /* Get a pointer to the relevant Access Descriptor */
@@ -1418,6 +1418,9 @@ int32 CFE_TBL_CleanUpApp(uint32 AppId)
             /* NOTE: If this removes the last access link, then   */
             /*       memory buffers are set free as well.         */
             CFE_TBL_RemoveAccessLink(i);
+	    
+            CFE_TBL_TaskData.Handles[i].AppId = (uint32)CFE_TBL_NOT_OWNED;
+
         }
     }
 


astrogeco:
CCB 20200422 - Approved

jphickey:
This pull request also breaking the current integration candidate:
```
/home/joe/code/cfecfs/github/cfe/fsw/cfe-core/src/tbl/cfe_tbl_internal.c:1394:99: error: ‘TRUE’ undeclared (first use in this function)
 1394 |         if (CFE_TBL_TaskData.Handles[i].AppId == AppId && CFE_TBL_TaskData.Handles[i].UsedFlag == TRUE)
      |                                                                                                   ^~~~
```

Should be using `true` not `TRUE` here....

Issue #643, opened 2020-04-27
Add "special" ES shell commands as real commands


skliper:
**Is your feature request related to a problem? Please describe.**
ES special commands using the SHELL FC and a special string don't follow command processing model, missing requirements, etc
CFE_ES_ListApplications
CFE_ES_ListTasks
CFE_ES_ListResources

**Describe the solution you'd like**
Implement these as real commands

**Describe alternatives you've considered**
Delete these

**Additional context**
#484 

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
Other options - Could add an event as part of QueryAllCmd (prints name to replace CFE_ES_ListApplications), QuerryAllTasksCmd (prints task name/id, app name/id to replace CFE_ES_ListTasks).  

Note CFE_ES_ListResources still seems unique/useful but it only does counts.  Might make more sense to include names or possibly other info... or maybe would make even more sense to include resource counts in the housekeeping message?  It's got counts for core apps, external apps, tasks, libs, why not resources?  Or maybe just a by request packet (add new) that has all the counts?  Send to a file doesn't really seem to have that much of a benefit.

astrogeco:
CCB 2020-04-29 : Discussed

skliper:
Note - no strong opinions expressed, seems like approach in the comments above would be beneficial to users.  Suggesting that as a path forward (events for the name/tasks as part of the already existing commands, new more informative output for resources).

igorluppi:
Is this thread still open for draco-rc4?

PR #645, opened 2020-04-27
Fix #484, Deprecate shell output command

### LINKED ISSUE #84: File operations in CFE_ES_ShellOutputCommand() need cleanup
This function can be simplified and also made more robust.

* Does unnecessary copying of the input parameters.  

* Does NOT always ensure proper null termination of the inputs.  (Copying strings could be justified if it was to ensure null termination of the inputs).  The "CFE_ES_ShellCmd" function that is calling this should ensure null termination but it does not.

* Matching of the "ES" special commands needs improvement - It will not run any OS command that happens to start with "ES_", and it will also not handle it correctly if one happens to be a substring of another but not a complete match.

* It does not check the return codes from either "OS_read" or "OS_write" calls and assumes they all work perfectly - the correct operation of this function, in fact, does depend on them all working perfectly.  In the worst-case scenario, a benign failure of an OS_write() call could have a cascade effect causing a later OS_read() call to block indefinitely.

* The return value of OS_lseek() is not properly checked.  It also relies on having an OS implementation of lseek that can accurately "measure" the size of the file.  Not all filesystems/file types are seekable in this nature.

* It will not work with any "special" file or file system that doesn't support seeking or read/write file handles (such as a pipe).

* It unnecessarily extends the file on disk to add up to 3 extra spaces, which is only to coax the message generator loop into producing an extra message.

* Overall, the loop that is generating the telemetry messages can be simplified quite a bit.

* The fixed 200ms delay between messages should at least be configurable, but there is still no way to tell if buffer overflow is occurring

* There is no sequence number, and way to tell at the receiving side of one intermediate component message was lost.  (Possibly outside the scope of this, and no way to really fix this without changing the binary format of the messages).

### LINKED ISSUE #484: Separate execute in shell functionality from core cFE service
Command to execute in shell a concern in some security scenarios.

Remove from the core, better suited to an app (optional, can be included or loaded based on mission requirements/needs)

N/A

**Additional context**
N/A

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
**Describe the contribution**
Deprecates the shell output command
Fix #484 
Fix #84 

See #643 for approaches to replacing "list" functionality (list apps, list tasks, list resources) that was implemented as "special" shell commands.

**Testing performed**
Build and ran tests locally with SIMULATION=native, ENABLE_UNIT_TESTS=true with and without OMIT_DEPRECATED, passed

**Expected behavior changes**
When deprecated attempt to send output to shell command will result in command error counter increment (unrecognized function code).

**System(s) tested on**
 - Hardware: cFS Dev Server 3
 - OS: Ubuntu 18.04
 - Versions: Based on integration branch since it had critical fixes + this branch.  Can rebase once integration branch is merged.

**Additional context**
#643 

**Third party code**
None

Jacob Hageman - NASA/GSFC


skliper:
Forced push to add "deprecated" where appropriate for doxygen.

astrogeco:
CCB 2020-04-29 : APPROVED

PR #653, opened 2020-04-29
Fix #411, rework exception handling in CFE

### LINKED ISSUE #76: CFE_ES_ProcessCoreException() is not interrupt-safe
The CFE_ES_ProcessCoreException() architecture needs to be re-examined.  

This is the equivalent of an interrupt handler and likely triggered as the result of a hardware interrupt.  However the implementation calls other CFE_ES functions, some of which take the global data mutex (CFE_ES_LockSharedData), do console output (CFE_ES_WriteToSyslog and other functions which ultimately do printf), or call OS_TaskDelay.  

All of these operations are unsafe to do in an interrupt context on most platforms.

This may not be an issue if the end-result of the exception is a processor reset, which is probably the only safe thing to do.  Although the option to restart the task does exist, the system may be too far gone after this.

### LINKED ISSUE #411: Exception and Reset Log possible race conditions
**Describe the bug**
CFE_ES_ClearERLogCmd and CFE_ES_WriteToERLog both modify shared CFE_ES_ResetDataPtr values.  CFE_ES_ProcessCoreException and CFE_ES_ResetCFE both use CFE_ES_WriteToERLog (both are API's, so could be out of ES context).

**To Reproduce**
Looks to me like if CFE_ES_ClearERLogCmd gets interrupted by the processing of an app core exception, the log could get corrupted.

**Expected behavior**
No race.

**Code snips**
See functions above.

**Versions**
 - latest cFS bundle dev branch

**Additional context**
Not observed, via code review.

**Reporter Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
Move exception handling to a PSP function.  In this approach the CFE only logs data after the event as a background job.

Replaces the CFE_ES_ProcessCoreException with a simple notification that causes the ES background job to run, which in turn polls the PSP for logged exceptions and writes entries to the ES ER log.

Both the PSP execption scan and the ER log file dump are converted to background jobs.

Fixes #411
Fixes #76 

**Testing performed**
Validate CFE exception handling behavior on POSIX and VxWorks (not implemented on RTEMS).
Add divide by zero "bug" to sample_app for testing.
Confirm exception is logged correctly
Confirm ER dump to file works

**Expected behavior changes**
No change from CMD/TLM perspective
API `CFE_ES_ProcessCoreException` is removed, replaced with async event.

**System(s) tested on**
Ubuntu 20.04 LTS 64-bit
VxWorks 6.9 on MCP750

**Additional context**
Needs corresponding changes in PSP

### DIFF
--- a/cmake/sample_defs/cpu1_platform_cfg.h
+++ b/cmake/sample_defs/cpu1_platform_cfg.h
@@ -565,7 +565,7 @@
 **       in the error log. Any context information beyond this size will
 **       be truncated.
 */
-#define CFE_PLATFORM_ES_ER_LOG_MAX_CONTEXT_SIZE     128
+#define CFE_PLATFORM_ES_ER_LOG_MAX_CONTEXT_SIZE     256
 
 
 /**
--- a/cmake/sample_defs/cpu1_platform_cfg.h
+++ b/cmake/sample_defs/cpu1_platform_cfg.h
@@ -1206,52 +1206,6 @@
 #define CFE_PLATFORM_ES_DEFAULT_STACK_SIZE 8192
 
 /**
-**  \cfeescfg Define cFE Core Exception Function
-**
-**  \par Description:
-**       This parameter defines the function-to-call when a CPU or floating point exception
-**       occurs.  The parameter is defaulted to call the ES API function #CFE_ES_ProcessCoreException
-**       which handles the logging and reset from a system or cFE core exception.
-**
-**       Note: Exception interrupts are trapped at the Platform Support Package (PSP)
-**       layer.  In order to initiate the cFE platform defined response to an exception, this
-**       platform defined callback function must be prototyped and called from the PSP
-**       exception hook API function #CFE_PSP_ExceptionHook. For example:
-**
-**       -- cfe_psp.h --
-**
-**       .... Prototype for exception ISR function implemented in CFE ....
-**
-**       typedef void (*System_ExceptionFunc_t)(uint32  HostTaskId,
-**                                              const char *ReasonString,
-**                                              const uint32 *ContextPointer,
-**                                              uint32 ContextSize);
-**
-**       -- cfe_pspexception.c --
-**
-**       .... Setup function pointer to CFE exception ISR callback ....
-**
-**       static const System_ExceptionFunc_t CFE_ExceptionCallback = CFE_PLATFORM_ES_EXCEPTION_FUNCTION;
-**
-**       void CFE_PSP_ExceptionHook (int task_id, int vector, uint8 *pEsf )
-**       {
-**           .... platform-specific logic ....
-**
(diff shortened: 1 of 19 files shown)


jphickey:
Should be ready for merge now.

jphickey:
This was reviewed in a special topic on 2020-04-30, but may want to mention this in a CCB again.

astrogeco:
CCB-20200506 - APPROVED. Also covered in special topic on 4/30

dmills3:
Sorry for the delay in posting my comments. 

1. I have a question about how CFE handles exception. Should the system reset if a task not managed by CFE receives an exception?
The function CFE_ES_RunExceptionScan will call CFS_PSP_Restart in  cfe_es_erlog.c and restart the system when running on VxWorks.
2. The TblHandle is used before validating on line 681 in the file cfe_tbl_api.c. It is validated on line 696 in the function CFE_TBL_Load.

jphickey:
> 1. I have a question about how CFE handles exception. Should the system reset if a task not managed by CFE receives an exception?
>    The function CFE_ES_RunExceptionScan will call CFS_PSP_Restart in  cfe_es_erlog.c and restart the system when running on VxWorks.

In the current framework, yes, I think this is reasonable.  The "mcp750-vxworks" PSP is primarily a test/reference platform and it is likely/expected that CFE is the only service running within the system.   It is also consistent with the way MCP750 has worked before.

As the exception hook is implemented in the board-specific code this could do something else on other platforms, such as ignoring the exception if it was generated from a task outside CFE.  But even still, assuming whatever actually did generate the exception was a necessary service, it might be warranted to restart the processor if it fails anyway.

> 2. The TblHandle is used before validating on line 681 in the file cfe_tbl_api.c. It is validated on line 696 in the function CFE_TBL_Load.

This PR does not change anything in `cfe_tbl_api.c` ... what you might have seen earlier was likely due to a baseline mismatch.  This is corrected since the rebase - so if you look at the diff now it is easier to read.

skliper:
> 2. The TblHandle is used before validating on line 681 in the file cfe_tbl_api.c. It is validated on line 696 in the function CFE_TBL_Load.

I wrote this up as a separate issue, see #687

dmills3:
The exception hook does not manage the CFE task. Maybe CFE could call a board specific function to do something else after CFE has determined it not a CFE task.

dmills3:
Another thought the CFE exception handler could return and let the board specific code decide to restart or not.

jphickey:
If the exception is not relevant to CFE then the hook can just drop it and not log it, or it can do something else entirely.  Any exceptions which ARE written to the internal buffer are assumed to be related to CFE and will trigger a restart if no specific action is configured.

This is entirely up to the PSP as to how to disposition.  The current implementation just assumes that everything is related to CFE because that is expected to be the only thing running.

PR #660, opened 2020-04-30
Fix #583, ES start table

### LINKED ISSUE #583: ES should use a table for the "startup script"

CDKnightNASA:
**Describe the contribution**
This is a re-do of the CFE ES "startup" table code. This includes unit tests (coverage) and requested "volatile" boot option.

**Testing performed**
both unit tests and running the core

**Expected behavior changes**
This changes ES to use a table to load libraries and load and start applications. This is a significant change to the internals, as libraries are loaded and applications are started in the ES task (spawned by the ES main task) and, of course, a table is used instead of a .scr file.

 - API Change: internal API changes
 - Behavior Change: loading from .tbl instead of .scr

**System(s) tested on**
Debian 9 VM

Christopher.D.Knight@nasa.gov


CDKnightNASA:
I realize I need to update documentation. Will work on that next week.

CDKnightNASA:
doc changes added (note that the doxygen references a deployment guide...guessing that doesn't exist? ;) )

CDKnightNASA:
Also should I remove the ability to exclude TBL from a build as part of this pull request?

skliper:
> Also should I remove the ability to exclude TBL from a build as part of this pull request?

I suggest a separate issue/pull request

skliper:
> doc changes added (note that the doxygen references a deployment guide...guessing that doesn't exist? ;) )

The deployment guide costs extra.  Yeah, should remove reference.

docs/cFE Application Developers Guide.md
```diff
@@ -583,7 +583,7 @@ Reference cFE API documentation for more detail on reset types.
 
 The cFE contains support for shared libraries. For the current version
 of the cFE, the shared libraries must be loaded on cFE startup (see the
-cFE Deployment Guide on how to modify the cfe_es_startup.scr in order
+cFE Deployment Guide on how to configure the cfe_es_start.tbl in order
```
astrogeco, line 586: Open ticket for this

astrogeco: Should be in users guide

astrogeco:
CCB-20200506 Concept is good. Review again once 6.8 is finished

skliper:
Consider as alternate implementation inclusion in Draco (split into module and allow selection).  Trade additional test/maintenance complexity for supporting both options.  Likely an easier full switch/replacement if there was support for a simple table editor (EDS based?), in general all tables would benefit from this...

PR #662, opened 2020-05-01
Fix #488, Remove compiler added padding in tlm packets

### LINKED ISSUE #488: cFE Tlm Packet 64-bit alignment issue
**Describe the bug**
Compiler add padding is applied to cFE Telemetry packets if the data region does not start on a 64 bit boundary and the packet contains a 64 bit variable.

**To Reproduce**
1. Nominal Build Process.
2. Enable Tlm.
3. Bit bust a telemetry packet containing a 64 bit variable.

**Expected behavior**
No compiler added padding should be applied.

**System observed on:**
Ubuntu 64-bit - 19.10

**Reporter Info**
Dan Knutsen
NASA/GSFC

dmknutsen:
**Describe the contribution**
Fix #488 
Removes compiler added padding by padding CCSDS_TlmSecHdr_t such that the payload will always start on a 64-bit boundary.  Also adjusts packet alignment such that packets with 64 bit variables are aligned correctly.

Related to nasa/cFS-GroundSystem#81
Intended to be delivered with nasa/cFS-GroundSystem#82

**Testing performed**
1. Modified GroundSystem.py to display raw packets as they are being decommed
2. Bit busted packets to ensure no compiler added padding is being applied.
3. Ran unit tests.

**Expected behavior changes**
Compiler added padding will no longer be applied to packets.

**System(s) tested on**
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cFE 6.7.14.0, OSAL 5.0.13.0, PSP 1.4.9.0

Dan Knutsen
NASA/Goddard

### DIFF
--- a/fsw/cfe-core/src/inc/ccsds.h
+++ b/fsw/cfe-core/src/inc/ccsds.h
@@ -122,6 +122,12 @@ typedef struct {
 
    uint8  Time[CCSDS_TIME_SIZE];
 
+#if ( CCSDS_TIME_SIZE == 6 )
+   uint8  tempPad1[4];
+#elif ( CCSDS_TIME_SIZE == 8 )
+   uint8  tempPad1[2];
+#endif
+
 } CCSDS_TlmSecHdr_t;
 
 /*----- CCSDS Endian Flag in the APID Qualifier Field. -----*/
--- a/fsw/cfe-core/src/inc/cfe_es_msg.h
+++ b/fsw/cfe-core/src/inc/cfe_es_msg.h
@@ -1412,8 +1412,8 @@ typedef struct
 **/
 typedef struct
 {
-  char                  Application[CFE_MISSION_MAX_API_LEN];   /**< \brief - RESERVED - should be all zeroes */
-  CFE_ES_MemHandle_t    PoolHandle;                     /**< \brief Handle of Pool whose statistics are to be telemetered */
+  char                  Application[CFE_MISSION_MAX_API_LEN];   /**< \brief Application Name */
+  CFE_ES_MemHandle_t    PoolHandle;                             /**< \brief Handle of Pool whose statistics are to be telemetered */
 
 } CFE_ES_SendMemPoolStatsCmd_Payload_t;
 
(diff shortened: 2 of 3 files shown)


astrogeco:
CCB-20200506 - Will conflict with some 6.8 changes so we'll need to handle that. Related discussion on #678 .

fsw/cfe-core/src/inc/ccsds.h
```diff
@@ -122,6 +122,12 @@ typedef struct {
 
    uint8  Time[CCSDS_TIME_SIZE];
 
+#if ( CCSDS_TIME_SIZE == 6 )
+   uint8  tempPad1[4];
+#elif ( CCSDS_TIME_SIZE == 8 )
+   uint8  tempPad1[2];
+#endif
```
jphickey, line 129: Padding shouldn't be added this way -- it isn't a fixed quantity.  Furthermore the "ccsds.h" definition should be the source of truth -- that is, it should be the actual definition of the structure without getting into any architecture-specific padding requirements.

The other reason not to do it this way is that the opposite problem exists when extended headers is enabled.  This adds 4 bytes to the primary, so now the size of TLM hdr becomes 16 and the CMD header becomes 12, so now CMDs will need padding and TLM packets will not.

The paradigm should be that `ccsds.h` definitions represent the basic fundamental primitives, and `cfe_sb.h` structures represent how these are actually realized for use within the software bus and applications.

So - in the related PR #678 the `CFE_SB_TlmHdr_t`  becomes a padded and aligned version of `CCSDS_TlmSecHdr_t`

One can reliably determine the actual amount of padding on the current system by checking `sizeof(CFE_SB_TlmHdr_t) - sizeof(CCSDS_TlmSecHdr_t)` or equivalent...

This will work on 32 bit or 64 bit, with or without extended headers, and should continue to work even if the headers are changed in the future.

skliper:
Marking as duplicate and closing for now, can reinvestigate as part of future packet refactor for the padding within packets.

fsw/cfe-core/src/inc/ccsds.h
CDKnightNASA: or what about using a (surrounding) union, like:

 `typedef union { uint8 Time[CCSDS_TIME_SIZE], TotalSz[10] } CCSDS_TlmSecHdr_t;`

PR #678, opened 2020-05-06
Fix #666, alignment of CMD/TLM message definitions

jphickey:
**Describe the contribution**
Define the SB message headers such that they are aligned appropriately for casting to/from a `CFE_SB_Msg_t` type.  Note this should not affect the size of these structures on a 32-bit machine with the default configuration, as they are already multiples of `sizeof(uint32)`....

Then change only the CMD/TLM definitions which were generating warnings on the 32-bit build to use this correct definition, rather than a `uint8[]` array to reserve space for the header.  Again in these cases this should not change the size or padding because it already was multiples of `sizeof(uint32)`.  This just makes it so the compiler will appropriately align the overall buffer.

Fix #666

**Testing performed**
Build for 64 + 32 bit platforms with strict alignment settings and confirm that the warnings are gone.  Confirm all tests pass.

**Expected behavior changes**
No impact to behavior.

**System(s) tested on**
Ubuntu 20.04
MIPS yocto/poky embedded Linux (32-bit big endian processor with strict alignment requirements)
RTEMS on QEMU (i686-rtems4.11)

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #690, opened 2020-05-08
Fix #689, Align all software bus message definitions

### LINKED ISSUE #689: Align all software bus message definitions
As an extension to the changes in issue #666/pull #678, the same paradigm should be applied to all other message definitions.  This is currently only working "by chance" in that the payloads already contain a uint32 (or larger) value so it is already aligned.

Change the header structure which is currently defined as a `uint8 []` array into either CFE_SB_CmdHdr_t or CFE_SB_TlmHdr_t so it will be correctly aligned.

**Additional context**
It is safe to assume all remaining message types were already 32-bit aligned because there were no remaining compiler warnings about this, which means this change by itself will not have any effect on the existing message sizes or alignment.

However, when moving to a 64-bit build it may become necessary to enforce 64-bit alignment rather than 32-bit alignment for `CFE_SB_Msg_t` and in that case, the sizes and padding may change from what it currently is.

**Requester Info**

jphickey:
**Describe the contribution**
Replace uint8[] arrays which reserve space for the header with an instance of the header struct as defined by SB.
    
Note this structure was the basis for the array size, so it is the same size, but by actually using the structure  the resulting message will have the correct alignment.

Fixes #689

**Testing performed**
Execute all unit tests and confirm passing
Sanity check on CFE, builds and runs and processes commands
Also Build and run on 32-bit platform with strict alignment and confirm no warnings/issues

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04 LTS (native, little endian 64 bit)
MIPS 32-bit via QEMU (big endian, strict alignment requirements)

**Additional context**
Just applies the same fix from #678 to all messages.

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_es_msg.h
+++ b/fsw/cfe-core/src/inc/cfe_es_msg.h
@@ -1120,7 +1120,7 @@
 */
 typedef struct
 {
-  uint8                 CmdHeader[CFE_SB_CMD_HDR_SIZE];     /**< \brief cFE Software Bus Command Message Header */
+  CFE_SB_CmdHdr_t CmdHeader;     /**< \brief cFE Software Bus Command Message Header */
 
 } CFE_ES_NoArgsCmd_t;
 
@@ -1151,7 +1151,7 @@ typedef struct
 
 typedef struct
 {
-    uint8                 CmdHeader[CFE_SB_CMD_HDR_SIZE];    /**< \brief cFE Software Bus Command Message Header */
+    CFE_SB_CmdHdr_t CmdHeader;    /**< \brief cFE Software Bus Command Message Header */
     CFE_ES_RestartCmd_Payload_t Payload;
 } CFE_ES_Restart_t;
 
@@ -1176,7 +1176,7 @@ typedef struct
  */
 typedef struct
 {
-    uint8                       CmdHeader[CFE_SB_CMD_HDR_SIZE];    /**< \brief cFE Software Bus Command Message Header */
+    CFE_SB_CmdHdr_t CmdHeader;    /**< \brief cFE Software Bus Command Message Header */
     CFE_ES_ShellCmd_Payload_t   Payload;
 } CFE_ES_Shell_t;
 #endif /* CFE_OMIT_DEPRECATED_6_7 */
@@ -1198,7 +1198,7 @@ typedef struct
 
 typedef struct
 {
-    uint8                        CmdHeader[CFE_SB_CMD_HDR_SIZE];    /**< \brief cFE Software Bus Command Message Header */
+    CFE_SB_CmdHdr_t CmdHeader;    /**< \brief cFE Software Bus Command Message Header */
     CFE_ES_FileNameCmd_Payload_t Payload;
 } CFE_ES_FileNameCmd_t;
 
@@ -1226,7 +1226,7 @@ typedef struct
 
 typedef struct
 {
-    uint8                               CmdHeader[CFE_SB_CMD_HDR_SIZE];    /**< \brief cFE Software Bus Command Message Header */
+    CFE_SB_CmdHdr_t CmdHeader;    /**< \brief cFE Software Bus Command Message Header */
     CFE_ES_OverWriteSysLogCmd_Payload_t Payload;
 } CFE_ES_OverWriteSyslog_t;
 
@@ -1255,7 +1255,7 @@ typedef struct
 
 typedef struct
 {
-    uint8                           CmdHeader[CFE_SB_CMD_HDR_SIZE];    /**< \brief cFE Software Bus Command Message Header */
+    CFE_SB_CmdHdr_t CmdHeader;    /**< \brief cFE Software Bus Command Message Header */
     CFE_ES_StartAppCmd_Payload_t    Payload;
 } CFE_ES_StartApp_t;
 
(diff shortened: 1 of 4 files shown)


jphickey:
Initially created as draft to due to baseline issue.  Only real commit is d1da5f2.  Will rebase after next merge to master.

skliper:
CCB 20200523 - Approved conceptually on telemetry side, needs rebase

Recommended splitting up since cmd impacts should go into 7.0

astrogeco:
@jphickey can you split up this PR and separate the CMD and TLM changes?

skliper:
Rebase and go forward with this?

jphickey:
The issue with this one is that it does potentially change padding, so we need a consensus on how to deal with that.    Ideas have been tossed around but I don't know of an accepted path forward.  The real solution is a data dictionary/message definition DB of some type, but my understanding is that 7.0 still won't have that, so we should have some sort of steering committee discussion on how to handle changes int he binary formats for this release.

skliper:
So I've been tripping up on this lately.  Can we be explicit about the impacts?  Since the packets are already 32 bit aligned when set via CFE_SB_SendMsg (after the union with CFE_SB_Msg_t) and the header sizes are already multiples of 32 bits (8 or 12 bytes without extended headers, 12 or 16 bytes with extended headers), what would change relative to the packet structure after the union?  Isn't anything that was going to change already changed with the union?

Any payloads that require more than 32-bit alignment still break CFE_SB_GetUserData either way, since it won't point to the start of the payload due to padding to 64 bits after the header.

skliper:
Or was the issue with the variable legacy TIME formats, if set to a size of 8 this would have made the CCSDS_TelemetryPacket_t size 14?  But the union with CFE_SB_Msg_t to create CFE_SB_TlmHdr_t would have rounded that to 16 typically anyways...

jphickey:
Commands are (currently) not defined as aligned structures.  In fact they are explicitly UN-aligned.  

For instance, consider `CFE_ES_Restart_t`:
https://github.com/nasa/cFE/blob/a148b97511de3f45f6677c1a3a0b17197f867e28/fsw/cfe-core/src/inc/cfe_es_msg.h#L1152-L1156

The size of this command is "traditionally" 10 bytes, but if defined properly with 32-bit alignment the sizeof() this struct becomes 12 bytes, so it will fail the size check unless the ground system either includes this extra padding _OR_ the system accounts for/adjusts for the padding difference in either CI or within the software bus code itself.

skliper:
But doesn't the union with CFE_SB_Msg_t already make the resulting structure size 12 for your example above (typical compiler behavior is to pad)?  My point is we already did the union, so in effect we've already changed the sizes of the commands, and this change will not actually change the sizes again.  Confirmed on linux...

jphickey:
> the union with CFE_SB_Msg_t

Which union are you referring to, specifically?  `sizeof(CFE_ES_Restart_t)` is _not_ dependent on `sizeof(CFE_SB_Msg_t)` -- the header in this struct is defined as `uint8 CmdHeader[CFE_SB_CMD_HDR_SIZE];` so it has alignment of 1.  I can confirm that on my system right now `sizeof(CFE_ES_Restart_t)` is indeed 10, not 12.

Yes - `CFE_SB_Msg_t` has 32-bit alignment so one can cast from `CFE_SB_Msg_t*` to `CFE_ES_Restart_t*` without a warning but not vice-versa.

skliper:
OK, I'm on the same page again... the issue is with commands for which we didn't already make a typedef union w/ CFE_SB_Msg_t.  I was looking at telemetry, which we did already align.

jphickey:
Right - for packets which are locally generated (TLM) we make a union - so the buffer in memory is aligned even if the actual TLM structure definition isn't aligned - so it in turn can be safely passed to `CFE_SB_SendMsg()` which accepts a `CFE_SB_Msg_t*`.  But as a receiver (CMD) we don't make our own buffer, just cast the buffer we get to the correct type -- which should only be of _less stringent_ alignment than `CFE_SB_Msg_t` (double/long long being the exception/issue).

I believe we've done all we can in 6.8 to make this safe as it can be without a more comprehensive approach to dealing with CPU-specific padding and alignment issues.

skliper:
Note we expect to receive cmd/tlm structure definitions (potentially platform/compiler specific) from the stakeholder that are generated from a common data model and produced with alignment/padding per their requirements.  The open source community can evaluate use/incorporation of those structures, but the database for CTF automated testing will depend on those definitions.

skliper:
Latest discussion - stakeholder is going to use the method implemented here (updating both commands and telemetry definitions to use the real headers), for open source to match we should plan to get this merged and update the related ground databases.  For the framework it impacts cFS-GroundSystem and the EDS definitions.  Recommending community notification.

skliper:
We should get this updated and move forward.  It would simplify some other changes I'm working, and we aren't gaining anything by putting it off.

jphickey:
@skliper can you elaborate on what the issue is that you are facing?

As far as I can tell the earlier concern still exists (why I was holding off on this) in that it WILL change the overall size of any command whose size is not currently a multiple of 4 bytes.  For instance the "Restart" command but others too.  This change in size will effectively break anything sending these commands.

skliper:
I'm doing the tlm packets now as part of #777.  They don't break and it makes sense with the MSG APIs to do it right.  For command packets, we've got to do it at some point so why not now?  Will it break, yes.  Better to break it now than late in the cycle.

astrogeco:
[CCB 2020-11-12](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-November-12-2020)
- We already have telemetry messages
- Depends on how we decide to handle the cFE_Msg header
- Pending #1009

skliper:
#689 is OBE, no longer enforcing worst case alignment in all messages per #1009.  This is mostly a duplicate of #1015.

PR #708, opened 2020-05-16
Fix #707, Resolve highest MsgID of 0xFFFF bug

### LINKED ISSUE #414: Suggested upper limit value creates infinite loop
When the user sets `CFE_PLATFORM_SB_HIGHEST_VALID_MSGID`, the documentation specifies a range of 1 to 0xFFFF.  This value is later used in a `#define` in cfe_sb_priv.h as follows:

     #define CFE_SB_MAX_NUMBER_OF_MSG_KEYS   (1+CFE_PLATFORM_SB_HIGHEST_VALID_MSGID)

This value is then used as the upper limit of a for loop [here](https://github.com/nasa/cFE/blob/5d7d766941369b6e92a0328217e43d5fd02a8cf1/fsw/cfe-core/src/sb/cfe_sb_init.c#L224).
   
The problem is that if the user sets the upper limit of 0xFFFF as specified, the `CFE_SB_MAX_NUMBER_OF_MSG_KEYS` becomes `65536`.  This value is used as the upper range of the for loop linked above.  Since the iterator used in that loop is a unsigned 16 bit integer, it can never reach `65536` and creates an infinite loop.

Suggested fix would be to:   
1.  Stop using `#define`s which add to other `#define`s
2.  Never use `#define`s as iterator range variables
3.  Check types of all iterators
4.  Change the documentation of `CFE_PLATFORM_SB_HIGHEST_VALID_MSGID` to say max value of `0xFFFE`

### LINKED ISSUE #707: Setting highest valid message ID's to 0xFFFF hangs cfe
**Describe the bug**
Given max number of message keys is 1+ highest valid msgid:
https://github.com/nasa/cFE/blob/95f34d25cb2843b8ba9db6338bb0b53cb3e38f92/fsw/cfe-core/src/sb/cfe_sb_priv.h#L107

and message key is uint16:
https://github.com/nasa/cFE/blob/95f34d25cb2843b8ba9db6338bb0b53cb3e38f92/fsw/cfe-core/src/sb/cfe_sb_priv.h#L122

Setting CFE_PLATFORM_SB_HIGHEST_VALID_MSGID to 0xFFFF results in forever loop in:
https://github.com/nasa/cFE/blob/95f34d25cb2843b8ba9db6338bb0b53cb3e38f92/fsw/cfe-core/src/sb/cfe_sb_init.c#L219-L224

**To Reproduce**
Set CFE_PLATFORM_SB_HIGHEST_VALID_MSGID to 0xFFFF and build/run.

**Expected behavior**
Full 16 bits of message id should be usable.

**Code snips**
See above

**System observed on:**
 - Hardware: cFS Dev 3
 - OS: Ubuntu 18.04
 - Versions: master bundle

**Additional context**
None

**Reporter Info**
Jacob Hageman - NASA/GSFC (per JSC identification of the issue)

skliper:
**Describe the contribution**
Changes Message Key from uint16 to uint32 to avoid rollover and system hang
Fix #707 
Fix #414

**Testing performed**
1. Set CFE_PLATFORM_SB_HIGHEST_VALID_MSGID to 0xFFFF
1. Built (SIMULATION=native) and ran, confirmed startup
1. CI - https://travis-ci.com/github/skliper/cFS/builds/166553342

**Expected behavior changes**
Full message ID range available

**System(s) tested on**
 - Hardware: cFS Dev 3
 - OS: Ubuntu 18.04
 - Versions: Bundle w/ this change

**Additional context**
Identified/resolved by JSC

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/cmake/sample_defs/cpu1_platform_cfg.h
+++ b/cmake/sample_defs/cpu1_platform_cfg.h
@@ -178,7 +178,8 @@
 **       The recommended case to to have this value the same across all mission platforms
 **
 **  \par Limits
-**       This parameter has a lower limit of 1 and an upper limit of 0xFFFF. 
+**       This parameter has a lower limit of 1 and an upper limit of 0xFFFF. Note there
+**       is one MsgId value reserved as invalid, so the table is this size + 1. 
 */
 #define CFE_PLATFORM_SB_HIGHEST_VALID_MSGID      0x1FFF
 
--- a/fsw/cfe-core/src/sb/cfe_sb_priv.h
+++ b/fsw/cfe-core/src/sb/cfe_sb_priv.h
@@ -119,7 +119,7 @@
 **           be used directly, except by internal table lookups.
 **
 */
-typedef uint16  CFE_SB_MsgKey_Atom_t;
+typedef uint32  CFE_SB_MsgKey_Atom_t;
 
 /******************************************************************************
 **  Typedef:  CFE_SB_MsgKey_t


skliper:
Ping @tngo67

skliper:
@astrogeco requesting fast-track.

skliper:
> What about just making it "int", as it's only used as an index into an array and would be more compiler optimization friendly?

Minimum size for int is 16 bits.  Could go unsigned long?  Happy to go either way, I just went fixed size since it was fixed size.  Looking at the code in general fixed sizes are used even for iterators... I checked our standard and it doesn't care either way.

jphickey:
> > What about just making it "int", as it's only used as an index into an array and would be more compiler optimization friendly?
> 
> Minimum size for int is 16 bits. Could go unsigned long? Happy to go either way, I just went fixed size since it was fixed size. Looking at the code in general fixed sizes are used even for iterators... I checked our standard and it doesn't care either way.

I agree, in this case the fundamental issue is that MsgKey didn't have enough bits.  The algorithm requires that the key has enough bits to hold at least `(1 + CFE_PLATFORM_SB_HIGHEST_VALID_MSGID)`.  This means we need to ensure it actually has 32 bits in it, if we want to support a highest msgid of 0xFFFF or greater.

I think in the original implementation, there was an assumption that if the lookup table grew beyond 2^16 entries, it would use some other more memory efficient algorithm like a hash table, rather than a direct lookup table.  Even at 2^16 this is probably wasting a good chunk of memory.

skliper:
Added note about the table being  + 1.

skliper:
@astrogeco please fast track this one

astrogeco:
> @astrogeco please fast track this one

Will merge right after CCB

PR #720, opened 2020-05-27
Fix #718, add user-specified extra modules to build

### LINKED ISSUE #718: Add user-specified extra modules to build system
As a prerequisite to #554, the user needs to have the ability to specify their own set of extra modules to provide for the features they are customizing.  

A new setting in `targets.cmake` that allows users to add their own customization layers.

This is similar to but slightly different than loadable apps in that:

- The same set/config should be applied to all CPUs  (i.e. if using for a message abstraction layer, all CPUs should share the same one).  So it belongs as a global setting, rather than a setting associated with a single target.
- These would be linked with cfe core, rather than as dynamically-loaded libraries because something like a message abstraction layer is needed by SB itself, so it can't be loaded later.

**Additional context**
Initial function would be very simple - just include the libraries in the build, nothing more.

**Requester Info**

jphickey:
**Describe the contribution**
Add a new setting that can be set in targets.cmake, to enable users to add extra custom functions/modules to CFE core executable.

This can be used, among other things, for future support of a modular/customizable message header structure.

Fixes #718

**Testing performed**
Build and execute CFE, confirm normal operation/sanity check.
Build and execute all unit tests, confirm passing.

**Expected behavior changes**
Any user-supplied library can be added to the build.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This general concept can also be used to provide replacement implementations of core modules (e.g. TBL, FS) with some additional tweaking to exclude those modules from the subsequent build of CFE core.

### DIFF
--- a/cmake/mission_build.cmake
+++ b/cmake/mission_build.cmake
@@ -127,7 +127,7 @@ function(prepare)
   # so this is run in a loop until the list of unfound apps is empty
   string(REPLACE ":" ";" CFS_APP_PATH "$ENV{CFS_APP_PATH}")
   list(APPEND CFS_APP_PATH "apps" "apps/CFS" "libs" "psp/fsw/modules")
-  set(MISSION_DEPS "cfe-core" "osal")
+  set(MISSION_DEPS "cfe-core" "osal" ${MISSION_CORE_MODULES})
   set(APP_MISSING_COUNT 0)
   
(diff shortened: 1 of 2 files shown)


jphickey:
Withdrawing this one, see PR #740 instead which also includes this and some other things.

PR #726, opened 2020-05-28
Fix #711, Separate secondary header access functions

### LINKED ISSUE #529: CFE_SB_GetTotalMsgLength return type doesn't support max packet size
CFE_SB_GetTotalMsgLength returns an uint16, max packet CCSDS packet size can be slightly bigger (65542 vs 65536).  Note the c documentation claims the max return doesn't include the primary header (#528), but the API (h file documentation) claims it's the full packet.  The code actually returns the full packet size (the length field in the CCSDS packet + 7).

Support full CCSDS packet size.

None

**Additional context**
Initiated by JP/PACE

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #597: Remove local-endian SID macros, and unnecessary abstraction of mask/shift
The following macros aren't clearly documented as to use.  They only work on a local endian StreamID (like what comes from `CCSDS_RD_SID`).

https://github.com/nasa/cFE/blob/62252d11409f337d3dea2732739e068987363985/fsw/cfe-core/src/inc/ccsds.h#L437-L454

The CCSDS_RD_BITS/WR_BITS isn't CCSDS related, and is just a mask/shift.  More straight forward to just use mask/shift.  See conversation on https://github.com/skliper/cFE/commit/a8c24ea4370d99b097f91a2c3a8dd76d9202dc74#commitcomment-38382611

Remove these since they just add to confusion.  Just use the CCSDS_RD_SID/APID/SHDR/TYPE/VERS macros directly on the header.

Could deprecate, but no known uses.

**Additional context**
Conversation stemmed from #568

**Requester Info**
Jacob Hageman - NASA/GSFC

*EDIT* fixed code blob
*EDIT* fixed my initial issue title and updated description per @jphickey clarification of intended use

### LINKED ISSUE #711: Separate secondary header access API's from SB
Need to support #554 
Software bus logic doesn't care about message format, just use message access APIs like all the other services.

MSG module and header file for all the message getter/setters.  Deprecation of the SB API's.  Code separation to support source selection/mission configuration.

None

**Additional context**
#554

**Requester Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #733: CFE_SB_ValidateChecksum implementation does not match description
**Describe the bug**
CFE_SB_ValidateChecksum claims it returns true if not supported:

https://github.com/nasa/cFE/blob/d217ca39f1de74695f0d4e0eb427f90fbe8b1156/fsw/cfe-core/src/inc/cfe_sb.h#L1314-L1316

but also claims it returns false:

https://github.com/nasa/cFE/blob/d217ca39f1de74695f0d4e0eb427f90fbe8b1156/fsw/cfe-core/src/inc/cfe_sb.h#L1321-L1323

and then implements a false return:

https://github.com/nasa/cFE/blob/d217ca39f1de74695f0d4e0eb427f90fbe8b1156/fsw/cfe-core/src/sb/cfe_sb_util.c#L398-L401

**To Reproduce**
Read the code.

**Expected behavior**
Consistency

**Code snips**
See above

**System observed on:**
N/A

**Additional context**
None

**Reporter Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #736: Version 2 MsgId construction doesn't match description, overloads bits
**Describe the bug**
Version 2 code takes the full APID (0x7FF mask), or's in a bit for cmd/tlm (0x80 mask) then or's in the Subsystem ID shifted by 8

That means if a user defines an APID of 0x80 for a telemetry message (which is valid per CCSDS), the system will report it as type cmd if it gets the type from the msgid.  It's also a collision between 0x7 bits from the Subsystem ID and the 0x700 bits of APID.

Basically logic doesn't mirror:
CFE_SB_SetMsgId of 0x7FF -> APID = 0x7F, Type = Cmd, SubsystemID = 7
CFE_SB_GetMsgId from APID=0x7FF, Type = Tlm, SubsytemID = 0 -> MsgId = 0x7FF

**To Reproduce**
N/A - code inspection

**Expected behavior**
Get/Set should mirror (SetMsgId should not overload bits)

**Code snips**
https://github.com/nasa/cFE/blob/d217ca39f1de74695f0d4e0eb427f90fbe8b1156/fsw/cfe-core/src/sb/cfe_sb_msg_id_util.c#L118-L152
https://github.com/nasa/cFE/blob/d217ca39f1de74695f0d4e0eb427f90fbe8b1156/fsw/cfe-core/src/sb/cfe_sb_msg_id_util.c#L155-L187
https://github.com/nasa/cFE/blob/d217ca39f1de74695f0d4e0eb427f90fbe8b1156/fsw/cfe-core/src/sb/cfe_sb_msg_id_util.h#L64-L75

**System observed on:**
N/A

**Additional context**
Uncovered as part of #711 work

**Reporter Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #781: CCSDSv2 MsgId construction not 16 bits as described in cfe_sb_msg_id_util.h
**Describe the bug**
CCSDSv2 MsgId construction not 16 bits as described in cfe_sb_msg_id_util.h. It looks like the 8th bit of the APID qualifier subsystem ID is included, making the MsgID 17 bits. 

**To Reproduce**
N/A, code inspection

**Expected behavior**
MsgId should mask off bit 8 in subsystem ID

**Code snips**
https://github.com/nasa/cFE/blob/a47abba14b7e8ffbc8c8e78614babdb151412af2/fsw/cfe-core/src/sb/cfe_sb_msg_id_util.h#L40-L58

https://github.com/nasa/cFE/blob/a47abba14b7e8ffbc8c8e78614babdb151412af2/fsw/cfe-core/src/inc/ccsds.h#L346

https://github.com/nasa/cFE/blob/a47abba14b7e8ffbc8c8e78614babdb151412af2/fsw/cfe-core/src/sb/cfe_sb_msg_id_util.c#L137-L138

**System observed on:**
N/A

**Additional context**
N/A

**Reporter Info**
John N Pham, Northrop Grumman

skliper:
**Describe the contribution**
Fix #711 - Separates header manipulation and definitions for easier mission customization 
Fix #733 - Fixes validate checksum description to match implementation
Fix #736 - Fixes get msgid logic to not override bits
Fix #597 - Removes local endian SID macros
Fix #529 - By adding an API that does support maximum msg size reporting, note also fixes limit in CFE_SB_InitMsg by the same means.
Fix #781 - Fixes set/get message id logic to not override bits

 - Enables source selection and out-of-tree mission defined overrides in the msg directory.

NOTE - There are follow-on tickets to finalize and clean up after this change.

**Testing performed**
Ran unit tests for version 1 and version 2 header formats, all pass except sample_app due to a different bug (nasa/sample_app#87)

**Expected behavior changes**
This just enables override.  No actual changes in behavior other than the bug fixes.  Some name changes in structure fields, but apps shouldn't be directly accessing those fields to begin with.

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: master bundle + this commit (and the other branch configured to override)

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC


skliper:
Note - the override just allows use of the full 8 bits for function code, vs the cfe default of 7.  Really just using as an example of how it could be easily customized.

skliper:
There's still quite a bit of cleanup/refactor that could be done relative to ccsds.h and ccsds.c, and I'm sure many improvements in how I hacked in the CMake changes... but curious to hear if this is a step in a direction that we could build on.  It's an attempt at a trivial implementation to enable both the immediate needs and future enhancements.

fsw/cfe-core/src/inc/cfe_msg.h
```diff
@@ -0,0 +1,220 @@
...
+**          - If the underlying implementation of software bus messages does not
+**            include a command code field, then this routine will return a zero.
+**
+** \param[in]  MsgPtr      A pointer to the buffer that contains the software bus message.
+**                         This must point to the first byte of the message header.
+**
+** \return The command code included in the software bus message header (if present).
+**         Otherwise, returns a command code value of zero.
+**/
+uint16 CFE_MSG_GetCmdCode(CFE_SB_MsgPtr_t MsgPtr);
```
jphickey, line 157: Another gripe about the API - The "SetCmdCode" above returns an `int32` status and thereby offers a distinct return code `CFE_SB_WRONG_MSG_TYPE` if you pass it a message that isn't a command.  But "GetCmdCode" is not a mirror image of this and directly return a value.  Why?

It is clear that the existing CFE_SB API probably "grew" organically by different devs at different times which can explain the differences/mismatches.  But if we are making a new API we should address these.

Again - this would be fine if this `cfe_msg.h` were header were inside the library or otherwise private/restricted to SB, but this PR is making it publicly available in `fsw/cfe-core/src/inc` ... 

Public APIs will have to be lived with and maintained for years to come, so its best to really think about them and design them.  A new message abstraction offers the chance to fix mistakes of the past, we shouldn't just propagate the old (broken/mismatched) API paradigms to a new name without addressing the shortcomings.

Recommendation for the near term is to keep this include file quasi-private by making the user explicitly add an `include_directories` to their CMake recipe to get it and/or use the `target_include_directories()` cmake function with the PUBLIC or INTERFACE type to accomplish this.  Therefore users that want to call these functions in the near term still can, but we can do a proper design cycle/review before adding new public APIs that we will be committed to.

fsw/cfe-core/src/msg/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+#
+##################################################################
+
+# Add the basic set of files which are always built
+set(CFE_ALL_MODULE_SRCS ${CFE_ALL_MODULE_SRCS}
+    ${CMAKE_CURRENT_SOURCE_DIR}/cfe_msg_checksum.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/cfe_msg_fc.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/cfe_msg_time.c
+    PARENT_SCOPE
+)
```
jphickey, line 22: This is very unusual to say the least -- Why wouldn't this just do an `add_library` here?  I've never seen a CMake file add sources to its parent like this.  At the very minimum, you could make an OBJECT lib and then include the objects in the lib in the parent.  This is how OSAL does it.

skliper:
> It's OK as a first step - but there are some things I'd definitely want to address before merging.

Excellent improvements, thanks!

fsw/cfe-core/src/inc/cfe_msg.h
```diff
@@ -0,0 +1,220 @@
...
+**          - If the underlying implementation of software bus messages does not
+**            include a checksum field, then this routine will return a zero.
+**
+** \param[in]  MsgPtr      A pointer to the buffer that contains the software bus message.
+**                         This must point to the first byte of the message header.
+**
+** \return The checksum included in the software bus message header (if present), otherwise,
+**         returns a checksum value of zero.
+**/
+uint16 CFE_MSG_GetChecksum(CFE_SB_MsgPtr_t MsgPtr);
```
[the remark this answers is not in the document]
CDKnightNASA: > If we are creating a new/separate API layer this shouldn't assume that the checksum is a uint16. What if its bigger (i.e. CRC32/MD5/SHA1)? What if the message does not even have a checksum? There is no opportunity for a descriptive return code here.
> 
> While I understand this is only a first take, it would be OK if this was confined as an internal header file but it is being exposed in the public API so if users actually start calling this stuff, we'll be in the same boat where the API is limiting the implementation options.

Higher-level topic--what are checksums for, from the CFE application's point of view? I can see there being value in comparing two messages to see if they are identical, and possibly reporting a checksum in a log or event for debugging purposes, but I think just "returning a checksum" may be inappropriate. Another option would be caller passes in a (large enough) buffer and can use memcmp to see if two sums are the same.

fsw/cfe-core/src/inc/cfe_msg.h
CDKnightNASA: Also I've always wanted to typedef return status instead of just relying on int32/uint32, would now be a good time to introduce such a change?

fsw/cfe-core/src/inc/cfe_msg.h
```diff
@@ -0,0 +1,220 @@
...
+**                         This must point to the first byte of the message header.
+**
+** \return The command code included in the software bus message header (if present).
+**         Otherwise, returns a command code value of zero.
+**/
+uint16 CFE_MSG_GetCmdCode(CFE_SB_MsgPtr_t MsgPtr);
+
+/**@}*/
+
+/** @defgropu CFEAPIMSGTime cFE Message Time APIs
```
CDKnightNASA, line 161: "defgropu"

fsw/cfe-core/src/inc/cfe_msg.h
```diff
@@ -0,0 +1,220 @@
...
+**          messages.  It may be a checksum, a CRC, or some other algorithm.
+**
+** \par Assumptions, External Events, and Notes:
+**          - If the underlying implementation of software bus messages does not
+**            include a checksum field, then this routine will do nothing.
+**
+** \param[in]  MsgPtr      A pointer to the buffer that contains the software bus message.
+**                         This must point to the first byte of the message header.
+**/
+void CFE_MSG_GenerateChecksum(CFE_SB_MsgPtr_t MsgPtr);
```
CDKnightNASA, line 81: I'm wondering also, if we are funneling all header operations through API's, methinks the checksum should be automagic? I suppose you could argue that there's a performance hit to calculating every time a field is changed, but I don't know how many times I've written code that doesn't set the checksum.

jphickey: Indeed, when a packet format (container) is defined then a field can be separately designated as a checksum/error control field, so the framework can calculate it and the application software need not be concerned with it or really even know it exists.

Error control fields are mainly to cover data in transit between modules, it isn't worthwhile to (re-)calculate the checksum after every small data change.  App should just fill in _all_ data, and the framework should calculate the checksum/error control field at the time it is "sent", this is when you know the data shouldn't be changing anymore.

skliper: Agree with everything above.  This routine is envisioned as something SB could call when messages are "sent".  Apps never have to worry about it.

CDKnightNASA:
> > It's OK as a first step - but there are some things I'd definitely want to address before merging.
> 
> Excellent improvements, thanks!

Something else to consider--while 100%-1/infinity% of apps should be manipulating headers via the official methods, I have two cases (SBN, and an app I wrote years ago that has languished but should still be useful in the future called "ds_replay", which [as its name probably makes clear] takes ds-generated files and replays the messages back onto a bus) where I need to be able to push messages on to the software bus without having to go through all of the header fields.

Architecturally, these could be simply via a "private" API that SBN and ds_replay sneak through.

fsw/cfe-core/src/inc/cfe_msg.h
CDKnightNASA: > Agree with everything above. This routine is envisioned as something SB could call when messages are "sent". Apps never have to worry about it.

So would I be correct in understanding that we agree that all of the checksum API calls should be removed from the fsw/cfe-core/inc/*.h files?

Although I think there might be value in having utility functions to compare two messages to see if they're identical (whether by computing then checking the checksum or checking the entire message header/body via memcmp.)

jphickey: Incidentally, the proposed change that I had done (see PR #716) was basically this approach - provide a typedef for the error control value, a getter function, a compute function, and a compare function, all based on that type.

The other thing to consider is that the checksum actually changes (depending on algorithm) depending on whether the message needs to be padded for conformance with the CPU alignment requirements.

EDS, of course, handles all of this, if we choose to adopt it - checksums are computed once the message is in its final form for network transmission, using the EDS-specified algorithm.

fsw/cfe-core/src/inc/cfe_msg.h
skliper: > Also I've always wanted to typedef return status instead of just relying on int32/uint32, would now be a good time to introduce such a change?

Lets consider such a change as a separate issue/PR.

fsw/cfe-core/src/msg/cfe_msg_time.c
```diff
@@ -0,0 +1,134 @@
...
+    int32 Result = CFE_SB_WRONG_MSG_TYPE;
+
+    CFE_SB_TlmHdr_t *TlmHdrPtr;
+
+    /* declare format specific vars */
+    #if (CFE_MISSION_SB_PACKET_TIME_FORMAT == CFE_MISSION_SB_TIME_32_16_SUBS)
+    uint16 LocalSubs16;
+    #elif (CFE_MISSION_SB_PACKET_TIME_FORMAT == CFE_MISSION_SB_TIME_32_32_M_20)
+    uint32 LocalSubs32;
+    #endif
```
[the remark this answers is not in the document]
skliper: Yup, these will go away eventually since we'll be able to source select with the updates.  Minimizing changes this round just to get the capabilities in place.

skliper:
> Architecturally, these could be simply via a "private" API that SBN and ds_replay sneak through.

We can certainly provide an API that just puts the unmodified message on the bus (isn't this the intent of CFE_SB_PassMsg?)  Vs SendMsg which could be signing the message or adding checksum, or whatever modifications get plugged in under the hood.

fsw/cfe-core/src/inc/cfe_msg.h
skliper: Fixed in latest commit

fsw/cfe-core/src/msg/CMakeLists.txt
skliper: Replaced with more standard implementation in b5c8be6, also messed with includes to make it easier for SB to depend on MSG (or custom MSG) once it moves to it's own module.

fsw/cfe-core/src/inc/cfe_msg.h
skliper: All great suggestions, but in the short term just moved cfe_msg.h internal to the msg module in b5c8be6.  I will do a little bit of fix-up, but there will be plenty of opportunities for further improvement.

jphickey:
> > Architecturally, these could be simply via a "private" API that SBN and ds_replay sneak through.
> 
> We can certainly provide an API that just puts the unmodified message on the bus (isn't this the intent of CFE_SB_PassMsg?) Vs SendMsg which could be signing the message or adding checksum, or whatever modifications get plugged in under the hood.

Note that messages being passed around locally on the same CPU in the same memory space don't really need the checksum at all.  It's only required for when the message is sent to some other entity via some sort of interface (SBN, CI, TO, etc).

For this I think the notion of a "subscription flag" works well -- when subscribing to a MID one can indicate whether the intent is to locally cast it to a C struct and interpret/use the data (i.e. as a consumer, the default) or whether the intent is to forward it to another entity across a network (i.e. as a broker).  The former needs alignment padding to allow it to be cast/accessed directly as a C data type, but the latter would get a "packed" version that removes the padding and can also recompute the checksum.  (as I noted earlier the checksum of a padded/aligned message will likely be different than an unpadded/unaligned one).

fsw/cfe-core/src/inc/cfe_msg.h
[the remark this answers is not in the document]
skliper: Good conversation for future development, for now moved internal to msg (still planning some minor cleanup but it won't solve everything).  b5c8be6

fsw/cfe-core/CMakeLists.txt
```diff
@@ -31,10 +31,18 @@ add_definitions(-D_CFE_CORE_)
...
+endif (NOT MISSION_CORE_MODULES)
+
 foreach(MODULE ${CFE_CORE_MODULES} config shared)
   aux_source_directory(src/${MODULE} CFE_ALL_MODULE_SRCS)
 endforeach(MODULE ${CFE_CORE_MODULES})
 
+foreach(MODULE ${MISSION_CORE_MODULES})
+  add_subdirectory(${MODULE} ${CFE_CORE_TARGET})
+endforeach(MODULE ${MISSION_CORE_MODULES})
+
```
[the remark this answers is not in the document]
skliper: Implemented suggestion in b5c8be6

fsw/cfe-core/CMakeLists.txt
```diff
@@ -31,10 +31,18 @@ add_definitions(-D_CFE_CORE_)
 set(CFE_CORE_MODULES es sb evs tbl time fs)
 set(CFE_ALL_MODULE_SRCS)
 
+if (NOT MISSION_CORE_MODULES)
+  set(MISSION_CORE_MODULES ${cfe-core_MISSION_DIR}/src/msg)
+endif (NOT MISSION_CORE_MODULES)
+
```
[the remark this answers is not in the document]
skliper: Implemented suggestions in b5c8be6

skliper:
Ok, b5c8be6 is the next step.

Still on the docket - minor cleanup of MSG future APIs, refactor of ccsds.h to make replacement of secondary header easy, further progression towards msg being a "module"... as in have an inc_public, inc_private.  There is quite a bit of possible code cleanup, not sure how much I'll get to.  Trying to limit to what's required or at least only changing what bugs me the most.

Unit tests not fixed yet.  Did test with a custom mission_msg and it worked well (both append and replace).  Eventually recommending a transition of the rest of the core over once a good pattern is established.

Could have modules own their stubs, unit tests, EDS, etc.  The directory tree is begging for a simplification (could move modules to cfe/ !)

skliper:
> Note that messages being passed around locally on the same CPU in the same memory space don't really need the checksum at all. It's only required for when the message is sent to some other entity via some sort of interface (SBN, CI, TO, etc).

I have had seen requirements to checksum at the source and verify before any use (within the same memory space).  Radiation, security, etc, so I'd prefer if we don't preclude the option of checksumming before it goes on SB.

> For this I think the notion of a "subscription flag" works well -- when subscribing to a MID one can indicate whether the intent is to locally cast it to a C struct and interpret/use the data (i.e. as a consumer, the default) or whether the intent is to forward it to another entity across a network (i.e. as a broker). The former needs alignment padding to allow it to be cast/accessed directly as a C data type, but the latter would get a "packed" version that removes the padding and can also recompute the checksum. (as I noted earlier the checksum of a padded/aligned message will likely be different than an unpadded/unaligned one).

I  like the "subscription flag" concept.  I'd like to stay flexible in terms of what format is actually on SB, since I can see benefits from each approach.   If it's "packed" on the bus, it can get the checksum or signature at publish, and any subscriber can authenticate prior to use (nobody allowed to modify the packed message).  Then the packing is owned under the hood of the publish, vs packing owned by each endpoint.  Others may want local representation for performance reasons to avoid packing/unpacking if there is extensive local use of the message.

skliper:
Whoops... missed files in my commit, stand by.

skliper:
Now ready for a 2nd look if interested.  This update actually allows me to do the header replace and implementation customization without additional changes... 844f8af

fsw/cfe-core/CMakeLists.txt
```diff
@@ -37,6 +37,22 @@ endforeach(MODULE ${CFE_CORE_MODULES})
...
+  elseif(IS_DIRECTORY ${MISSION_SOURCE_DIR}/${MODULE})
+    add_subdirectory(${MISSION_SOURCE_DIR}/${MODULE} ${MODULE_NAME})
+  else()
+    message(FATAL_ERROR "Target build incomplete, ${MODULE} not found in ${MISSION_SOURCE_DIR} or ${cfe-core_MISSION_DIR}/src")
+  endif()
+
+  message (STATUS "Adding mission core module: " ${MODULE})
+  target_link_libraries(${CFE_CORE_TARGET} ${MODULE_NAME})
+endforeach(MODULE ${MISSION_CORE_MODULES})
+
```
jphickey, line 55: In my previous comments when I said "let the build system figure out the absolute path" ... I meant using the existing logic for doing so, not to add a new search function.

This message module shouldn't be treated specially or be any different than any other module, except for how its linked into the final product.

My previous commit in PR #720 covers this.  Can we look again/discuss integrating these and avoiding having another parallel module search implementation?

cmake/mission_build.cmake
```diff
@@ -40,6 +40,11 @@ function(initialize_globals)
       set(MISSION_SOURCE_DIR ${MISSION_SOURCE_DIR} CACHE PATH "Top level mission source directory")
     endif(NOT DEFINED MISSION_SOURCE_DIR)
 
+    # Default core modules, set early so it can be appended or replaced from user configuration
+    if (NOT DEFINED MISSION_CORE_MODULES)
+      set(MISSION_CORE_MODULES msg CACHE PATH "Mission core modules")
+    endif (NOT DEFINED MISSION_CORE_MODULES)
+
```
jphickey, line 47: This shouldn't be a cache variable (it wasn't before, right?)

Cache is used for when you want to retain a user-specified config from run to run.  Values like this are read from a file (e.g. targets.cmake) so there is no need to cache - its in the file.  It can actually get rather weird when mixing cache and non-cache values, as they both coexist even with the same name - one hides the other - and behavior can be confusing when you WANT to change it i.e. by updating your config and rebuilding - if cache is used then the old value will continue to be active.

fsw/cfe-core/CMakeLists.txt
skliper: I'm not a fan of: 

https://github.com/nasa/cFE/blob/d217ca39f1de74695f0d4e0eb427f90fbe8b1156/cmake/mission_build.cmake#L125-L171

I'd rather push off that discussion until we have an opportunity to more formally review and document.

jphickey: But that's all old/existing code -- it works for finding the apps and libs and everything else, even if they are out of tree.  Why wouldn't we just use it here?

PR #720 just reuses it for another module type, and I suggest doing the same here. 

If there is something wrong with it, I'm happy to fix/address that separately but we shouldn't add another implementation of the same thing.

jphickey: BTW, in addition to finding modules this is also responsible for setting the `${<modulename>_MISSION_DIR}` variable, which is important for any other module that depends on it so it can do e.g. `include_directories()` based on that.

The existing pattern is also to put things these user-specified modules in either `apps` or `libs` top level directories.  This appears to be allowing modules to exist as their own TLD.

I don't really care either way, but I don't see why for this type of module we would have a different path pattern than other types of user-specified modules.  Either one is fine but we should just be consistent in the paths.

cmake/mission_build.cmake
skliper: Good point, let me know what you think of 67dbcc4

fsw/cfe-core/CMakeLists.txt
skliper: I like the concept of a few "common" locations that can be extended by the user, and agree custom cfe core modules should follow a similar pattern, just seems like the implementation evolved to the point of being due a refactor or at minimum a careful review and documentation.  Rather than pile on I've gone simple for now.  We can fix prior to considering for an actual merge.

jphickey: We can certainly have that discussion about how to clean it up.  There is at least one aspect of that logic that can be simplified.    I don't think that's a reason to bypass it, though.

Problem is - it isn't just that one loop - from there a number of variables/lists are set and other parts of the build system need/expect these patterns to exist, and they won't be correct.  So there is negative fallout from _not_ following the same path/logic.  

Furthermore - the minimal change proposed in #720 to use the same logic is downright trivial compared to all the extra logic proposed here.  I even made it simpler (see commit jphickey/cfe@a921046) - as you don't strictly need the `arch_build.cmake` bits for include_directories - so it literally is only two lines that are actually needed to make this follow the same logic as every other module, which "just works" without the unintended consequences / inconsistencies.

We can talk about ways to clean up the search logic, but in the meantime I would strongly recommend against bypassing it..

skliper: Note still not transitioned yet.  I'm not following the resistance to using interface libraries, etc but if you've got a commit I can merge that implements your version I'll merge it in before I finish.

skliper:
Is the interface target concept acceptable to move forward with?  Makes it easier for apps to use the custom api's without having to explicitly link the custom include directories (which could quickly get messy).  I strongly prefer this approach over the current methods (bottom up adding to an interface target seems more maintainable than top down search of all possible include directories).  The custom api headers do have unique names (not enforced, but definitely the suggested pattern) so it's easy to tell which apps depend on them.

jphickey:
Yes, interface targets would (probably?) be fine to use, I haven't tried them yet though to see how it would change the CFE building/linking paradigm.  It would potentially clean up some logic where we set compile flags and include directories at the global scope.

skliper:
> Yes, interface targets would (probably?) be fine to use, I haven't tried them yet though to see how it would change the CFE building/linking paradigm. It would potentially clean up some logic where we set compile flags and include directories at the global scope.

Great, I'll use this as an opportunity to try them out.  If it looks good we could extend/update across the rest of the system or trash if it doesn't.

jphickey:
> Great, I'll use this as an opportunity to try them out. If it looks good we could extend/update across the rest of the system or trash if it doesn't.

I'm thinking it might be useful for the functional test app too, in order to provide the path to UT assert.  I may try it....

skliper:
Rebased on current master and fixed a minor bug.

astrogeco:
There's been a lot of movement on this since we last discussed. @skliper do you want to bring it up again?

skliper:
> There's been a lot of movement on this since we last discussed. @skliper do you want to bring it up again?

I need to rebase on #751 and clean up for it to match those established patterns.

skliper:
@jphickey The more I think about the scope/role of the core APIs (common APIs that cFE and many apps depend on) and the modules the more I favor the API header file remaining in the core (cfe/fsw/cfe-core/src/inc) vs existing in the module (cfe/modules/xxx).  A main point of the core API is that it is controlled, the point of the module is that it can be replaced/customized since it's the implementation of the API.  Packaging the core API with the implementation makes even less sense if we introduce a second implementation into the framework, for example if we want to introduce a new message table implementation that favors minimum size over performance but still maintain the sparse table implementation.  Which framework module would own the core API?  I don't think either of them should... since the core API is not implementation.  It would also get tangled in that one of the modules would need to add the other modules include directory that contains the core API... In the end I think based on scope/role (and historically since everyone is used to looking in the core for the API's) it just makes more sense to stick with the current model.  Also less of an extended search path seems like a good thing (don't have to search for core API's throughout all modules).  You could still create the interface target (doesn't matter where the file is), so it doesn't invalidate that approach (which I really do like)... but it still makes more sense to me scoped with the core.

Any other opinion's out there?  @acudmore @jwilmot @astrogeco?

jphickey:
Yeah, I was thinking about it more after the conversation, and I think it could make sense either way.  I'm not necessarily opposed to an API being defined in CFE, and I do agree it is necessary for apps.  

I still think that the most flexibility is achieved using a (notional) mapping layer of some type:

    CFE controlled API -> translation/mapping layer -> implementation API

The mapping layer can be a passthru/inline for typical/default cases so it doesn't need to cost anything overhead-wise.

So really - I think the answer to whether the API belongs at the module or CFE layer is "both".

This allows CFE to stabilize/control the API which is exposed to apps, using an API which is CM'ed with CFE, and an independent implementation of that down in some external module, with an API that is CM'ed in that module.  Then a shim layer "connects" the two.  

I do think having a passthru shim between them a beneficial thing to do as it allows for future decoupling if you need to add some backward-compatibility logic in the future.  In simple cases, the shim can be inline so overhead is next to nothing.

skliper:
I actually did implement it already with a layering concept... for example the MsgInit has generic code that implements the CFE API portion that is generic, but then utilizes an internal API to do the header initialization, which then calls the CCSDSPri and Ext initialization implementations.  This is already paying dividends since the internal API to do header initialization can be replaced by a custom initialization (that still meets the internal API such that the core MsgInit can still be used) that can also initialize the custom secondary header fields.  I didn't add a layer everywhere since I didn't think a layer as just a simple pass-through for MsgInit added anything... but it's trivial to add layers underneath wherever it's useful.  Really just left it up to module developers to layer however it makes sense.  I had a very specific use-case in mind, so I layered and grouped accordingly.  All can be updated if additional layering within the module makes sense.  I'm close at having the rebased version ready... still needs UT, stub updates, and deprecation cleanup (or just add back in what I removed/changed for testing).

skliper:
Rebased on IC (at risk...), adopted the #751 model.  Kudos to @jphickey, the updates in #751 made implementation much easier.  My one request is can we move modules into `cfe/fsw`?  They are fsw, and I really like at that level there are `modules` and `cfe-core`.

skliper:
To simplify review I did make the PR relative to the IC... we can change it back once IC is merged.

skliper:
Force pushed removal of a minor left-over.

skliper:
Force pushed to revert backing up cfe/cmake/arch_build.cmake too far (mistakenly) last week.  Rebased on IC.

skliper:
@jphickey could you also add the stubs for https://github.com/skliper/cFE/blob/fix711-msg-access/fsw/cfe-core/src/inc/cfe_msg_api.h (assuming your scripts help, if it's significant hand work I can do it).

jphickey:
> @jphickey could you also add the stubs for https://github.com/skliper/cFE/blob/fix711-msg-access/fsw/cfe-core/src/inc/cfe_msg_api.h (assuming your scripts help, if it's significant hand work I can do it).

FYI: rather than trying to push to your fork I just put the output of the script into a gist here:
https://gist.github.com/jphickey/0b32bd65b7723154544cf2b6a11a983f

Note some of these may overlap/duplicate custom ones because they are all called "CFE_MSG_" - this is one reason I had originally advocated for using a different name/prefix for customized routines and doing a wrapper/translator in SB/CFE.  Just something to watch out for - might cause duplicate symbol issues when linking UT.

skliper:
Perfect, thanks.

skliper:
Rebased on main

skliper:
@astrogeco @jphickey - I propose we review this for merge.  There are follow-on tickets require to really finalize/clean up, but this seems like a workable separation point.

See also (the cleanup tickets) - #777, #794, #795, #796

fsw/cfe-core/src/inc/cfe_sb.h
```diff
@@ -146,25 +147,37 @@
...
-    CCSDS_PriHdr_t      Hdr;   /**< \brief CCSDS Primary Header #CCSDS_PriHdr_t */
-    CCSDS_SpacePacket_t SpacePacket;
-    uint32              Dword; /**< \brief Forces minimum of 32-bit alignment for this object */
-    uint8               Byte[sizeof(CCSDS_PriHdr_t)];   /**< \brief Allows byte-level access */
-}CFE_SB_Msg_t;
+#ifndef CFE_OMIT_DEPRECATED_6_8
+
+typedef CFE_MSG_Message_t         CFE_SB_Msg_t;      /**< \brief Old style, member names won't match */
+typedef CFE_MSG_CommandHeader_t   CFE_SB_CmdHdr_t;   /**< \brief Old sytle, member names won't match */
+typedef CFE_MSG_TelemetryHeader_t CFE_SB_TlmHdr_t; /**< \brief Old style, member names won't match */
```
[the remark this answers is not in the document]
skliper: No longer deprecating these

fsw/cfe-core/src/inc/cfe_msg_hdr.h
```diff
@@ -0,0 +1,94 @@
...
+ * inherited for libs (just link the interface library)
+ */
+/**
+ * \brief Full CCSDS header
+ */
+typedef struct{
+    CCSDS_PrimaryHeader_t Pri;  /**< \brief CCSDS Primary Header */
+#ifdef MESSAGE_FORMAT_IS_CCSDS_VER_2
+    CCSDS_ExtendedHeader_t Ext;  /**< \brief CCSDS Extended Header */
+#endif
```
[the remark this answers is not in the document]
skliper: Subject of new ticket #796

fsw/cfe-core/CMakeLists.txt
skliper: Updated to built in module framework

CDKnightNASA:
Philosophical question -- There is no mention of CCSDS message format being a requirement for the SB in the requirements .csv, should there be a requirement? I think it's been just assumed that messages would always be CCSDS format, but what does that buy us, particularly if we want to force folks to use cFE-provided API's to read/write messages?

skliper:
SB shouldn't really care, as long as there is a concept of a message ID that is used for routing to each unique destination.  The message module is now set up such that we aren't forcing anyone into any specific format.  Only if your app uses specific CCSDS fields (without an implementation replacement to work with whatever format you are using) would it matter.

skliper:
Suggestion - possibly squash based on types of changes... or can do after review.

jphickey:
For squashing this, I suggest three commits:

1. Adding the new "msg" module (mostly new code)
2. FSW updates to existing CFE SB to use new module 
3. Unit testing

I also very much want to see an example of how this proposal works for common use-cases of overriding/extending the header.   For instance #795 removes the logic for the different timestamp formats and says users can use the modular method instead.  Which is fine, but how would that actually look?

If I (hypothetically) had a CFE mission based on bootes/6.8 that used one of the non-default header formats, how can I replicate the format with this new structure, by cloning the main branch and _only_ adding new code/config, not changing any code?

skliper:
Sent full internal example sent via email.  I can also add quick examples related to #795.  Will squash per suggestion.

skliper:
Squashed for review.

cmake/mission_defaults.cmake
```diff
@@ -14,6 +14,7 @@ set(MISSION_CORE_MODULES
     "cfe-core" 
     "osal"
     "psp"
+    "msg"
```
jphickey, line 17: Have we evaluated the possibility that the single `msg` perhaps should actually be three:

 - ccsds spacepacket (base)
 - ccsds extended
 - cfs cmd/tlm

I reviewed the internal "extension" library and noted that it is effectively replacing the _entire_ `msg` module with a different one, but then selectively re-compiling and referring to some (but not all) source files provided by the msg module, as well as giving itself an include path to the private headers within this module.

The paradigm should be that external entities don't depend on private headers - nor should source files be compiled by a script that isn't version controlled with the said source files.

It would seem, however, that if it the "msg" module were actually three different modules, then only one would need to be replaced for modifying the secondary header, and the other two (ccsds + extended) can be used as-is.

modules/msg/src/cfe_msg_ccsdspri.c
```diff
@@ -0,0 +1,352 @@
...
+
+    CFE_MSG_GetHeaderField(MsgPtr->CCSDS.Pri.StreamId, ApId, CFE_MSG_APID_MASK);
+
+    return CFE_SUCCESS;
+}
+
+/******************************************************************************
+ * Set message application ID - See API and header file for details
+ */
+int32 CFE_MSG_SetApId(CFE_MSG_Message_t *MsgPtr, CFE_MSG_ApId_t ApId)
```
jphickey, line 202: For "setter" API's, it might be worth considering passing by pointer (e.g. `const CFE_MSG_ApId_t*` rather than by value.  A pattern of passing by value kind of assumes that each logical "field" is just a single datum, like an int or a float.  If this becomes a struct or string this could get very inefficient.  But if it were a pointer then it could accept any data type.

skliper:
@tngo67 @excaliburtb - requesting feedback on suggestion to break up msg module.

Options:
 - Leave as is, customization implementations can either cherry-pick (fragile) or clone and own the entire module (to avoid dependence on core internals)
 - Break up implementations into building blocks, customization has the same option to cherry-pick or clone and own, but limited to just the block being customized
 - Or leave for now and move forward with merging this PR, open an issue to break up at some convenient time in the future?

modules/msg/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+#
+# Note this is different than applications which are dynamically
+# linked to support runtime loading.  The core applications all
+# use static linkage.
+#
+##################################################################
+
+# Add the basic set of files which are always built
+# Defined as absolute so this list can also be used to build unit tests
+set(${DEP}_SRC
```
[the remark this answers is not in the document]
skliper: I only see one suggestion which I'm fine with (document), but it seems like you had an alternative in mind?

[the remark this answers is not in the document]
jphickey: The alternative would be to use a fixed name (`msg`) and not make it variable.  Is there an expectation that this will be compiled under a different library name?  A compromise might be to set your own variable name at the beginning, and use that variable for the rest of the script.  Depends on the expected use-case for compiling the same library under a different name.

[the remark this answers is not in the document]
skliper: I'd think hard coding it would be a less desirable solution.  The name should match the module name such that the system can utilize common patterns.  I'd rather avoid setting up a potential for the name to not match.

tngo67:
> 
> 
> @tngo67 @excaliburtb - requesting feedback on suggestion to break up msg module.
> 
> Options:
> 
>     * Leave as is, customization implementations can either cherry-pick (fragile) or clone and own the entire module (to avoid dependence on core internals)
> 
>     * Break up implementations into building blocks, customization has the same option to cherry-pick or clone and own, but limited to just the block being customized
> 
>     * Or leave for now and move forward with merging this PR, open an issue to break up at some convenient time in the future?

Here's my 5-cent..
If the current implementation works, and we can get the issue resolved by v7.0 for Gateway, then I vote for option #3.  
The clone & own in option #1 seems to be more in-line with the current implementation approach for OSAL & PSP, correct?
Although option #2 has more flexibility, I'm not clear yet if it adds more complexity to the cFE build and certification.

skliper:
> in-line with the current implementation approach...

I see it as the same, in that using elements in osa/src/os/shared or portable in a custom OSAL requires the same "cherry-pick" approach.

astrogeco:
[CCB 2020-08-19](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-August-19-2020) APPROVED, open new issue to break down into small pieces in the future

excaliburtb:
I think splitting up the different forms of the ccsds header and corresponding elements is probably a case of over-abstraction. I would keep the implementation as-is and let a customer need drive more abstraction later, though they technically should have the tools they need with the current abstraction.

PR #731, opened 2020-06-03
fix #730 - adds padding to CFE_TBL_FileDef_t

### LINKED ISSUE #730: low CFE_MISSION_MAX_FILE_LEN and CFE_TBL_FILEDEF(..., "sample_app_table.tbl") causes silent non-truncation
**Describe the bug**
I was scratching my head why my table file was named "sample_app_table.tbl\004"  until I realized that CFE_MISSION_MAX_FILE_LEN is defined as "20" and "sample_app_table.tbl" is exactly 20 characters. If I had one more char, the compiler would err. But it accepts static definition of structs using strings that are the exact length, which loses the null terminator.

**Expected behavior**
Should allow file names of CFE_MISSION_MAX_FILE_LEN chars.

**Reporter Info**
Christopher.D.Knight@nasa.gov

CDKnightNASA:
**Describe the contribution**
Adds a uint32 padding (initialized with 0) to CFE_TBL_FileDef_t to ensure that the filename is null-terminated.

**Testing performed**
Builds fine with "sample_app_table.tbl" as the filename (which is exactly the length of the Filename field.)

**System(s) tested on**
Debian 10.3

Christopher.D.Knight@nasa.gov


CDKnightNASA:
Note that CFE_TBL_FILEDEF(), being a macro, does not have any unit tests. As such I have not put forward any tests. Perhaps for discussion whether tests should/could be developed.

fsw/cfe-core/src/inc/cfe_tbl_filedef.h
```diff
@@ -64,6 +64,12 @@ typedef struct
     char        TableName[CFE_MISSION_TBL_MAX_FULL_NAME_LEN]; /**< \brief Name of Table as defined onboard */
     char        Description[CFE_FS_HDR_DESC_MAX_LEN];       /**< \brief Description of table image that is included in cFE File Header */
     char        TgtFilename[CFE_MISSION_MAX_FILE_LEN];      /**< \brief Default filename to be used for output of elf2cfetbl utility  */
+
+    /** \brief  Padding to ensure that TgtFilename is null-terminated. If TgtFilename is initialized with a string of exactly
+     *          CFE_MISSION_MAX_FILE_LEN characters, the null of the initializer is lost and the string is no longer null-terminated.
+     */
+    uint32      Padding;
+
```
jphickey, line 72: I don't like that the reader is still overrunning the buffer and just happens to hit a zero "Padding" to stop it.  What about `TableName` and `Description` fields?  Wouldn't they have the same problem?

Recommend fixing the code that is actually reading this and overrunning the buffer.

CDKnightNASA: Technically it's not that the buffer is being overrun, it's that C allows you to do:

`struct foo { char a[10]; int b }; foo bar = {"0123456789", 10};`

If you do `foo bar = {"0123456789a", 10};` it throws a compiler error. So this is somewhat an edge case of having the exact wrong number of chars. At least for the other strings, having that exact wrong number results in a longer string but with the case of filename, it gave me a weird file when elf2cfetbl processed it.

The problem is that the CFE_TBL_FILEDEF is a macro that just defines a struct with a static initialization that is compiled into the symbol table. Otherwise we'd be talking about rearchitecting the table system.

I propose I add padding between all of the strings. While I could use int32/uint32, I don't see a reason to not use uint8/char?

CDKnightNASA:
> Note that SB API has a utility function to deal with this: `CFE_SB_MessageStringGet`
> 
> This is like a version of `strcpy()` that has fixed sizes for both input and output buffers and deals with the null termination issues on both sides (src + dest) in a more sane manner than `strncpy` does (or does not, in reality).
> 
> Even though this isn't a "message" per se, all FSW code can use this function when reading any fixed size string buffer. It's pretty generic. Recommend to use that for all of these fields.

See below, CFE_TBL_FILEDEF is a macro that maps to a static struct initialization, no code.

CDKnightNASA:
Note also that CFE_TBL_LoadFromFile fails if the table file name is exactly CFE_MISSION_MAX_FILE_LEN bytes long. (This is cause by `OS_TranslatePath()` rejecting because the OS_MAX_FILE_NAME is also set to 20. Should OS_TranslatePath() allow if`strlen(nameptr) == OS_MAX_FILE_NAME`?

fsw/cfe-core/src/inc/cfe_tbl_filedef.h
jphickey: Yeah, its a fixed-length string buffer, there should be no assumption that it has a null terminating character on it.  No fixed length buffers (either in messages or table files or anywhere else) should be assumed to be null terminated unless the reader makes a copy and enforces it.  That's just the nature of C strings.

Adding zero padding between every field is very hack-ish and doesn't really address the problem that `elf2cfetbl` (or whatever reads this struct) isn't properly reading these fields as fixed length string buffers.  No code should be allowed to overrun the buffer it is reading.  That's just bad.

It just means `elf2cfetbl` should copy the strings, adhering to the fixed length limit, before using them or passing them through to C library calls.  This is what CFE FSW does, all tools should do the same.

CDKnightNASA:
pushed an alternative approach of tacking on "\0" to all strings in the FILEDEF() macro, so if strlen is equal to the array size, it will generate a compiler error.

CDKnightNASA:
(I'll squash the commits prior to the CCB meeting, assuming this approach is acceptable.)

CDKnightNASA:
> The alternate take in [511b5a2](https://github.com/nasa/cFE/commit/511b5a26cb3cf5a01d0142442ddafec9a9e97f73) is better.
> 
> However I don't think it completely lets `elf2cftbl` off the hook -- when reading data from a file/structure you do not control, it must enforce null termination on strings. It's still a buffer overrun bug in that tool.
> 
> Also - just to confirm - if I remember correctly we do indicate in the mission config doxygen documentation that the actual length of these strings is one less than the configured value? If the intent is to allow all the way up to the configured value, then we need a "+1" in all the field lengths. However the FSW doesn't have a +1 so I think the practical limit is actually one less than the configured value.

Copy on elf2cfetbl I'll take a look at it today. WRT +1, yes OSAL barfs when I tried that. So yes, docs should make clear that those strings are all LEN-1.

CDKnightNASA:
> Also - just to confirm - if I remember correctly we do indicate in the mission config doxygen documentation that the actual length of these strings is one less than the configured value? If the intent is to allow all the way up to the configured value, then we need a "+1" in all the field lengths. However the FSW doesn't have a +1 so I think the practical limit is actually one less than the configured value.

See #735

astrogeco:
@CDKnightNASA can I remove the CCB label or do you still want to discuss this closed PR?

PR #737, opened 2020-06-08
Fix #734, continue cleanup of SB unit test

### LINKED ISSUE #734: Further simplify SB unit tests
Recently the software bus unit tests were updated to employ macros for common code bits, but this only amplifies the fact that the SB unit tests don't follow the typical UT assert model.

In particular, they "collect" a bunch of conditions together and then assert at the end that all the conditions where true.  There is no reason to do this deferred reporting with UT assert, and it only serves to obfuscate the true failure because the UT assert failure message line number can be quite different than what actually failed.

Now that the macros/wrappers are in place, rather than having them only mimic the old test logic, have them implement the preferred UT assert patterns.

- Remove `TestStat` global variable.  Do not keep global state outside UT assert.
- Remove separate text messages inside "if" conditionals and replace with UT assert statements.  That's what the UT assert API is for, after all.
- Remove `VERBOSE` compile-time option.  The UT assert has a runtime verbosity flag, just call `UtDebug` and the message will only be printed when set to verbose mode.

**Additional context**
This really just the next step along the path to using UT assert as it was intended, and removing the "compatibility crutches" that were put into place because CFE (and SB in particular) didn't employ the same test patterns as other apps/modules.

**Requester Info**

jphickey:
**Describe the contribution**
Makes the SB unit test closer to recommended UT assert patterns

- Do not keep a separate `TestStat` state variable outside UT assert.
- Do not report separate status messages from the asserts.  Use UT assert.
- Do not "reset" in the middle of a test routine, split into separate test routines where this is done.
- No need for `START` and `STARTBLOCK` or `ENDBLOCK` macros.  UT assert has messages for these test actions.  Each block should be a separate test routine and then these are unnecessary.

Fixes #734

**Testing performed**
Run all unit tests and confirm passing.  No FSW code changes here, only UT.

**Expected behavior changes**
SB Unit tests now use the UT assert framework more as intended.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This changes some items that were only reported on failure to a normal assert statement, which means its always reported.  The side effect is that there are now about 600,000 "test cases" in the SB test.  Most of that is from remaining validation of the length field, in particular the User Length test, which validates every possible uint16 value (65536), for each of the 4 header structures (x4), for both the return value and packet content (x2) totalling 524288 test cases for this one function.  There are similar examples for cmd code but as this is only an 8-bit field it only results in a few thousand cases.

If it is not desirable to have so many test cases, recommendation is to change the unit test to _not_ loop through every possible value.   No other functions are tested this way i.e. we do not call other APIs accepting a uint16 with every representable value.

### DIFF
--- a/fsw/cfe-core/unit-test/sb_UT.h
+++ b/fsw/cfe-core/unit-test/sb_UT.h
@@ -1385,7 +1385,7 @@ void Test_SB_Cmds_SubRptOff(void);
 **
 ** \par Description
 **        This function tests the command handler response to an invalid
-**        command code.
+**        command code in the CFE_SB_CMD_MID handler.
 **
 ** \par Assumptions, External Events, and Notes:
 **        None
@@ -1398,7 +1398,28 @@ void Test_SB_Cmds_SubRptOff(void);
 ** \sa #UT_Report
 **
 ******************************************************************************/
-void Test_SB_Cmds_UnexpCmdCode(void);
+void Test_SB_Cmds_CmdUnexpCmdCode(void);
+
+/*****************************************************************************/
+/**
+** \brief Test command handler response to an invalid command code
+**
+** \par Description
+**        This function tests the command handler response to an invalid
+**        command code in the CFE_SB_SUB_RPT_CTRL_MID handler
+**
+** \par Assumptions, External Events, and Notes:
+**        None
+**
+** \returns
+**        This function does not return a value.
+**
+** \sa #UT_Text, #SB_ResetUnitTest, #CFE_SB_InitMsg, #CFE_SB_SetCmdCode,
+** \sa #CFE_SB_ProcessCmdPipePkt, #UT_GetNumEventsSent, #UT_EventIsInHistory,
+** \sa #UT_Report
+**
+******************************************************************************/
+void Test_SB_Cmds_SubRptUnexpCmdCode(void);
 
 /*****************************************************************************/
 /**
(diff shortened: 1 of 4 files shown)


CDKnightNASA:
TestStat is/was useful for "falling through" if you have a failure, all other test calls would just be skipped. I am on the fence whether that's preferred or not--it makes it easier to find the root cause of the failure, rather than having to wade through all of the following errors, but there may be situations where we might want those calls to run even in the case of a failure (particularly teardown calls.)

jphickey:
> TestStat is/was useful for "falling through" if you have a failure, all other test calls would just be skipped. I am on the fence whether that's preferred or not--it makes it easier to find the root cause of the failure, rather than having to wade through all of the following errors, but there may be situations where we might want those calls to run even in the case of a failure (particularly teardown calls.)

Yeah, but this was the only module that did that.  If we want the concept we can add a feature to UT assert to exit a test function early and move on to the next test after the first failure.  But in most other modules it just proceeds with the remainder of the test. Sure, you might get a lot of superfluous "FAIL" cases due to the first failure but that's not really a problem - the rule of thumb when debugging a UT issue should be to focus on the first reported failure.

I definitely wouldn't say there was ever any confusion about where to focus attention within a swath of failure messges - it's the always the first reported problem.

fsw/cfe-core/unit-test/sb_UT.h
```diff
@@ -3242,7 +3285,10 @@ void Test_CFE_SB_SetGetMsgId(void);
 ** \sa #UT_GetActualPktLenField, #UT_Report
 **
 ******************************************************************************/
-void Test_CFE_SB_SetGetUserDataLength(void);
+void Test_CFE_SB_SetGetUserDataLength_Cmd(void);
```
CDKnightNASA, line 3288: dumb question - why are individual test fn's defined in the .h file? It's not like anything else is calling them...

PR #738, opened 2020-06-11
fix #730 - adds null to FILEDEFs

### LINKED ISSUE #730: low CFE_MISSION_MAX_FILE_LEN and CFE_TBL_FILEDEF(..., "sample_app_table.tbl") causes silent non-truncation

CDKnightNASA:
**Describe the contribution**
Fix #730 - Adds "\0" in the CFE_TBL_FILEDEF() macro so that if a string is exactly the size of the field, it will produce a compiler error.

Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_tbl_filedef.h
+++ b/fsw/cfe-core/src/inc/cfe_tbl_filedef.h
@@ -68,6 +68,12 @@ typedef struct
 } CFE_TBL_FileDef_t;
 
 /** The CFE_TBL_FILEDEF macro can be used to simplify the declaration of a table image when using the elf2cfetbl utility.
+**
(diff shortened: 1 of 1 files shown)


astrogeco:
CCB 2020-06-17: APPROVED

Issue #744, opened 2020-06-17
SB Pipe Info 'LastSender' Field is Unused 


dmccomas:
**Describe the bug**
The LastSender field in in the file produced by a SB 'Send Pipe Info to file' command is always zero. After doing some seraches it appears that this field defined in CFE_SB_PipeD_t in cfe_sb_priv.h is never set.

**To Reproduce**
1. Send a SB CFE_SB_SEND_PIPE_INFO_CC command
2. Transfer the file to the ground
3. Decode the binary file 

**Expected behavior**
The structure definition does not have any comments as to what to expect in this field. I would expect the AppId of the of the sender of last message send to the particular pipe.

**Code snips**
n/a

**System observed on:**
 Wndows VM running Ubuntu 18.04
 cfE 6.71, OSAL 5.0.1.0, PSP 1.4.0.0

**Additional context**
None

**Reporter Info**
David McComas, NASA Emeritus

jwilmot:
Is the just the reporting or does CFE_SB_GetLastSenderId API also not work?

dmccomas:
Just the reporting. It looks like the LastSender field is never referenced
by anyone.  CFE_SB_GetLastSenderId doesn't reference that field.

CDKnightNASA:
So cfe_sb.h defines:
```
typedef struct {
    uint32  ProcessorId;/**< \brief Processor Id from which the message was sent */
    char    AppName[OS_MAX_API_NAME];/**< \brief Application that sent the message */
} CFE_SB_SenderId_t;
```

The definition in cfe_sb_priv.h does not define what the LastSender field should be, but I'm guessing ProcessorId is not all that helpful. (Even SBN-relayed messages would show up with the local ProcessorId.)

I will implement as AppId as that seems the most sensible. I'll also add AppId to CFE_SB_SenderId_t as well, and at some point it might make more sense to deprecate AppName from this struct. Note that SBN uses the GetLastSenderId() API to ensure that it doesn't end up with an echo problem.

skliper:
This is OBE by #784 which deprecates it.

PR #750, opened 2020-06-19
Fix #745 - getlastsenderid now returns appid

### LINKED ISSUE #745: CFE_SB_GetLastSenderId() returns a pointer to a shared data location
CFE_SB_GetLastSenderId() in cfe_sb_api.c returns a pointer to shared data location which is risky. The function description also warns of some unpredictable or at least inconsistent behavior. 

I think the use cases for this function should be identified and a safer & behaviorally consistent implementation should be implemented. 
Are the use cases?
1.  Determine if app X send the last message 
2.  Determine whether the last message sent is from a lsit of apps

A couple of options are
1. Caller passes in the name of an app or a list of apps to SB determines whether the last message sent is a match.
2. The current type implementation copies the last message information rather than returning a pointer. 

**Additional context**
n/a

**Requester Info**
David McComas, NASA Emeritus

CDKnightNASA:
Fix for #745 - This removes the CFE_SB_SenderID_t struct and, instead, the CFE_SB_GetLastSenderId() function now takes a pointer to an AppId (uint32) that it uses as an out parameter for the call.

**Testing performed**
Built with ENABLE_UNIT_TESTS

**Expected behavior changes**
Significantly changes the GetLastSenderId() function, will break any users of this API (particularly SBN.)

**System(s) tested on**
Debian 10.3

**Additional context**
This obviates #744 and I recommend closing that ticket with this fix.

Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_sb.h
+++ b/fsw/cfe-core/src/inc/cfe_sb.h
@@ -208,16 +208,6 @@ typedef  struct {
 extern CFE_SB_Qos_t CFE_SB_Default_Qos;/**< \brief  Defines a default priority and reliabilty for off-board routing */
 
 
-/** \brief Message Sender Identification Type Definition
-**
-** Parameter used in #CFE_SB_GetLastSenderId API which allows the receiver of a message
-** to validate the sender of the message.
-**/
-typedef struct {
-    uint32  ProcessorId;/**< \brief Processor Id from which the message was sent */
-    char    AppName[OS_MAX_API_NAME];/**< \brief Application that sent the message */
-} CFE_SB_SenderId_t;
-
 /****************** Function Prototypes **********************/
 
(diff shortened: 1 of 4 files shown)


skliper:
Marked as 7.0 since this is an API change.  Does this also fix #744?  I may have missed it, but I don't see where this field is filled in.

CDKnightNASA:
> Marked as 7.0 since this is an API change. Does this also fix #744? I may have missed it, but I don't see where this field is filled in.

Gah forgot that part! I've committed/pushed. (I can squash merge if you need.)

CDKnightNASA:
Closing and will create a new ticket removing the GetLastSenderId() API and adding a variant of RecvMsg() that returns the AppId as an out parameter.

astrogeco:
CCB-20200624 - @CDKnightNASA will set up a splinter to further discuss the design.

PR #751, opened 2020-06-19
Fix #746, CMake mission dependency cleanup

### LINKED ISSUE #746: Consolidate implicit entries in MISSION_DEPS to defaults file
The build currently sets `MISSION_DEPS` as a list of dependencies which is basically the CFE core and OSAL by default.  There is a desire to consolidate all "fixed" entries into a separate defaults file for increased visibility and manageability.

These two dependencies can be consolidated into the `MISSION_CORE_MODULES` default list being proposed in PR #740, but this also depends on #728 so it needs to be implemented separately.

**Additional context**
This is a dependency of both PR #728 and #740.  Original discussion here:
https://github.com/nasa/cfe/pull/740#discussion_r441589460

**Requester Info**

jphickey:
**Describe the contribution**

Put more dependencies into the "mission_defaults.cmake" file for more visibility and ease of configuration.  This now includes all "implicit" modules such as cfe-core, osal, and psp.
    
Also push the calls to "generate_config_includefile" to a sub-script which can be distributed with each app and evaluated as part of the build.  This reduces dependencies on special naming conventions like `fsw/mission_inc` and `fsw/platform_inc`, and app developers can explicitly manage the files that users are expected to override.

Fixes #746 

**Testing performed**
Build and execute CFE and sanity test
Build and execute all unit tests

**Expected behavior changes**
No impact to behavior, affects build system only.
Should be fully backward compatible - in that the defaults are applied if a user has nothing specifically configured in their `targets.cmake` file.  The defaults will select osal, cfe-core, and psp as they always have.  However, the user now has the option to explicitly configure/control the inclusion of these modules and also provide mission-specific search paths to override them as desired.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Related to previous PR #720, #728, and #740.  Not easy to separate this out from those so this is currently based on a merge of those previous changes.  Hopefully the combined solution can be acceptable.


jphickey:
Note - only commit c80294d is relevant here to this PR.

If acceptable then all three of these can be rebased and merged together.

astrogeco:
CCB-20200624: APPROVED. Will open an issue to document the cmake build system.

skliper:
I'd like to start using this change ASAP, is it ready?  Can it be merged to IC soon or should I pull it into my work at risk?

skliper:
Basically I want to try out the generate_config_includefile to facilitate the message module header redefinition, to avoid multiple header files of the same name in the search paths since it's prone to error/confusion.

jphickey:
It's ready just need to decide how to handle the baseline/merge ... one option is to rebase (and close) the previous #728 and #740 and put everything into this PR.  Should I do that?

skliper:
I'm good with that approach...

jphickey:
Commit 09cbe08 rebases this change and all dependencies to latest master, should be just a simple merge, ready to go now.  This supersedes the two previous PRs (will close).

skliper:
Did this break how native_osconfig.cmake (in sample_defs) gets appended to default_osconfig.cmake?  I just saw a startup error... haven't dug into it much but the generated file doesn't seem to have the additional line from the native file.

skliper:
Note I moved `set(OSAL_CONFIG_DEBUG_PERMISSIVE_MODE TRUE)` into default_osconfig.cmake as a quick fix and it worked.

jphickey:
Yep, I pushed a patch in commit ab995786 to fix.  Needed to treat TARGETSYSTEM as a list, I hadn't noticed because - for other reasons - I had my default config set to permissive.

Issue #95, opened 2019-09-30
Suspicious implementation of SHORT_FORMAT mode in EVS_SendPacket()


skliper:
This code sequence occurs within the {{{EVS_SendPacket()}}} function:

{{{
/* (LSW) Is the intent to write the event text to the log but not the SB msg ??? */

   if (CFE_EVS_GlobalData.EVS_TlmPkt.Payload.MessageFormatMode == CFE_EVS_SHORT_FORMAT)
   {
      /* Send an empty message if short format is enabled */
      EVS_PktPtr->Payload.Message[0] = '\0';

/* (LSW) This is pointless -- why bother to send a buffer with an empty string ??? */

   }
}}}

It appears that someone (LSW?) already noticed the strangeness here some time prior to the 6.4.1 release.

The intent here may have been to send only the Event ID and omit the actual string, since most event ID's have fixed strings with them.

However, the length of the actual packet (in the CCSDS header) is never adjusted, so although the first character of the message payload is overwritten with a NUL, **the full message payload is still sent on the software bus** so there is absolutely no benefit to doing this in the current form.

skliper:
_Imported from trac issue 64.  Created by jphickey on 2015-06-04T12:54:04, last modified: 2019-03-05T14:58:28_

skliper:
_Trac comment by jphickey on 2015-06-04 12:54:27:_

Candidate for CCB discussion

skliper:
_Trac comment by glimes on 2016-11-08 14:17:08:_

Current crop of cfe-next are all going into CFE 6.6

skliper:
_Trac comment by sstrege on 2017-10-04 18:57:35:_

Commit [changeset:0a4ff97] (branch: sls-ic-trac64-merge) is ready for CCB review

skliper:
_Trac comment by cdknight on 2017-10-05 10:48:56:_

recommend approve

skliper:
_Trac comment by jphickey on 2017-10-05 11:43:54:_

My concerns are:

A nitpick: I think overwriting the first char of the message with a null is unnecessary (line 366-367) and I'd rather see this taken out.

Bigger issue: how will ground systems and/or other subscribers handle this size change?

Currently there is no precedent (as far as I know) for dynamically-sized telemetry messages.  Basically, a MsgId is associated with a certain C structure which defines its contents, and the size is either right or wrong.  (of course for commands it is the combination of MsgId and Command Code - but this still maps to a certain C data type).

[...]

PR #758, opened 2020-06-23
Fix95, CI build multiple targets

JandlynBentley-at-NASA:
**Describe the contribution**
Added a CPU2 as a second target to the build configuration to ensure build system is functioning properly with Travis CI for multiple targets.

Also added the following files:
- cpu2_cfe_es_startup.scr
- cpu2_platform_cfg.h
- cpu2_msgids.h

Resolves: cFS bundle's issue 95 <https://github.com/nasa/cFS/issues/95>

**Testing performed**
Went through the build steps from "Build and Run" section of <https://github.com/nasa/cFS>

**Expected behavior changes**
Expected to see cpu2 in the exe directory, but it did not appear; still only cpu1 was made for me. 
However, this worked for Chris Knight after he cloned cFS and pulled from my fork (cpu1 and cpu2 were made).

**System(s) tested on**
 - Hardware: Mac OS
 - OS: Ubuntu 18.04 via VirtualBox

Jandlyn Bentley, NASA-GSFC

### DIFF
--- a/cmake/sample_defs/targets.cmake
+++ b/cmake/sample_defs/targets.cmake
@@ -90,11 +90,14 @@ SET(TGT1_FILELIST cfe_es_startup.scr)
 
 # CPU2/3 are duplicates of CPU1.  These are not built by default anymore but are
 # commented out to serve as an example of how one would configure multiple cpus.
-#SET(TGT2_NAME cpu2)
-#SET(TGT2_APPLIST sample_app ci_lab to_lab sch_lab)
-#SET(TGT2_FILELIST cfe_es_startup.scr)
+
+# CPU2 is uncommented to test Travis CI's ability to build multiple targets, ensuring
+# the build system is functioning properly.
+SET(TGT2_NAME cpu2)
+SET(TGT2_APPLIST sample_app sample_lib ci_lab to_lab sch_lab)
+SET(TGT2_FILELIST cfe_es_startup.scr)
 
 #SET(TGT3_NAME cpu3)
-#SET(TGT3_APPLIST sample_app ci_lab to_lab sch_lab)
+#SET(TGT3_APPLIST sample_app sample_lib ci_lab to_lab sch_lab)
 #SET(TGT3_FILELIST cfe_es_startup.scr)
 
--- a/cmake/sample_defs/cpu2_cfe_es_startup.scr
+++ b/cmake/sample_defs/cpu2_cfe_es_startup.scr
@@ -1,0 +1,30 @@
+CFE_LIB, /cf/sample_lib.so,  SAMPLE_LibInit,  SAMPLE_LIB,    0,   0,     0x0, 0;
+CFE_APP, /cf/sample_app.so,  SAMPLE_AppMain,  SAMPLE_APP,   50,   16384, 0x0, 0;
+CFE_APP, /cf/ci_lab.so,      CI_Lab_AppMain,  CI_LAB_APP,   60,   16384, 0x0, 0;
+CFE_APP, /cf/to_lab.so,      TO_Lab_AppMain,  TO_LAB_APP,   70,   16384, 0x0, 0;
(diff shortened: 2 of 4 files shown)


JandlynBentley-at-NASA:
In my original commits from git, it looks like it might be linking to the cFE issue #95 which is not correct. This pull request is intended to resolve <https://github.com/nasa/cFS/issues/95>.

cmake/sample_defs/targets.cmake
```diff
@@ -90,11 +90,14 @@ SET(TGT1_FILELIST cfe_es_startup.scr)
...
 # commented out to serve as an example of how one would configure multiple cpus.
-#SET(TGT2_NAME cpu2)
-#SET(TGT2_APPLIST sample_app ci_lab to_lab sch_lab)
-#SET(TGT2_FILELIST cfe_es_startup.scr)
+
+# CPU2 is uncommented to test Travis CI's ability to build multiple targets, ensuring
+# the build system is functioning properly.
+SET(TGT2_NAME cpu2)
+SET(TGT2_APPLIST sample_app sample_lib ci_lab to_lab sch_lab)
+SET(TGT2_FILELIST cfe_es_startup.scr)
```
skliper, line 98: Don't change the default behavior of cFE out of the box.  This should be part of CI if you want to check 2 targets with CI.

skliper:
> In my original commits from git, it looks like it might be linking to the cFE issue #95 which is not correct. This pull request is intended to resolve [nasa/cFS#95](https://github.com/nasa/cFS/issues/95).

You need to open an issue in this repository specific to the changes you are doing here.  As noted above, don't change the default, out-of-the-box behavior of cFE/cFS.  You can add the sample configuration files (which is what I recommend you write the issue to do), but the actual configuration change to build 2 targets should be done with CI from cFS.

JandlynBentley-at-NASA:
I didn't see how to make adjustments in .travis.yml because it looks like CI only does one build for all targets.

skliper:
> I didn't see how to make adjustments in .travis.yml because it looks like CI only does one build for all targets.

The CI in nasa/cFE doesn't build anything.  That's why the issue is in cFS, at which it does a matrix build of the entire framework.  I recommend from cFS, write a script to reconfigure the system to build 2 CPUs (the part that I said don't do in this repo), and add another build to the cFS matrix that calls that script and does the two CPU build.

skliper:
In fact it may make more sense to not add these files, since they are (I assume?) just a copy of the cpu1 files.  Rather just copy them from cpu1_* as part of the script.  Otherwise we'd now have 2 sets of files that we need to keep up to date.  The copy is trivial and I'd think it would always work if just building.

CDKnightNASA:
We currently have targets.cmake with (commented out) CPU2 & CPU3, either we should remove those entries and have .travis.yml create entries or we should have all of the relevant files in sample_defs to support CPU2/3 if they are uncommented and just have .travis.yml uncomment the entries. We can discuss at CCB today?

JandlynBentley-at-NASA:
The only reference I could find regarding multiple targets in the Travis CI documentation was here: <https://docs.travis-ci.com/user/build-stages/#naming-your-build-stages>

It looks like we'd have to make two scripts?

CDKnightNASA:
> The only reference I could find regarding multiple targets in the Travis CI documentation was here: https://docs.travis-ci.com/user/build-stages/#naming-your-build-stages
> 
> It looks like we'd have to make two scripts?

What I suggest you do is have travis alter the targets.cmake and create the addl. .scr and .h files, then just do a single build (which will create multiple CPU's).

jphickey:
Note you can put targets inside an "if" statement like this:

```
if (TEST_TARGETS GREATER 1)
   set(TGT2_NAME cpu2)
   ...
endif()
```

Note this isn't a pattern I'd recommend projects clone for their own mission config, but it works for CI.  We can also have CI-specific "defs" directories stored in a separate CI repo.  Ideally for an extended validation test we should build with other CFS apps, and old/imported configs like real projects would.  This may not always work, but we should know if we broke it, as it means we probably need to add something in the release notes to guide users that are upgrading.

Issue #292, opened 2019-09-30
Update CFE_SB_SendMsgFull() to match latest SBN function prototype


skliper:
Reported by Astrobotic 12/28/18 via email.

The SBN product has a prototype of int32  CFE_SB_SendMsgFull(CFE_SB_Msg_t   *MsgPtr, uint32 TlmCntIncrements, uint32 CopyMode, CFE_SB_SenderId_t *SenderPtr) (4 variables)

The SB prototype has 3 int32  CFE_SB_SendMsgFull(CFE_SB_Msg_t   *MsgPtr, uint32 TlmCntIncrements, uint32 CopyMode);  There is a conflict between the header files.

skliper:
_Imported from trac issue 261.  Created by dcmccom3 on 2019-01-31T12:47:52, last modified: 2019-07-03T12:48:08_

skliper:
_Trac comment by jhageman on 2019-07-03 12:48:08:_

Moved unfinished 6.6.1 issues to next minor release

skliper:
Needs more investigation/documentation and a design/architectural discussion.  If the SBN call is different, it doesn't conform with the currently defined API.

skliper:
Related to https://github.com/nasa/cFE/issues/127, which is a request for this API change.

skliper:
Note the two references above for ic-20191115 were because of a link typo.  There was no work on this issue as part of that pull request.

skliper:
OBE - marking invalid

Issue #127, opened 2019-09-30
Add support to allow SBN to pass sender information across the network (GSFC DCR 22063)


skliper:
A cFE interface is needed to allow SBN to pass sender information (ProcessorID, Application Name) across a network.
 
One proposed change is a new private function called CFE_SB_PassMsgWithSenderID. This function will be needed by SBN ONLY. The CCB should discuss how this function can be used by SBN and blocked from being called/used by other applications. Analysis is needed to determine the best solution.

skliper:
_Imported from trac issue 96.  Created by sstrege on 2015-08-27T15:08:45, last modified: 2019-03-21T16:31:26_

skliper:
_Trac comment by cdknight on 2016-07-12 15:39:20:_

Currently SBN-injected messages have the application name of "SBN"--which SBN then uses to ensure that it ignores messages with the sender of "SBN", preventing loops. If the above functionality is approved and adopted, some other mechanism will be needed to ensure that SBN can identify which messages it generated versus which messages are from other applications in the cFE.

skliper:
_Trac comment by glimes on 2016-10-18 15:07:03:_

Remove my name from many tickets (where it was applied by default), so others can tell that there are tickets that need some attention.

skliper:
See also https://github.com/nasa/cFE/issues/292

skliper:
@CDKnightNASA is this enhancement request still valid?  It hasn't been updated in years, so I'm wondering if it's OBE.

CDKnightNASA:
In researching this, I noticed there's a SB global variable "SenderReporting", which is initialized but not referenced anywhere in cFE. I do think it would still be worth pursuing this, but I would need to give it some deep thought as it may be possible to determine by looking up the current task name.

skliper:
So one approach is to include the source information in the message, which is then a mission managed identifier that is unique.

CDKnightNASA:
> So one approach is to include the source information in the message (or message header), which is then a mission managed identifier that is unique.

True, I think this may actually be the preferred mechanism. I'm guessing that for > 90% of messages that are published, the publisher is known to the mission. In considering this, I will close this ticket.

jwilmot:
I know this is now  closed but this was part of SB and SBN in 2013. Not sure who took it out or why. It was intended for last sender, local and over SBN.  It was added DCR469:1 Added GetLastSender code ** Revision 1.68 2007/04/24 11:07:18EDT
From cfe_sb_api.c
[...]

PR #761, opened 2020-06-25
fix #759 - removes CFE_SB_GetLastSenderId() and adds CFE_SB_RcvMsgSenderId()

### LINKED ISSUE #759: remove GetLastSenderId() API and replace with a RecvMsg() API that returns the AppId of the sender of the message being received
**Describe the bug**

**Expected behavior**

**Reporter Info**
Christopher.D.Knight@nasa.gov

CDKnightNASA:
Fix #759

**Describe the contribution**
This removes the significantly-broken `CFE_SB_GetLastSenderId()` function and adds a new variant `CFE_SB_RcvMsgSenderId()` that returns (via an out parameter) the AppId of the application that sent the message being read off the pipe.

Note that, to reduce the redundant code, `CFE_SB_RcvMsg()` calls `CFE_SB_RcvMsgSenderId()` with a null pointer. It may be better to use a #define macro for improved performance, or deprecate the RcvMsg() API eventually.

**Testing performed**
Built and ran updated unit tests.

**System(s) tested on**
Debian 10.3

**Additional context**
GetLastSenderId() is basically broken, I suggest this be merged into 6.8, removing the function entirely (no deprecation). It will break any users of GetLastSenderId(), obv. (SBN being the only one I am aware of.)

Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_sb.h
+++ b/fsw/cfe-core/src/inc/cfe_sb.h
@@ -208,16 +208,6 @@ typedef  struct {
 extern CFE_SB_Qos_t CFE_SB_Default_Qos;/**< \brief  Defines a default priority and reliabilty for off-board routing */
 
 
-/** \brief Message Sender Identification Type Definition
-**
-** Parameter used in #CFE_SB_GetLastSenderId API which allows the receiver of a message
-** to validate the sender of the message.
-**/
-typedef struct {
-    uint32  ProcessorId;/**< \brief Processor Id from which the message was sent */
-    char    AppName[OS_MAX_API_NAME];/**< \brief Application that sent the message */
-} CFE_SB_SenderId_t;
-
 /****************** Function Prototypes **********************/
 
 /** @defgroup CFEAPISBPipe cFE Pipe Management APIs
@@ -574,7 +564,7 @@ int32 CFE_SB_UnsubscribeLocal(CFE_SB_MsgId_t MsgId, CFE_SB_PipeId_t PipeId);
 ** \retval #CFE_SB_MSG_TOO_BIG  \copybrief CFE_SB_MSG_TOO_BIG
 ** \retval #CFE_SB_BUF_ALOC_ERR \copybrief CFE_SB_BUF_ALOC_ERR
 **
-** \sa #CFE_SB_RcvMsg, #CFE_SB_ZeroCopySend, #CFE_SB_PassMsg
+** \sa #CFE_SB_RcvMsg, #CFE_SB_RcvMsgSenderId, #CFE_SB_ZeroCopySend, #CFE_SB_PassMsg
 **/
 int32  CFE_SB_SendMsg(CFE_SB_Msg_t   *MsgPtr);
 
@@ -607,7 +597,7 @@ int32  CFE_SB_SendMsg(CFE_SB_Msg_t   *MsgPtr);
 ** \retval #CFE_SB_MSG_TOO_BIG  \copybrief CFE_SB_MSG_TOO_BIG
 ** \retval #CFE_SB_BUF_ALOC_ERR \copybrief CFE_SB_BUF_ALOC_ERR
 **
-** \sa #CFE_SB_RcvMsg, #CFE_SB_ZeroCopySend, #CFE_SB_SendMsg
+** \sa #CFE_SB_RcvMsg, #CFE_SB_RcvMsgSenderId, #CFE_SB_ZeroCopySend, #CFE_SB_SendMsg
 **/
 int32  CFE_SB_PassMsg(CFE_SB_Msg_t   *MsgPtr);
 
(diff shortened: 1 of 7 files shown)


skliper:
Also likely needs a requirement's update, or at minimum review that the requirement is general enough to cover this change.

fsw/cfe-core/src/sb/cfe_sb_api.c
```diff
@@ -1320,12 +1320,7 @@ int32  CFE_SB_SendMsgFull(CFE_SB_Msg_t    *MsgPtr,
                 RtgTblPtr->SeqCnt);
     }/* end if */
 
-    /* store the sender information */
-    if(CFE_SB.SenderReporting != 0)
-    {
-       BufDscPtr->Sender.ProcessorId = CFE_PSP_GetProcessorId();
-       strncpy(&BufDscPtr->Sender.AppName[0],CFE_SB_GetAppTskName(TskId,FullName),OS_MAX_API_NAME);
-    }
+    BufDscPtr->SenderId = CFE_SB.AppId;
```
jphickey, line 1323: Doesn't this mean "SenderId" is always the CFE_SB AppID?

fsw/cfe-core/src/inc/cfe_sb.h
```diff
@@ -208,16 +208,6 @@ typedef  struct {
...
-/** \brief Message Sender Identification Type Definition
-**
-** Parameter used in #CFE_SB_GetLastSenderId API which allows the receiver of a message
-** to validate the sender of the message.
-**/
-typedef struct {
-    uint32  ProcessorId;/**< \brief Processor Id from which the message was sent */
-    char    AppName[OS_MAX_API_NAME];/**< \brief Application that sent the message */
-} CFE_SB_SenderId_t;
-
```
jphickey, line 220: I think `CFE_SB_SenderId_t` should remain as a struct. Ideally not even change the definition if we don't have to.

fsw/cfe-core/src/sb/cfe_sb_api.c
CDKnightNASA: > Doesn't this mean "SenderId" is always the CFE_SB AppID?

Yes.

CDKnightNASA:
> I concur that we should have some sort of requirements review here. This API has been broken for quite some time and it's not clear to me why a fix needs to be fast tracked suddenly, especially one that changes an important API.

I disagree that it changes an important API. GetLastSenderId() is broken and rarely used, and I'm adding a new RcvMsg() variant, not changing RcvMsg.

> If the intent is to avoid loops in SBN - does it even work for that? It might catch an A->B->A forwarding loop but what about A->B->C->A? Wouldn't a simple TTL (hop limit) be more effective and easier to implement?

SBN doesn't handle multiple hops, but if there are subscribers for MID X on CPU1 and CPU2, and CPU1 receives an X from CPU2 SBN publishes it and immediately receives it on its pipe. It uses GetLastSenderId() to ignore this new message so it doesn't try to send it back to CPU2.

> If we do decide to keep this API we need to make sure the output `CFE_SB_SenderId_t` type makes sense. Not sure a plain old uint32 AppId is enough because these are ephemeral. The string name, although more data to copy around, at least had the quality of being the same before + after a CPU gets restarted.

See my comments on the ticket to Jon, I think there are a variety of ways to protect the "Sender ID"--one way would be to not re-use the AppId unless the app name is the same (to handle when an app crashes or is otherwise restarted, it gets its old ID); and nobody else can get that ID unless they happen to pick the same name at just the right time.

jphickey:
OK, I see the issue with not wanting SBN to receive its own packets.  But couldn't that problem be solved in a more generic way?  Don't we (or if not, could we) have a pipe subscription option that tells SB not to directly loop packets back to itself?  Then we don't need to add an API at all.

This seems ugly in that (if I'm understanding correctly), SB will still be looping the packet back to SBN - SBN will then drop it - but it still actually puts the packet in the queue, taking up a slot, and waking up SBN again to read the message, only to figure out it was one of its own messages and drop it?  Seems less than ideal.  The message takes memory in the queue and buffer as it goes through this process, taking up space that could be used for real messages.

Also I'm wondering why this is an issue at this juncture - hasn't SBN existed for a while now?  What has changed that made this an issue that needs a fast resolution?

jwilmot:
The SB send local API was the way to avoid loopback. I assumed SBN used this. Now I am not sure as it appears GetLastSenderId() is used. Looks like I need to dig deeper to understand the issue.

CDKnightNASA:
> The SB send local API was the way to avoid loopback. I assumed SBN used this. Now I am not sure as it appears GetLastSenderId() is used. Looks like I need to dig deeper to understand the issue.

Yes, years ago I had the CFE_SB_PIPEOPTS_IGNOREMINE option added to pipe options.

https://github.com/nasa/cFE/blob/84ec281b4a432db7f41f0814e37fc239c723ad39/fsw/cfe-core/src/sb/cfe_sb_api.c#L1350

But looks like that never got incorporated into SBN. I can certainly add that capability. (And hopefully it [still] works. :D )

In thinking about the goal of "security", one option would be to have an app "private key" (a uint64 would be sufficiently large search space; could even be a hash of a string like app name) that, when it sends a message, is put into the buffer. Another app cannot read that message's private key but can provide it to a "validate message" API and it would get a true/false return. Then you'd just share the private key between the two apps who want to talk. Monitoring the number of calls to validate message, and killing any app that tries it too many times, would reduce attacks too.

Sorry a lot of this work has moved to my swap. I've been swapping it back into RAM but it's taking my brain a bit.

jphickey:
> But looks like that never got incorporated into SBN. I can certainly add that capability. (And hopefully it [still] works. :D )

Indeed, if this works (I think it should?) It would seem to be a way cleaner and more reliable way to avoid getting your own traffic back and having to drop it, without the extra roundabout.

I would also think that the real sender information (if we decide we need this API at all) would be propagated through SBN - so that at its final destination, it would reflect the original app on the other CPU - not SBN.

But the other use-case of message authentication/validity has been solved already - we should not need to invent a new method to authenticate messages, we should use an existing scheme/algorithm to do so.  

One option might be HMAC / RFC 2104:

https://www.ietf.org/rfc/rfc2104.txt
https://en.wikipedia.org/wiki/HMAC.

These HMAC codes can (for instance) be but into a custom header extension.  They are relatively simple and fast to compute/verify but require a pre-shared secret key so that might be a limiting factor.

CDKnightNASA:
> I would also think that the real sender information (if we decide we need this API at all) would be propagated through SBN - so that at its final destination, it would reflect the original app on the other CPU - not SBN.

That's an interesting problem, that will need some rumination.

> One option might be HMAC / RFC 2104:

HMAC should certainly be an option, but does mean you're computing across the entire message every send and receive and certainly should be enabled but used sparingly.

For simple "authentication", I thought of a fairly elegant solution--add a `SubscribePSK(CFE_SB_MsgId_t MsgId, CFE_SB_PSK_t PSK, CFE_SB_PipeId_t PipeId)` that would add another field/block of data to the subscription table. Publishing apps, when starting up, would `RegisterPSK(CFE_SB_PSK_t PSK)` and then any SendMsg() will automagically check the PSK of the publisher against the PSK in the DestinationD_t entry. Not changing any existing API's. If the PSK could be 3 bytes long, we have 3 bytes of 'spare' in DestinationD that we could use. :)

jphickey:
My suggestion is that for this PR/Issue we just deprecate the `CFE_SB_GetLastSenderId()` function as I think we all agree it's broken and doesn't really serve a useful purpose.

You'd just need to remove the use of this function from SBN - the IGNOREMINE flag is a better way, if it works.

AFAICT, there should be nothing else preventing us from deprecating this API.

__Going forward__ - we should have a requirements review of the perceived use-cases that were the impetus of this API, and come up with a real set of requirements FIRST - then we can implement something that fulfills those requirements.  I don't think we should rush to a new implementation here without reviewing the requirements first.

While there are plenty of ways to implement some sort of "simple authentication" that checks that a number matches, how is that different/better than the combination of MsgId+FunctionCode?

We need to define the specific risk factors that we are concerned about, and make sure that the solution adequately addresses those risk factors.

CDKnightNASA:
> My suggestion is that for this PR/Issue we just deprecate the `CFE_SB_GetLastSenderId()` function as I think we all agree it's broken and doesn't really serve a useful purpose.

That works for me. I'll put in a new PR this eve with the more limited set of changes.

> ...requirements FIRST...

WAAAAAAATTTTT?!?!? ;)

skliper:
Might also be worth looking at #292 and #127 again in this context...

PR #764, opened 2020-06-26
fix #759 - removes GetLastSenderId()

### LINKED ISSUE #608: CFE_SB_GetLastSenderID returns pointer to internal data, possible race
**Describe the bug**
CFE_SB_GetLastSenderID returns pointer to data that could be overwritten at any time

See https://github.com/nasa/cFE/pull/605#discussion_r407649126 for related comments

**To Reproduce**
Conceptually - app calls CFE_SB_GetLastSenderID, gets pointer, data gets overwritten, app takes action based on overwritten data vs original context

**Expected behavior**
No race.

**Code snips**
https://github.com/nasa/cFE/blob/60a5f6580d1e737f46efce0418aba5b8dac6df6a/fsw/cfe-core/src/sb/cfe_sb_api.c#L1651-L1657

**System observed on:**
N/A

**Additional context**
#605

**Reporter Info**
Jacob Hageman

### LINKED ISSUE #759: remove GetLastSenderId() API and replace with a RecvMsg() API that returns the AppId of the sender of the message being received

CDKnightNASA:
Fixes #759 
Fixes #608

Removes `CFE_SB_GetLastSenderId()`, which is supremely broken.

**Testing performed**
sb_UT passes

**Expected behavior changes**
Removes the CFE_SB_GetLastSenderId() API.

**System(s) tested on**
Debian 10.3

Christopher.D.Knight@nasa.gov


skliper:
Likely needs a requirements update.

CDKnightNASA:
> Likely needs a requirements update.

Done. I can squash prior to CCB if that's preferred.

Looking at https://github.com/nasa/cFE/blob/master/docs/cfe%20requirements.docx I have some architectural questions about the SB in general in a multi-processor environment. Perhaps we can discuss at a CCB meeting or have a splinter.

skliper:
@dmknutsen Heads up this change (assuming stakeholders accept it) will need to be reflected in JIRA, with impacts to test cases.

jphickey:
The requirement necessitates a different discussion than the API.  Reading that requirement text/rationale, it actually seems reasonable to me, but the problem is the implementation didn't match, so I'm not sure it should be removed.

The `CFE_SB_GetLastSenderId()` didn't actually implement the function described or fulfill this requirement.  It was tracking/returning the wrong thing and was not adequate for the rationale described in the requirement text.  That's why we need to remove/deprecate this existing API - it doesn't match.

Is it OK to keep a requirement but note that it is not met in the current version(s)?  This wouldn't really be a change - as `CFE_SB_GetLastSenderId()` has _never_ met this requirement - so in theory every CFE version to date has not actually met this requirement.

There are three things we need to do:
1. Remove/deprecate the existing API as it doesn't correctly implement/fulfill a requirement
2. Revalidate/discuss the requirement itself and determine if it is still applicable or needs an update
3. Implement whatever was the outcome of (2) above.

This PR should only do item 1, nothing else at this time.  There is certainly a requirements aspect to this, but I wouldn't jump to the conclusion of removing the existing requirement.

jphickey:
Also, if my understanding is correct, the requirements CSV file is an export from a Jira repository maintained internally by GSFC - so we should not change/modify the CSV file directly anyway.  

I'm good with commit 2a17ed2 alone.

astrogeco:
@CDKnightNASA want to discuss today?

astrogeco:
CCB 2020-07-01: APPROVED

CDKnightNASA:
replaced by #784

PR #772, opened 2020-07-08
CFE_SB_GetLastSenderId() future deprecation note

CDKnightNASA:
A short note that the CFE_SB_GetLastSenderId() is to be deprecated in a future release.

Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_sb.h
+++ b/fsw/cfe-core/src/inc/cfe_sb.h
@@ -1178,6 +1178,8 @@ CFE_TIME_SysTime_t CFE_SB_GetMsgTime(CFE_SB_MsgPtr_t MsgPtr);
 ** \brief Retrieve the application Info of the sender for the last message.
 **
 ** \par Description
+**          TO BE DEPRECATED, SEE BELOW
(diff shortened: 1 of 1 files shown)


astrogeco:
@skliper mind if we fast track this?

skliper:
@astrogeco I recommend getting concurrence from @jwilmot.  I'm not sure what this really buys us and future deprecation notes in the software deviates from how we've deprecated things in the past, why not just deprecate it (or open an issue to deprecate, and debate there?)  Is the issue that there is chance it won't be deprecated?  If it's pending a community/stakeholder review, maybe do that first?  I guess in short I'd rather just either deprecate it or not, vs add more documentation about something we might do someday.

astrogeco:
I like the idea of deprecating it outright. I believe this was in response to the conversation we had on CCB-2020-07-01 where we decided to undo the feature change but mark this as deprecated and add a note that it doesn't work as expected.

CDKnightNASA:
replace with deprecate

CDKnightNASA:
replaced with #784

Issue #265, opened 2019-09-30
Add cmake function to add cpus by function rather than assume consecutive cpu ids.


skliper:
Currently, the cmake rules require that cpus are built in order and will stop looking for cpu targets when a cpu idx is not found.

Propose that a cmake function can be executed from the targets.cmake file which explicitly adds the cpus to be built for the mission.

skliper:
_Imported from trac issue 234.  Created by tbrain on 2018-03-30T13:08:24, last modified: 2019-07-03T12:58:32_

skliper:
_Trac comment by sseeger on 2018-05-28 13:38:55:_

Thomas,

Just to be clear are you wanting something in targets.cmake that looks like

SET(TARGETS cpu1 cpu3 test123)

Which will then tell cmake to build TGT_CPU1 TGT_CPU3 and TGT_TEST123?

skliper:
_Trac comment by jhageman on 2019-03-04 18:03:43:_

Owner - please provide updates/responses.

skliper:
_Trac comment by jhageman on 2019-07-03 12:58:32:_

Moved unfinished 6.7.0 tickets to next release

skliper:
Ping @excaliburtb - no action in a while, I'll tag as a mission priority.

skliper:
@jphickey thoughts?

excaliburtb:
The actual requirement is to be able to specify multiple targets in targets.cmake

set(TGT1_NAME)
set(TGT2_NAME)

but if I have an environmental or variable-based conditional do:
if(DEFINED foo)
## build TGT1
else()
## build TGT2
endif()

Last I looked, this was an impossible setup because read_targetconfig looks for sequential targets. Also, one might want to build just TGT3 for some reason. Right now, one has to rename the TGT3 configurations. Not impossible but definitely inconvenient.

jphickey:
FWIW, the CMake build system was designed to view the build as a "system" - a set of services/apps that are deployed across one or more CPUs (maybe or maybe not discrete/separate) that collectively implement your spacecraft operation software.   In that view it shouldn't be "different" depending on an environment variable.   

The thought was hardware differences should be abstracted by the PSP, allowing the same set of high level apps to run, with only perhaps different configuration files/tables as you moved to a different host system type.  It wasn't anticipated that there might be a fully different CPU definition.  Or if there was, that one of these two approaches could work:

- it could be built but optionally not deployed/ignored; OR 
[...]

PR #773, opened 2020-07-09
Fix #710, Specify ProcessorID in targets.cmake

### LINKED ISSUE #710: Clarify/clean up spacecraft/cpu names/ids
Defined in platform config:
- CFE_PLATFORM_CPU_ID (referenced by CFE_PLATFORM_TBL_VALID_PRID_1 and CFE_CPU_ID)
  - CFE_PLATFORM_TBL_VALID_PRID_1 is used in internal table logic
  - CFE_CPU_ID looks unused
- CFE_PLATFORM_CPU_NAME (referenced by CFE_CPU_NAME)
  - CFE_CPU_NAME looks unused

Defined in mission config:
- CFE_MISSION_SPACECRAFT_ID (referenced by CFE_PLATFORM_TBL_VALID_SCID_1 and CFE_SPACECRAFT_ID)
  - CFE_SPACECRAFT_ID looks unused
  - CFE_PLATFORM_TBL_VALID_SCID_1 used in internal table logic

Defined by cmake:
- CFE_CPU_ID_VALUE (from TGTID), sets .Default_CpuId in target config
- CFE_CPU_NAME_VALUE (from TGTNAME) sets .Default_CpuName in target config
- CFE_SPACECRAFT_ID_VALUE (from SPACECRAFT_ID) sets .Default_SpacecraftId in target config

Remove/deprecate unused and/or clarify use/intent.  Looks like spacecraft ID may actually be defined differently (42 vs 0x42).

Stay confused.

**Additional context**
Table use looks like it could lead to inconsistencies.

**Requester Info**
Jacob Hageman - NASA/GSFC, triggered by @johnphamngc comments on nasa/psp#154

jphickey:
**Describe the contribution**
Allow explicitly setting of the processor ID in `targets.cmake`

**Testing performed**
Build and sanity-check CFE
Build and run all unit tests.
Confirm that `CFE_PSP_GetProcessorId()` now returns the expected value.

**Expected behavior changes**
The `TGTx_PROCESSOR_ID` setting in `targets.cmake` will be passed to the final build/link of CFE core as the CPU ID.  If unspecified then the CMake index value is used instead (backward compatible).

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Minimal/simple fix that at least should address part of the problem.  Note this value isn't passed to `elf2cfetbl` for table builds at the moment, this only affects the CFE runtime value of processor ID.

At least on pc-linux one can still provide the value on the command line which takes precedence over anything in the build.

### DIFF
--- a/cmake/sample_defs/targets.cmake
+++ b/cmake/sample_defs/targets.cmake
@@ -5,11 +5,15 @@
 # This file indicates the architecture and configuration of the
 # target boards that will run core flight software.
 #
-# The following variables are defined per board, where <x> is the 
-# CPU number starting with 1:
+# The following variables are defined per board, where <x> is a
+# sequential index number starting with 1:
 #
 #  TGT<x>_NAME : the user-friendly name of the cpu.  Should be simple
 #       word with no punctuation.  This MUST be specified.
+#  TGT<x>_PROCESSORID : the default numeric ID for this processor at
+#       runtime.  If not specified, then the sequential index number is
+#       used.  This translates to the numeric value returned by
+#       CFE_PSP_GetProcessorId() at runtime.
 #  TGT<x>_APPLIST : list of applications to build and install on the CPU.
 #       These are built as dynamically-loaded applications and installed
 #       as files in the non-volatile storage of the target, and loaded
--- a/cmake/target/CMakeLists.txt
+++ b/cmake/target/CMakeLists.txt
@@ -48,7 +48,11 @@ include_directories(${CMAKE_BINARY_DIR}/${CFE_CORE_TARGET}/inc)
 
 # The CPU ID and name are defined by the build scripts for this target
 if (DEFINED TGTID)
-  add_definitions(-DCFE_CPU_ID_VALUE=${TGTID})
+  if (DEFINED TGT${TGTID}_PROCESSORID)
+      add_definitions(-DCFE_CPU_ID_VALUE=${TGT${TGTID}_PROCESSORID})
+  else()
+      add_definitions(-DCFE_CPU_ID_VALUE=${TGTID})
+  endif()
 endif()
 if (DEFINED SPACECRAFT_ID)
   add_definitions(-DCFE_SPACECRAFT_ID_VALUE=${SPACECRAFT_ID})


jphickey:
Ping @excaliburtb - this is a very minimalist approach to start with - if you like this we can potentially merge it.

Going forward I'm starting to like the idea of going fully name-based and remove the index numbers from targets.cmake.

something like:
```
set(MISSION_CPUNAMES cpu1 cpu2)

set(cpu1_PROCESSOR_ID 1234)
set(cpu1_APPLIST ...)
set(cpu1_PLATFORM ...)

set(cpu2_PROCESSOR_ID 5678)
set(cpu2_APPLIST ...)
set(cpu2_PLATFORM ...)
```

Obviously this is a bigger change to do, but it would actually clean up a bunch of implementation aspects if we did it this way, and it would probably be more logical from a user perspective too.

excaliburtb:
yes, that works and addresses my immediate problem.

When we were doing our own cmake script mods, we basically took the approach of, if a cpu list of names exists, process the list, otherwise use the index approach

so the read_targetconfig function would look like:

```
  if (NOT DEFINED TGT_CPUIDS)
    set(TGT_CPUIDS)
    while(1)
      math(EXPR TGTID "${TGTID} + 1")
      if (NOT DEFINED TGT${TGTID}_NAME)
        break()
      endif()
      if (NOT DEFINED TGT${TGTID}_SYSTEM)
        set(TGT${TGTID}_SYSTEM "cpu${TGTID}")
        set(TGT${TGTID}_SYSTEM "${TGT${TGTID}_SYSTEM}" PARENT_SCOPE)
      endif()
      if (NOT DEFINED TGT${TGTID}_PLATFORM)
        set(TGT${TGTID}_PLATFORM "default" "${TGT${TGTID}_NAME}")
        set(TGT${TGTID}_PLATFORM "${TGT${TGTID}_PLATFORM}" PARENT_SCOPE)
      endif()
      LIST(APPEND TGT_CPUIDS "${TGTID}")
    endwhile()
  endif()

  foreach(TGTID IN LISTS TGT_CPUIDS)

    if (SIMULATION)
    .. (etc... )
```

something like that would keep it backwards compatible but, as you say, may end up keeping up some of the implementation aspects you were hoping to get rid of without names.

jphickey:
Thanks!  So if this at least is a step in the right direction I've squashed it and marked at as CCB ready.  I can't foresee this breaking anything.

In the meantime I will work on the name-based approach.

jphickey:
Note to reviewers - This is related to #710 and #265, but does not fully address either one of them, so I intentionally did _NOT_ link this PR to those issues because it doesn't address all aspects of either issue, this just addresses the ability to set a processor ID explicitly rather than assuming the sequential number.

excaliburtb:
thanks for the support!

skliper:
@astrogeco Could you merge this into IC?  Priority.

jphickey:
See also follow-on PR in #776 - this takes a stab at name-based config described above - seems to work nicely.

PR #776, opened 2020-07-10
Fix #265, implement cmake name based targets

### LINKED ISSUE #265: Add cmake function to add cpus by function rather than assume consecutive cpu ids.

jphickey:
**Describe the contribution**
Allow targets to be configured in `targets.cmake` by name, rather than a numeric index.  This is more user friendly, allows more configuration flexibility, and cleans up a bunch of extra logic too.

Fixes #265 

**Testing performed**
Build system, sanity check CFE, and run all unit tests

Testing not fully completed - Initially submitting PR for design review/feedback.

**Expected behavior changes**
No change to FSW, but changes the way targets are defined in `targets.cmake`.
This includes backward compatibility for existing files/configs, however.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Also related to #710, this cleans up hardcoded CPU IDs that were in the platform config header file.

### DIFF
--- a/cmake/sample_defs/cpu1_platform_cfg.h
+++ b/cmake/sample_defs/cpu1_platform_cfg.h
@@ -41,16 +41,6 @@
 */
 #include "cfe_mission_cfg.h"
 
-/*
-** CPU Id for target Processor
-*/
-#define CFE_PLATFORM_CPU_ID 1
-
-/*
-** CPU Name for target Processor
-*/
-#define CFE_PLATFORM_CPU_NAME "CPU1"
-
 /**
 **  \cfesbcfg Maximum Number of Unique Message IDs SB Routing Table can hold
 **
@@ -1786,7 +1776,7 @@
 **  \par Limits
 **       This value can be any 32 bit unsigned integer.
 */
-#define CFE_PLATFORM_TBL_VALID_PRID_1            (CFE_PLATFORM_CPU_ID)
+#define CFE_PLATFORM_TBL_VALID_PRID_1            (10)
 #define CFE_PLATFORM_TBL_VALID_PRID_2            (CFE_PLATFORM_TBL_U32FROM4CHARS('a', 'b', 'c', 'd'))
 #define CFE_PLATFORM_TBL_VALID_PRID_3            0
 #define CFE_PLATFORM_TBL_VALID_PRID_4            0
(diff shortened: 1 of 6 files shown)


cmake/sample_defs/cpu1_platform_cfg.h
```diff
@@ -1786,7 +1776,7 @@
 **  \par Limits
 **       This value can be any 32 bit unsigned integer.
 */
-#define CFE_PLATFORM_TBL_VALID_PRID_1            (CFE_PLATFORM_CPU_ID)
+#define CFE_PLATFORM_TBL_VALID_PRID_1            (10)
```
skliper, line 1779: Shouldn't this come from the target configuration?

skliper:
Would it make sense to consider transitioning the platform and mission cfe config file into cmake, similar to osconfig?

cmake/sample_defs/cpu1_platform_cfg.h
jphickey: In this case - no, I don't think so....

This particular item is an optional extra table validation step, and it need not match the actual processor ID of the CFE instance running TBL services.  The code is actually compiled out by default and its been arguably broken for a while, due to the discrepancies outlined in #710 

If the TBL services feature is always supposed to allow the current/active processor ID to be loaded, then it must be coded that way by calling CFE_PSP_GetProcessorId().  We shouldn't make users repeat themselves (DRY principle), and more importantly, the value is overridable at runtime via command line options.  

This list of alternative allowable processor IDs should be used only when you specifically want _different_ (non-local) IDs to be capable of being loaded.

jphickey:
> Would it make sense to consider transitioning the platform and mission cfe config file into cmake, similar to osconfig?

Yes, I think it would!  Particularly for some of the CFE TIME options - these could then be different source files rather than `#ifdef` chunks mixed in everywhere.

But that's a fairly big change to existing paradigms obviously - probably should be a separate steering committee discussion.

skliper:
The msg module as currently drafted has started in this direction with source selection.  Which is clumsy since the rest of cFE uses the header file config.

astrogeco:
CCB-2020-07-15 - Mark as deprecated for 7.0

jphickey:
Update - rebased to current main branch.  Should be good to go.

astrogeco:
[CCB 2020-08-19](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-August-19-2020) Might break CI-Lab, test at the bundle level

PR #784, opened 2020-07-27
fix #759 - deprecates GetLastSenderId()

### LINKED ISSUE #608: CFE_SB_GetLastSenderID returns pointer to internal data, possible race

### LINKED ISSUE #759: remove GetLastSenderId() API and replace with a RecvMsg() API that returns the AppId of the sender of the message being received

CDKnightNASA:
**Describe the contribution**
Fixes #759 - deprecates CFE_SB_GetLastSenderId() API.
Fixes #608

**Testing performed**
make with and without OMIT_DEPRECATED defined.

**System(s) tested on**
Debian 10.3

Christopher.D.Knight@nasa.gov

### DIFF
--- a/cmake/sample_defs/global_build_options.cmake
+++ b/cmake/sample_defs/global_build_options.cmake
@@ -19,7 +19,7 @@
 set(OMIT_DEPRECATED $ENV{OMIT_DEPRECATED} CACHE STRING "Omit deprecated elements")
 if (OMIT_DEPRECATED)
   message (STATUS "OMIT_DEPRECATED=true: Not including deprecated elements in build")
-  add_definitions(-DCFE_OMIT_DEPRECATED_6_7 -DCFE_OMIT_DEPRECATED_6_6 -DOSAL_OMIT_DEPRECATED)
+  add_definitions(-DCFE_OMIT_DEPRECATED_6_8 -DCFE_OMIT_DEPRECATED_6_7 -DCFE_OMIT_DEPRECATED_6_6 -DOSAL_OMIT_DEPRECATED)
 else()
   message (STATUS "OMIT_DEPRECATED=false: Deprecated elements included in build")
 endif (OMIT_DEPRECATED)
--- a/docs/cFE Application Developers Guide.md
+++ b/docs/cFE Application Developers Guide.md
@@ -1769,7 +1769,6 @@ for extracting that field from the header:
 | Total Message Length        | CFE_SB_GetTotalMsgLength                | Command & Telemetry |
 | User Data Message Length    | CFE_SB_GetUserDataLength                | Command & Telemetry |
 | Command Code                | CFE_SB_GetCmdCode                       | Command Only        |
-| Sender ID                   | CFE_SB_GetLastSenderId                  | Command & Telemetry |
 | Checksum                    | CFE_SB_GetChecksum                      | Command Only        |
 
 In addition to the function for reading the checksum field, there is
@@ -1777,15 +1776,6 @@ another API that automatically calculates the checksum for the packet
 and compares it to the checksum in the header. The API is called
 CFE_SB_ValidateChecksum() and it simply returns a success or failure
 indication.
-
-It should be noted that the function, CFE_SB_GetLastSendId, is ideal
-for verifying that critical commands are arriving from a legitimate
-source. This function allows the Developer(s) to define a strict ICD
-between two or more Applications to ensure that an erroneous Application
-does not accidentally issue a critical command. However, its use for
-routine command verification is discouraged since it would increase the
-cross-coupling between Applications and require multiple Applications to
-be modified if a command's source changes.
 
 If the Application's data structure definitions don't include the header
 information, then the CFE_SB_GetUserData API could be used to obtain
--- a/fsw/cfe-core/src/inc/cfe_sb.h
+++ b/fsw/cfe-core/src/inc/cfe_sb.h
@@ -208,7 +208,10 @@ typedef  struct {
 extern CFE_SB_Qos_t CFE_SB_Default_Qos;/**< \brief  Defines a default priority and reliabilty for off-board routing */
 
 
-/** \brief Message Sender Identification Type Definition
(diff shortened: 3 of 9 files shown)


jphickey:
Another note - I believe for other APIs that have been deprecated, rather than clutter up the unit test code with `#ifdef` directives - we can just remove it outright, as (AFAIK) there is no requirement to unit test deprecated functions.  UT code is ugly enough already.

CDKnightNASA:
> Looks OK to me.
> 
> Two minor nitpicks though:
> 
> * The existing `OMIT_DEPRECATED` flags have a CFE prefix.

Whoop, don't know how I missed this!

> * Also if introducing a new deprecation we need to add it to the global_build_options.cmake file so it gets turned on with OMIT_DEPRECATED=true like the others do.

Copy.

I've squashed updates.

CDKnightNASA:
> Another note - I believe for other APIs that have been deprecated, rather than clutter up the unit test code with `#ifdef` directives - we can just remove it outright, as (AFAIK) there is no requirement to unit test deprecated functions. UT code is ugly enough already.

Done, the stubs remain (of course) but are likewise #ifndef'd.

skliper:
Moved the requirement back to "submitted" in requirement tracking tool.  Will need to export and update text requirements.

CDKnightNASA:
> Moved the requirement back to "submitted" in requirement tracking tool. Will need to export and update text requirements.

Would it be worthwhile to export the requirements into an integrated requirements tracking system, like https://pypi.org/project/doorstop/ ?

astrogeco:
CCB 2020-07-29, APPROVED, @skliper to provide doxygen deprecation example.

astrogeco:
@CDKnightNASA can you move the cmake 6_8 definition into it's own commit?

skliper:
Example of deprecated doxygen markings:
https://github.com/nasa/cFE/blob/deeb2942b0766a87d7b399bd8c11c2d5c8deffc0/fsw/cfe-core/src/inc/cfe_fs.h#L194-L214

gerycruz:
> > Moved the requirement back to "submitted" in requirement tracking tool. Will need to export and update text requirements.
> 
> Would it be worthwhile to export the requirements into an integrated requirements tracking system, like https://pypi.org/project/doorstop/ ?

I looked into it and it looks interesting but I'm not sure what it buys us. Have you used it?

skliper:
Would need to trade against the internal tools and process requirement compliance.  We get some things for "free" using standard internal processes, but that doesn't mean it's better.

CDKnightNASA:
> > > Moved the requirement back to "submitted" in requirement tracking tool. Will need to export and update text requirements.
> > 
> > 
> > Would it be worthwhile to export the requirements into an integrated requirements tracking system, like https://pypi.org/project/doorstop/ ?
> 
> I looked into it and it looks interesting but I'm not sure what it buys us. Have you used it?

I have not, but some folks at the 2019 FSW mentioned they used it. Certainly would need to do a study of it.

CDKnightNASA:
> @CDKnightNASA can you move the cmake 6_8 definition into it's own commit?

done

CDKnightNASA:
> CCB 2020-07-29, APPROVED, @skliper to provide doxygen deprecation example.

done, ready for review again

astrogeco:
@skliper this looks good to me, are we good to merge?

skliper:
Can this go in the 6.8 dev cycle?  It's wrapped with the 6.8 omit define.  Other than that, yes looks good to me.

astrogeco:
> Recommend merge to 6.8 development

If you you mean make this part of 6.8 then yes!

skliper:
No, I mean development on top of 6.8.  Otherwise the OMIT numbering breaks convention.  I'm requesting making the rc tags BEFORE merging this change.

skliper:
Marked my preference as targeting this for 7.0 release.  Sorry, my quick comments weren't very clear.

astrogeco:
The milestone helps, thanks!

astrogeco:
What's the convention then? Should the deprecate tag then say deprecate 7_0 or 7_1?

skliper:
> What's the convention then? Should the deprecate tag then say deprecate 7_0 or 7_1?

6_8.

astrogeco:
> > What's the convention then? Should the deprecate tag then say deprecate 7_0 or 7_1?
> 
> 6_8.

For completeness then, does DEPRECATE_6_8 means that 6.8 was the last version which supports this function?

skliper:
We've typically just marked with the last released version, so it could be interpreted that way but I'm not aware of any hard rules or published deprecation process.  See nasa/cFS#67.

jphickey:
> does DEPRECATE_6_8 means that 6.8 was the last version which supports this function?

Yes, that is the paradigm we've been using up to this point -- the number reflects the last release were it was _not_ deprecated.  The idea of putting a version number in the tag was so we'd know which ones had "aged out" in a future release and should be taken out completely.  

However as I mentioned before now that we have a codename for API compatibility then maybe we should consider using that instead.

astrogeco:
@CDKnightNASA does this actually fix #608? Just wondering since you don't mention it in the commits

CDKnightNASA:
> @CDKnightNASA does this actually fix #608? Just wondering since you don't mention it in the commits

Yes it fixes #608.

PR #840, opened 2020-08-25
HOTFIX - no longer add unit tests from within unit tests in msg UT

skliper:
**Describe the contribution**
HOTFIX - unit tests added from within unit tests will not execute, replaced this pattern with direct calls to the main subtest setup routine.

**Testing performed**
Build unit tests and ran, all tests (including subtests) ran.

**Expected behavior changes**
All tests run

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: integration candidate bundle + this commit

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/modules/msg/unit-test-coverage/msg_UT.c
+++ b/modules/msg/unit-test-coverage/msg_UT.c
@@ -45,9 +45,9 @@ void UtTest_Setup(void)
     UT_Text("Message header coverage test...");
 
     UT_ADD_TEST(Test_MSG_Init);
-    UT_ADD_TEST(Test_MSG_CCSDSPri);
-    UT_ADD_TEST(Test_MSG_CCSDSExt);
-    UT_ADD_TEST(Test_MSG_MsgId_Shared);
+    Test_MSG_CCSDSPri();
+    Test_MSG_CCSDSExt();
+    Test_MSG_MsgId_Shared();
     UT_ADD_TEST(Test_MSG_MsgId);
     UT_ADD_TEST(Test_MSG_Checksum);
     UT_ADD_TEST(Test_MSG_FcnCode);


astrogeco:
Does this mean that we need to revisit this behavior in `UT_ADD_TEST()`?

skliper:
I just closed the issue in OSAL as "wontfix" with the recommendation to avoid the pattern of adding tests from within other tests.  I'd think this is enough (just don't do it).  It is a result of developing the related tests on an older OSAL, then being surprised they no longer run.  Could update things in ut_assert such that this could work (or at least reject the request), but not sure if it's worth the effort.

astrogeco:
Can you link that osal issue?

astrogeco:
> I just closed the issue in OSAL as "wontfix" with the recommendation to avoid the pattern of adding tests from within other tests. I'd think this is enough (just don't do it). It is a result of developing the related tests on an older OSAL, then being surprised they no longer run. Could update things in ut_assert such that this could work (or at least reject the request), but not sure if it's worth the effort.

Just looking at the code we might want to add some comments for why some of these don't follow the semantic pattern. Since I can't tell from a quick glance why some of these use `UT_ADD_TEST` and others don't

```
    UT_ADD_TEST(Test_MSG_Init);
    UT_ADD_TEST(Test_MSG_CCSDSPri);
    Test_MSG_CCSDSPri();
    UT_ADD_TEST(Test_MSG_CCSDSExt);
    Test_MSG_CCSDSExt();
    UT_ADD_TEST(Test_MSG_MsgId_Shared);
    Test_MSG_MsgId_Shared();
    UT_ADD_TEST(Test_MSG_MsgId);
    UT_ADD_TEST(Test_MSG_MsgId);
```

skliper:
Related to nasa/osal#577.  Individual tests are added w/ the UT_ADD_TEST call.  Some tests are grouped at a lower level (Test_MSG_CCSDSExt is a group of tests), and those functions are called directly.  Same pattern is used in `sb`:

https://github.com/nasa/cFE/blob/a148b97511de3f45f6677c1a3a0b17197f867e28/fsw/cfe-core/unit-test/sb_UT.c#L134-L135

astrogeco:
> I just closed the issue in OSAL as "wontfix" with the recommendation to avoid the pattern of adding tests from within other tests. I'd think this is enough (just don't do it). It is a result of developing the related tests on an older OSAL, then being surprised they no longer run. Could update things in ut_assert such that this could work (or at least reject the request), but not sure if it's worth the effort.

I opened nasa/osal#578 to have a targeted discussion about it.

PR #857, opened 2020-09-02
Fix #823, avoid infinite loop in CDS registry find

### LINKED ISSUE #823: CFE_ES_FindCDSInRegistry() infinte loop if registry size is zero
**Describe the bug**
The loop inside this function has a poorly-constructed condition for ending the loop.  It is comparing a `uint32` type to an `int32` type, and in the event that the `MaxNumRegEntries` is zero, it becomes impossible for this condition to be true and the loop runs forever.

**To Reproduce**
Easy to see in unit test if one clears the `CFE_ES_Global` data structure between tests.

**Expected behavior**
Loops should never have ending conditions that are impossible to reach unless they are supposed to be infinite.  In the event that `CFE_ES_Global.CDSVars.MaxNumRegEntries` is 0, it should exit immediately.

**Code snips**
https://github.com/nasa/cFE/blob/a148b97511de3f45f6677c1a3a0b17197f867e28/fsw/cfe-core/src/es/cfe_es_cds.c#L657

It is generally a bad idea to do any sort of relational comparison (greater than/less than) between signed and unsigned types, C++ actually errors about this but C does not.

**System observed on:**
Ubuntu 20.04

**Additional context**
This variable is initialized in FSW from the config `CFE_PLATFORM_ES_CDS_MAX_NUM_ENTRIES` which does say that the value needs to be at least 8.  But during unit test the value can be zero.

Interestingly, `CFE_ES_FindFreeCDSRegistryEntry` swaps the weirdly-structured do-while for a normal while loop, so it is OK however it still does a signed/unsigned compare which should be fixed.

**Reporter Info**
Joseph Hickey, Vantage Systems. Inc.

jphickey:
**Describe the contribution**
The CFE_ES_FindCDSInRegistry function had an unusual loop control structure with mixed types of signed and unsigned.  This has the possibility of being infinite if the MaxNumRegEntries is zero due to the way the end condition is structured.  Simplify to be like other loops and use unsigned int control variable.

Fixes #823 

**Testing performed**
Build and sanity check CFE
Run all unit tests

**Expected behavior changes**
No impact to behavior so long as CFE was configured where the number of CDS entries is not zero.
If configured to be zero, this avoids an infinite loop.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_cds.c
+++ b/fsw/cfe-core/src/es/cfe_es_cds.c
@@ -637,13 +637,10 @@ int32 CFE_ES_UnlockCDSRegistry(void)
 int32 CFE_ES_FindCDSInRegistry(const char *CDSName)
 {
     int32 RegIndx = CFE_ES_CDS_NOT_FOUND;
-    int32 i = -1;
+    uint32 i = 0;
 
-    do
+    while ( (RegIndx == CFE_ES_CDS_NOT_FOUND) && (i < CFE_ES_Global.CDSVars.MaxNumRegEntries) )
     {
-        /* Point to next record in the CDS Registry */
-        i++;
-
         /* Check to see if the record is currently being used */
         if (CFE_ES_Global.CDSVars.Registry[i].Taken == true)
         {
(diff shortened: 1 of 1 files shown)


fsw/cfe-core/src/es/cfe_es_cds.c
```diff
@@ -654,7 +651,11 @@ int32 CFE_ES_FindCDSInRegistry(const char *CDSName)
                 RegIndx = i;
             }
         }
-    } while ( (RegIndx == CFE_ES_CDS_NOT_FOUND) && (i < (CFE_ES_Global.CDSVars.MaxNumRegEntries-1)) );
+
+        /* Point to next record in the CDS Registry */
+        i++;
```
astrogeco, line 656: Would it make more sense to do a for loop with a break?

astrogeco:
[CCB 2020-09-02](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-02-2020) APPROVED

PR #859, opened 2020-09-02
Fix #797, refactor internal table/id management

### LINKED ISSUE #797: Improve resource management and internal consistency in ES
Executive Services (ES) maintains many internal tables of resources/objects, which track applications, libraries, tasks, counters, and memory pools, etc.

There is a lot of inconsistency in how these internal objects are managed/tracked.  Some have a `RecordUsed` boolean that is set true/false depending on whether the record is used.  App table uses the `AppState` member.  The memory pool passes around direct pointers which are dereferenced (potentially dangerous). 

Furthermore, all "ID" values issued to external apps are zero based, and therefore can easily alias other object types or even other objects of the same type.  For instance, if one app had ID "5" and it was deleted, and after this a new/different app was started, the new app might also be assigned ID "5" ... this means any old/stale reference to AppID 5 will now be referring to the wrong app.

Define a properly abstract "resource ID" type and use it consistently across all these various internal tables.  The abstraction should be based on/compatible with what OSAL does for its internal records.

- The "id" value also serves as a marker to indicate whether the respective table entry is in use or not.
- Zero is reserved as an invalid value, and marks entries which are NOT in use.  (e.g. so a `memset()` to all zero can consistently clear an entry).  Valid entries/ids are never zero.
- Valid values are split into a "type" and sequential "index" value
- Type is unique for apps. libs, counters, etc so these cannot get crossed/misinterpreted (i.e. can't pass an appID in place of a libID or vice versa).
- Index is sequential and does not immediately repeat (i.e. don't wrap until 0xFFFF, do not recycle/reassign IDs after deletion).
- Provide a consistent mechanism to convert ID to a zero based index where an array/table is needed.

**Additional context**
This internal cleanup is a prerequisite to several related tickets:

- #28 - This blurs the difference between Libraries and Applications and makes the App API also apply to libraries, so they need a consistent means of identification and (possibly) make a single unified table.
- #651 - Need a better way to identify mem pools in CMD/TLM messages, not a direct address/pointer.
- #173 - More examples of ugliness/duplication of logic between apps and libs

**Requester Info**

jphickey:
**Describe the contribution**
Introduce wrapper/accessor functions to look up table entries by ID for ES & EVS subsystems.

__Do not use AppID as a table index__.

Note - This does not change existing external APIs and AppIDs are still zero-based uint32.  This only changes the internal
structures to remove use of ID as an array index, and to use a lookup function to locate the table entry from an ID.  All entry access is then performed via the table entry pointer, rather than as an array index.

This provides the groundwork for abstract IDs without actually changing anything fundamental about resource IDs.

Fixes #797 

**Testing performed**
Build and sanity check CFE - start up apps and send commands, confirm normal operation
Run all unit tests.

**Expected behavior changes**
No impact to behavior - internal refactoring only.
API additions to support more abstract resource IDs.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_api.c
+++ b/fsw/cfe-core/src/es/cfe_es_api.c
@@ -170,8 +170,10 @@ int32 CFE_ES_ResetCFE(uint32 ResetType)
 int32 CFE_ES_RestartApp(uint32 AppID)
 {
     int32 ReturnCode = CFE_SUCCESS;
+    CFE_ES_AppRecord_t *AppRecPtr;
 
-    if ( AppID < CFE_PLATFORM_ES_MAX_APPLICATIONS )
+    AppRecPtr = CFE_ES_LocateAppRecordByID(AppID);
+    if ( AppRecPtr != NULL )
     {
 
        CFE_ES_LockSharedData(__func__,__LINE__);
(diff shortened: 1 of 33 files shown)


astrogeco:
Would it be easier to review this by splitting it up into a couple of separate commits?

jphickey:
> Would it be easier to review this by splitting it up into a couple of separate commits?

I could probably split into a separate commit for each subsystem (ES, EVS, SB, etc) if that helps?

astrogeco:
That might be helpful, I like atomized commits as much as possible. The subsystem approach makes sense

jphickey:
OK - separate commits for each subsystem. ES is still the biggest one as its where the bulk of changes lie.  

Note that most changes are mechanical -- that is, changing something like `CFE_ES_Global.AppTable[AppID].`_something_ into `AppRecPtr->`_something_ instead.  The idea here is that all this internal stuff needs to be cleaned up first which shouldn't change anything other than make it consistently apply the lookup function to get/change internal data.

Externally the API uses the abstract AppID (`uint32`) but internally (after validation of said AppID) we can pass around a pointer to the entry instead, avoiding repeat lookups and clean up the code substantially.

astrogeco:
@jphickey should we convert this to draft while working on the rename of `CFE_ES_GetTaskID`?

jphickey:
In regards to the original concern about differentiating OSAL tasks from ES tasks -- the more I think about this, I don't think we really need to do this via a different name, because these two should really never be mixed.  The fact that they are mixed now is actually the problem that this PR is fixing.

Moving forward -- _all_ tasks at the CFE level and above should be ES tasks, and _all_ management of those tasks should happen though ES APIs.  There really shouldn't be any areas where OSAL task IDs are intermixed with ES task IDs.  They should all be ES task IDs - EXCEPT, of course, in OSAL itself, and the ES APIs that actually call OSAL.

skliper:
Copy, I think if it's all ES tasks then it's fine.

PR #852, opened 2020-08-31
Fix #851, Pass aligned message into CFE_MSG_ComputeCheckSum

skliper:
**Describe the contribution**
Fix #851 - Fixes the cast-align error (use the aligned Msg since it's available already)

**Testing performed**
Standard build and unit test, passes

**Expected behavior changes**
No more alignment error

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: bundle main + this commit

**Additional context**
None

**Third party code**
None

**Contributor Info - All information REQUIRED for consideration of pull request**
Jacob Hageman - NASA/GSFC

PR #817, opened 2020-08-18
Fix #815, remove all deprecated code

jphickey:
**Describe the contribution**
In the next major CFE release, this code will be no longer supported at all.  It should be removed early in the cycle
to avoid needing to maintain this compatibility code.

Fixes #815 

**Testing performed**
Build and run, sanity check CFE.
Run all unit tests.

**Expected behavior changes**
Removes code previously deprecated.
Identical to previous build if the `OMIT_DEPRECATED=true` flag was set during configuration.

**System(s) tested on**
Ubuntu 20.04
RTEMS 4.11.3

**Additional context**
Also depends on nasa/to_lab#55 - should be merged in same cycle.

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #861, opened 2020-09-03
Integration Candidate: 2020-09-02

### LINKED ISSUE #815: Remove CFE deprecated code for next release
Now that the "6.8.x" release branch has been created, the next release should remove all code/functions that are currently marked as deprecated.

Removal of all code currently contained within an `OMIT_DEPRECATED` conditional compile switch.

**Additional context**
The removal only applies to "main" branch - not the 6.7.x release branch, where it will continue to exist for that release series.  This assumes the next release (main) will become version 7.0.

There are other changes currently in development which do break some deprecated items - hence why I'd like to remove this sooner rather than later, as it saves the work of updating/fixing this old code if the intent is to remove it anyway.

**Requester Info**

### LINKED ISSUE #823: CFE_ES_FindCDSInRegistry() infinte loop if registry size is zero

### LINKED ISSUE #851: RPi4 make process throws alignment error
Following the "quick start" section of the README, generates an error and make failure:
```
Scanning dependencies of target msg
[ 83%] Building C object msg/CMakeFiles/msg.dir/src/cfe_msg_ccsdspri.c.o
[ 83%] Building C object msg/CMakeFiles/msg.dir/src/cfe_msg_init.c.o
[ 84%] Building C object msg/CMakeFiles/msg.dir/src/cfe_msg_msgid_shared.c.o
[ 85%] Building C object msg/CMakeFiles/msg.dir/src/cfe_msg_sechdr_checksum.c.o
/home/pi/src/NASA/cFS/cfe/modules/msg/src/cfe_msg_sechdr_checksum.c: In function ‘CFE_MSG_GenerateChecksum’:
/home/pi/src/NASA/cFS/cfe/modules/msg/src/cfe_msg_sechdr_checksum.c:82:49: error: cast increases required alignment of target type [-Werror=cast-align]
     cmd->Sec.Checksum = CFE_MSG_ComputeCheckSum((CFE_MSG_Message_t *)cmd);
                                                 ^
cc1: all warnings being treated as errors
make[7]: *** [msg/CMakeFiles/msg.dir/build.make:102: msg/CMakeFiles/msg.dir/src/cfe_msg_sechdr_checksum.c.o] Error 1
make[6]: *** [CMakeFiles/Makefile2:613: msg/CMakeFiles/msg.dir/all] Error 2
make[5]: *** [Makefile:141: all] Error 2
make[4]: *** [CMakeFiles/native_default_cpu1-all.dir/build.make:57: CMakeFiles/native_default_cpu1-all] Error 2
make[3]: *** [CMakeFiles/Makefile2:298: CMakeFiles/native_default_cpu1-all.dir/all] Error 2
make[2]: *** [CMakeFiles/Makefile2:177: CMakeFiles/mission-all.dir/rule] Error 2
make[1]: *** [Makefile:214: mission-all] Error 2
make: *** [Makefile:120: all] Error 2
pi@renegade:~/src/NASA/cFS $ 
```
At first I thought this was because I was running on the 64-bit Beta Raspberry Pi OS, so I tried it on a different Pi4, this one running the stable 32-bit distribution (Buster, latest release (8/20), updated yesterday (8/30)). It fails on the 32-bit version also.

yammajamma:
**Describe the contribution**
Fix #815 
Fix #823 
Fix #851 

**Testing performed**
Bundle CI - https://github.com/nasa/cFS/pull/136/checks

**Expected behavior changes**
PR #817 - In the next major CFE release, this code will be no longer supported at all. It should be removed early in the cycle
to avoid needing to maintain this compatibility code.

PR #857 - The `CFE_ES_FindCDSInRegistry` function had an unusual loop control structure with mixed types of signed and unsigned. This has the possibility of being infinite if the `MaxNumRegEntries` is zero due to the way the end condition is structured. Simplify to be like other loops and use unsigned int control variable.

PR #852 -  Fixes the cast-align error (use the aligned Msg since it's available already).

HOTFIX-20200902 - Fix sb unit test setup issue.

HOTFIX 20200902 - Update documentation links for deprecated symbols. 

HOTFIX 20200902 - Fix SB Test_CleanupApp_API AppID.

**System(s) tested on**
Ubuntu - CI

**Additional context**
https://github.com/nasa/cFS/pull/136

Jacob Hageman, NASA-GSFC


jphickey:
FYI - @skliper and @yammajamma I have a fix for the documentation symbol name changes that gets rid of the doxygen warnings.  Should I push as hotfix directly to integration-candidate?

yammajamma:
@jphickey that should work

jphickey:
See commit d47a869 for the fixes.  Most changing old names to new (non-deprecated) names within documentation, but there were some items e.g. shell related event ID that the entire feature was deprecated.

astrogeco:
@skliper and @jphickey we still have a unit test failure, see https://github.com/nasa/cFS/pull/136/checks: 

```
8/76 Test #68: cfe-core_sb_UT ....................***Failed    0.02 sec
```

skliper:
Started digging in, but seems like a behavior change... still need to nail down the difference.

jphickey:
This is the same as it was last week ... in the `Test_CleanupApp_API()` test we are getting 4 events instead of 3.

The previous hotfix in commit 3baef94 changed it so it is now a "real" failure rather than a test teardown (TTF) failure which makes the count actually show up in the summary.  However the failure is due to the event count difference and it is still a failure....

skliper:
> This is the same as it was last week...

Signature is the same but cause is different.

The previous problem is there were four events counted because it wasn't clearing the number of events between running the last subtest of Test_RcvMsg_API and running Test_CleanupApp_API.  Now there really are 4 events being sent as part of Test_CleanupApp_API.  Behavior change.

jphickey:
I am taking a closer look ... I do see that that `CFE_SB_DeletePipeWithAppId` is being invoked several times, looks like a possible case of stale data in the table from a previous test case making the entry look "valid" ... so it is trying to delete but ultimately triggering event 46 (`CFE_SB_DEL_PIPE_ERR1_EID`)

jphickey:
FWIW, this is happening on the first pass (line 3563 - which is commented to be the "bad" application ID - and isn't supposed to actually do anything AFAICT).  

This is sort of confirming that there are remnants in the table which appear valid....

jphickey:
I think I found it ... AppID "0" was both the valid and invalid AppID here in this test ... 

Adding a call to `UT_SetAppID(1);` prior to this test case seems to fix it and turns the first call back into a no-op (which agrees with the comments).

In a near term change the AppID "0" will become always invalid so this is really just a patch to get things working for this IC.

skliper:
I did check the code at 8b9a3db at the point where my hotfix was merged and it did reject the first call and delete on the 2nd... seems like there was some other logic change in the other commits that affected behavior?  Not saying the current logic isn't correct, just surprised to see a change in behavior of a unit test.

jphickey:
Possibly related to the internal table management changes in #797 / #859 ....
This updated the code to use the output of CFE_ES_GetAppID() as the "real" AppID but this is returning 0 at this stage.  The code seems to be expecting AppID 1.

skliper:
I just found:

jphickey:
Right ... it is the `UT_AppID` here which overrides that:

https://github.com/nasa/cFE/blob/08f6eab17702eac689a468974a32af8d41d4362a/fsw/cfe-core/unit-test/ut_support.c#L173

Once I submit PRs for the rest of the AppID cleanup stuff I think this can actually go away for good.

skliper:
I suspect this should be != AppID to get branch coverage.  You need an entry in use but != AppID:

https://github.com/nasa/cFE/blob/68f871afcc0f053ffba156e63cbd5b79cf746937/fsw/cfe-core/unit-test/sb_UT.c#L3557

jphickey:
Just as a sanity check I locally added another assert:

    EVTSENT(CFE_SB_DEL_PIPE_ERR1_EID);

And this is also passing -- so it appears that we are getting the error case.  Probably worth adding this to ensure we cover that event path (but that doesn't need to be a hotfix).

astrogeco:
@yammajamma can you update the "Expected behavior changes" above with a summary of the hotfixes?

PR #865, opened 2020-09-04
fix #862 - unsub of a message ID that is already unsubbed

### LINKED ISSUE #862: If a message is subscribed, then unsubscribed, additional unsubscribes do not raise error events
**Describe the bug**
If a message is subscribed, then unsubscribed, additional unsubscribes do not raise error events 

**To Reproduce**
Have SB subscribe to a message
Have SB unsubscribe to that message
Have SB unsubscribe again to that message

**Expected behavior**
Event message should be raised

**Code snips**
https://github.com/nasa/cFE/blob/08f6eab17702eac689a468974a32af8d41d4362a/fsw/cfe-core/src/sb/cfe_sb_api.c#L1059-L1112

If a message was previously subscribed to, and all pipes subscribing to it were subsequently unsubscribed, and an additional unsubscribe is issued, a valid routing index would still exist in the msgmap table, and thus an error event would not be raised.

**System observed on:**
N/A, discovered via code inspection

**Additional context**
N/A

**Reporter Info**
John N Pham, Northrop Grumman

CDKnightNASA:
Closes #862 

**Describe the contribution**
After the changes implemented in #101, there may be routing table entries with no subscribers (RoutePtr->ListHeadPtr would be NULL.) This could cause a seg-fault. Also, even if there are entries in the routing table, there will be no event generated if the unsubscribe does not find a matching route entry.

**Testing performed**
Ran unit test (with updated event count.)

**Expected behavior changes**
Repeated unsubscriptions should function fine, generating an informational event if there is not a current subscription.

**System(s) tested on**
Debian 10.5

Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -1011,7 +1011,6 @@ int32 CFE_SB_UnsubscribeFull(CFE_SB_MsgId_t MsgId,CFE_SB_PipeId_t PipeId,
     CFE_SB_RouteEntry_t* RoutePtr;
     uint32  PipeIdx;
     uint32  TskId = 0;
-    bool    MatchFound = false;
     CFE_SB_DestinationD_t   *DestPtr = NULL;
     char    FullName[(OS_MAX_API_NAME * 2)];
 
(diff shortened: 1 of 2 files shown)


skliper:
I appreciate that in the last couple of weeks we've seen three different loop structures to accomplish the same logic.  While loop with an incrementer, for loop with a break, and now a for loop with dual exit condition.  Way to keep things interesting!

astrogeco:
> I appreciate that in the last couple of weeks we've seen three different loop structures to accomplish the same logic. While loop with an incrementer, for loop with a break, and now a for loop with dual exit condition. Way to keep things interesting!

I don't know if you're genuinely excited or not-so-subtly indicating that we should choose one way and stick to it as much as possible :)

CDKnightNASA:
I feel like we keep going around and around on this...

(Sorry, I just couldn't resist!)

yammajamma:
@CDKnightNASA is this ready for a CCB review to get more inputs?

CDKnightNASA:
> @CDKnightNASA is this ready for a CCB review to get more inputs?

yes thx

astrogeco:
[CCB 2020-08-09](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-09-2020)  APPROVED, How do we ensure we test and catch this? Open a new issue for creating a test.

skliper:
I opened #873 for the added test case.

skliper:
Added the test case in #875

skliper (approved):
As long as it passes test w/ #875 I'm good with this.

Issue #249, opened 2019-09-30
Please provide va_list variants of variadic functions


skliper:
I've recently encountered a need to conditionally call `CFE_EVS_SendEvent` or `CFE_ES_WriteToSysLog` depending on the app's status. Unfortunately, both of these functions are variadic functions, and they can't easily be wrapped in another variadic function. 

The C standard library provides variants of its variadic functions that take a `va_list`. For example, `printf` has the `vprintf` variant.

Please provide similar variants for cFE's variadic functions. The implementation would be simple: the majority of the function's logic can reside in the `va_list` version, and the variadic function can become thin a wrapper around the `va_list` version.

skliper:
_Imported from trac issue 218.  Created by tflemin1 on 2017-11-03T11:25:30, last modified: 2019-03-01T15:27:58_

skliper:
_Trac comment by tflemin1 on 2017-11-15 12:06:45:_

See also [cfs_osal:193 OSAL issue 193].

skliper:
_Trac comment by tflemin1 on 2018-01-04 14:02:19:_

Initial priority value was erroneous.

PR #867, opened 2020-09-08
Fix #788, Simplified CFE_EVS_SendEvent macros

### LINKED ISSUE #788: CFE_EVS_Send{Crit|Critical|Err|Error|Info|Information|Debug} wrapper for CFE_EVS_SendEvent
With the expansion of the `CFE_EVS_EventType_*` constants as part of `CFE_EVS_SendEvent()`, it got me thinking that we have a lot of extraneous typing when there's only four types of events. What about adding wrapper macros/functions that have the "EventType" as part of the function name? I also recommend both long and short versions, such as `CFE_EVS_SendCrit()` and `CFE_EVS_SendCritical()`.

So, for example:

```
CFE_EVS_SendEvent(MYAPP_EVENT_EID, CFE_EVS_EventType_ERROR, "error message (%d)", status);
```
becomes
```
CFE_EVS_SendErr(MYAPP_EVENT_EID, "error message (%d)", status);
```

**Requester Info**
Christopher.D.Knight@nasa.gov

CDKnightNASA:
Closes #788 

**Describe the contribution**
Macros for more compact calls to CFE_EVS_SendEvent, making the type be part of the fn name.

Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_evs.h
+++ b/fsw/cfe-core/src/inc/cfe_evs.h
@@ -59,6 +59,14 @@
 #define OS_PRINTF(m,n) 
 #endif
 
+/*
+** Utility macros to make for simpler/more compact/readable code.
+*/
+#define CFE_EVS_SendDbg(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_DEBUG, __VA_ARGS__)
+#define CFE_EVS_SendInfo(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_INFORMATION, __VA_ARGS__)
+#define CFE_EVS_SendErr(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_ERROR, __VA_ARGS__)
+#define CFE_EVS_SendCrit(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_CRITICAL, __VA_ARGS__)
+
 /** \name Common Event Filter Mask Values  */
 /** \{ */
 #define  CFE_EVS_NO_FILTER        0x0000       /**< \brief Stops any filtering.  All messages are sent. */
--- a/fsw/cfe-core/src/inc/cfe_evs.h
+++ b/fsw/cfe-core/src/inc/cfe_evs.h
@@ -62,10 +62,11 @@
 /*
 ** Utility macros to make for simpler/more compact/readable code.
 */
-#define CFE_EVS_SendDbg(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_DEBUG, __VA_ARGS__)
-#define CFE_EVS_SendInfo(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_INFORMATION, __VA_ARGS__)
-#define CFE_EVS_SendErr(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_ERROR, __VA_ARGS__)
-#define CFE_EVS_SendCrit(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_CRITICAL, __VA_ARGS__)
+#define CFE_EVS_Send(E,T,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_##T, __VA_ARGS__)
+#define CFE_EVS_SendDbg(E,...) CFE_EVS_Send(E, DEBUG, __VA_ARGS__)
+#define CFE_EVS_SendInfo(E,...) CFE_EVS_Send(E, INFORMATION, __VA_ARGS__)
+#define CFE_EVS_SendErr(E,...) CFE_EVS_Send(E, ERROR, __VA_ARGS__)
+#define CFE_EVS_SendCrit(E,...) CFE_EVS_Send(E, CRITICAL, __VA_ARGS__)
 
 /** \name Common Event Filter Mask Values  */
 /** \{ */


CDKnightNASA:
Should I, as part of this PR, change a bunch of cFE code to use these simplified macro calls?

astrogeco:
> Should I, as part of this PR, change a bunch of cFE code to use these simplified macro calls?

I like keeping things separate. So at the very least do it in separate commits. Another option is to provide a sample implementation or documentation with this PR. And, once the approach is "approved" by the CCB open a PR to update the cFE code.

iamthad:
This could be done with an inline function instead of a macro if a version of `CFE_EVS_SendEvent` that takes `va_list` (as requested in  #249) were implemented.

klystron78:
Suggestion for less duplicate code:

#define CFE_EVS_SendDbg(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_DEBUG, __VA_ARGS__)
#define CFE_EVS_SendInfo(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_INFORMATION, __VA_ARGS__)
#define CFE_EVS_SendErr(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_ERROR, __VA_ARGS__)
#define CFE_EVS_SendCrit(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_CRITICAL, __VA_ARGS__)

instead:

#define CFE_EVS_Send(E,T,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_##T, __VA_ARGS__)
#define CFE_EVS_SendInfo(E,...) CFE_EVS_Send(E, INFORMATION, __VA_ARGS__)
... and so on

skliper:
> Suggestion for less...

@klystron78 - copy, a helper macro for the helper macros.  I'm fine with either approach.

klystron78:
Looks like double underscore bolds in the comments. There's probably a way to do a code block or something like that.

skliper:
@klystron single back quote (just learned also known as birk, blugle, or grave) `inline` code or triple
```
for block of code
```

CDKnightNASA:
> This could be done with an inline function instead of a macro if a version of `CFE_EVS_SendEvent` that takes `va_list` (as requested in #249) were implemented.

I'm confused what the benefit is of an inline (which, of course, isn't always inline) over a macro. Can you describe/prototype?

CDKnightNASA:
> Suggestion for less duplicate code:
> 
> ```
> #define CFE_EVS_SendDbg(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_DEBUG, __VA_ARGS__)
> #define CFE_EVS_SendInfo(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_INFORMATION, __VA_ARGS__)
> #define CFE_EVS_SendErr(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_ERROR, __VA_ARGS__)
> #define CFE_EVS_SendCrit(E,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_CRITICAL, __VA_ARGS__)
> ```
> 
> instead:
> 
> ```
> #define CFE_EVS_Send(E,T,...) CFE_EVS_SendEvent((E), CFE_EVS_EventType_##T, __VA_ARGS__)
> #define CFE_EVS_SendInfo(E,...) CFE_EVS_Send(E, INFORMATION, __VA_ARGS__)
> ```
> 
> ... and so on

I like, I'll change, thanks!

iamthad:
> > This could be done with an inline function instead of a macro if a version of `CFE_EVS_SendEvent` that takes `va_list` (as requested in #249) were implemented.
> 
> I'm confused what the benefit is of an inline (which, of course, isn't always inline) over a macro. Can you describe/prototype?

General advice these days seems to be to prefer inline functions over function-like macros. For example, see [SEI CERT Recommendation PRE00-C](https://wiki.sei.cmu.edu/confluence/display/c/PRE00-C.+Prefer+inline+or+static+functions+to+function-like+macros). I admit, for this sort of trivial wrapper, the benefit is probably small and probably outweighed by the additional lines of code.

It would look something like the following (using a hypothetical `CFE_EVS_VSendEvent`):

```c
static inline int32 CFE_EVS_SendDbg(uint16 EventID, const char *Spec, ...)
{
    va_list ap;
    va_start(ap, Spec);
    int32 res = CFE_EVS_VSendEvent(EventID, CFE_EVS_EventType_DEBUG, Spec, ap);
    va_end(ap);
    return res;
}
```

jphickey:
I'm generally a fan of inline functions rather than macros because they are more type-safe, and they offer more predictable/defined behavior as to when things are evaluated and when side effects occur, such as the classic example of a macro that evaluates its arguments more than once.

Admittedly they aren't quite as useful in C as they are in C++, where you have templates.  So in C we need a explicitly-defined inline function for every type variant, whereas a macro or C++ template can be evaluated at compile-time for any type.

It is true they are not always actually inlined - it's a suggestion to the compiler, but depends on compiler capabilities, settings, etc as to whether it accepts that suggestion.

But I'm not sure if any compiler can inline something that uses `va_list` because AFAIK it depends on having those arguments in a stack frame.  If truly inline-d then there is no stack frame for it.  For anything that uses variable arguments/`va_list` then it is probably not worth trying to inline it.

astrogeco:
@CDKnightNASA is this ready to merge?

CDKnightNASA:
> @CDKnightNASA is this ready to merge?

yes please

skliper:
@astrogeco - what's the status on this one?

astrogeco:
> @astrogeco - what's the status on this one?

Whoops, looks like this never made it to a CCB since it hasn't been "github reviewed". Tagging it for next week

astrogeco:
@skliper or we could fast-track it

astrogeco:
@CDKnightNASA can you rebase/squash the commits?

Issue #616, opened 2020-04-15
Duplicate CFE error codes


jphickey:
**Is your feature request related to a problem? Please describe.**
While coding fixes for another ticket, happen to stumble upon two CFE ES error codes with the same value.

https://github.com/nasa/cFE/blob/60a5f6580d1e737f46efce0418aba5b8dac6df6a/fsw/cfe-core/src/inc/cfe_error.h#L617-L634

**Describe the solution you'd like**
Should have unique/distinguishable values

**Requester Info**
Joseph Hickey, Vantage Systems, Inc.

skliper:
Duplicate of #846

Issue #483, opened 2020-01-21
Deprecate multiple "success" code responses


skliper:
**Is your feature request related to a problem? Please describe.**
API's with multiple "success" codes are frequently mishandled

**Describe the solution you'd like**
Single success response, unique information should be passed back in parameters

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
Some discussion of this topic in the context of #846 with respect to `CFE_ES_ERR_SYS_LOG_TRUNCATED` as (possibly) returned by `CFE_ES_WriteToSysLog`

My take on the topic is that all API functions should strive to meet the "ACID" test that is often employed in database software, described here: 

https://en.wikipedia.org/wiki/ACID

In particular the atomicity principle (A) - A function either succeeds completely or fails completely.

Applied to something like `CFE_ES_WriteToSysLog()` .... the choice is simple - one must first decide whether a "job half done" in the form of a truncated message is either acceptable or it is not acceptable.

If it is decided that it is acceptable to store only a partial log message, then `CFE_ES_WriteToSysLog()` should return `CFE_SUCCESS`, not `CFE_ES_ERR_SYS_LOG_TRUNCATED` when this happens, because it is not an error.  If the result is acceptable, it means the function was successful.

If it is decided that it is not acceptable to store only a partial log message, then `CFE_ES_WriteToSysLog()` should return an error code failure when this happens and write nothing at all to the log.

This way it either succeeds completely or does nothing - there should be no half-done/middle-ground/gray area where it did something but not the whole thing.

jphickey:
I went ahead and documented all the cases of extraneous CFE "success" status/information codes.  Most offenders are in TBL, a few in ES.  Here is my list along with a synopsis.


CFE_ES_CDS_ALREADY_EXISTS:
    - CFE_ES_RegisterCDSEx: does nothing in this case 
    - should be an error code, not success code.
    
CFE_ES_LIB_ALREADY_LOADED:
    - CFE_ES_LoadLibrary: no action but does return the ID.
    - Library API is (currently) non-public and nothing handles this output
    - should be an error code, not success code.
    - no need to return id - make a separate API to find a lib by name.
    
CFE_ES_ERR_SYS_LOG_TRUNCATED:
    - CFE_ES_SysLogAppend_Unsync (internal) but returned through e.g. CFE_ES_WriteToSysLog
[...]

PR #872, opened 2020-09-09
Fix #846, Deconflict CFE_ES_LIB_ALREADY_LOADED and CFE_ES_ERR_SYS_LOG_TRUNCATED

### LINKED ISSUE #846: CFE_ES_ERR_SYS_LOG_TRUNCATED Value Isn't Unique
**Describe the bug**
`CFE_ES_LIB_ALREADY_LOADED` and `CFE_ES_ERR_SYS_LOG_TRUNCATED` are both defined to be `((int32)0x44000028)` (in [`cfe_error.h`](https://github.com/nasa/cFE/blob/06c08268f2d7a7fe72aceec8c33201ce6ccba2f4/fsw/cfe-core/src/inc/cfe_error.h#L634)).

**To Reproduce**
N/A

**Expected behavior**
I think `CFE_ES_ERR_SYS_LOG_TRUNCATED` should instead be set to `((int32)0xC4000029)` or some other unique value that follows the status code format.

**Code snips**
N/A

**System observed on:**
N/A

**Additional context**
N/A

**Reporter Info**
Keegan Moore, NASA/GSFC

skliper:
**Describe the contribution**
Fix #846 - deconflict CFE_ES_LIB_ALREADY_LOADED and CFE_ES_ERR_SYS_LOG_TRUNCATED EIDs

**Testing performed**
Built and ran unit tests (checks both those returns), passed.

**Expected behavior changes**
EIDs no longer overloaded.

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: bundle main + this commit

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_error.h
+++ b/fsw/cfe-core/src/inc/cfe_error.h
@@ -631,7 +631,7 @@
  *  due to insufficient space in the log buffer.
  *
  */
-#define CFE_ES_ERR_SYS_LOG_TRUNCATED  ((int32)0x44000028)
+#define CFE_ES_ERR_SYS_LOG_TRUNCATED  ((int32)0x44000029)
 
 /**
  * @brief Not Implemented


skliper:
Just realized these aren't EIDs... error codes.  I'll fix commit.

jphickey:
See also #616, seems to be a duplicate

keegan-moore:
Should the value start with 0xC, to indicate that this return value is an error?

skliper:
> Should the value start with 0xC, to indicate that this return value is an error?

I left them with the "Informational" severity code, this change just deconflicts.  I'd recommend severity code changes be a new issue if you want to re-evaluate them.

keegan-moore (approved):
We might want to open a new issue to discuss changing this value to be an error type (start with 0xC instead of 0x4).

skliper:
> We might want to open a new issue to discuss changing this value to be an error type (start with 0xC instead of 0x4).

See also #483 for the general ticket.

jphickey:
My take is that the log getting truncated isn't really an error - its informational - because the message WAS logged.  An error code should indicate no action was performed.  That's so the caller has a recourse - they can check for the error, and do some sort of mitigation, possibly re-try the call.

In this situation - its not like anyone can take useful action on this response - code is not going to check for this condition and log a shorter message or something.  (which would end up as a duplicate too).

So eventually (and in line with #483) we should get rid of `CFE_ES_ERR_SYS_LOG_TRUNCATED` entirely.  Just return `CFE_SUCCESS`.  If anything, ES can internally track the number of times this happens so the user knows they need to increase `CFE_PLATFORM_ES_SYSTEM_LOG_SIZE` ... but returning it to apps as a status code makes no sense.  What is a generic app going to do with that info?

jphickey:
FWIW - in the framework I see 278 references to `CFE_ES_WriteToSysLog()` and with the exception of unit test cases - none of them actually check the returned status _at all_ or do any sort of recovery if the syslog doesn't work or is truncated or anything.

(Which makes sense when you think about it, because we wouldn't expect a CFS app to handle system-level issues like a broken syslog - it isn't its domain).

keegan-moore:
I looked at this from a different perspective. It seems like CFE_ES_ERR_SYS_LOG_TRUNCATED was created to tell the calling app that the app was using the syslog wrong, and the function call didn't work as expected. I wouldn't consider this to mean that syslog was broken, I'd consider that to be an app error, and I'd expect CFE_ES_WriteToSysLog() to return some type of non-CFE_SUCCESS code if an app used the API incorrectly. It would be great if 0x2 was used to indicate a warning, or something like that.
Even though apps don't typically look at that return value, _if it's going to return a value_, I don't think it should return CFE_SUCCESS in that case.
Thinking out loud: Knowing that most apps don't check the return, I'd say it would be easier for ES to report that error somehow. I'm not sure what the best mechanism for that would be, though.

jphickey:
I see this as a case where the app invoked `CFE_ES_WriteToSysLog()` perfectly well, but was limited by the internal constraints of the implementation.  

At any rate, happy to discuss the philosophy around error codes further in #483 .... should probably transition discussion over there ...

PR #887, opened 2020-09-14
Fix #886, library table id management

### LINKED ISSUE #886: Update Library code to decouple IDs from table indices
Library IDs are currently coupled to the Library table index.   

Library IDs should be opaque values and not interpreted directly as a table index.
Apply the same pattern as in previous PRs #859 and #871 to the Library code.

**Additional context**
Part of ongoing effort to clean up IDs across all of ES.  

**Requester Info**

jphickey:
**Describe the contribution**
Apply the appid/taskid/counterid pattern to Library resources.

**Testing performed**
Build and sanity check CFE, ensure library is loaded correctly
Run all unit tests

**Expected behavior changes**
No impact to behavior.  Internal change only.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Implement the same pattern on Library IDs which was previously implemented for apps, tasks, and counters.

For unit testing this also moves the Library-specific UT calls into a separate function.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_api.c
+++ b/fsw/cfe-core/src/es/cfe_es_api.c
@@ -1686,6 +1686,25 @@ int32 CFE_ES_AppID_ToIndex(uint32 AppId, uint32 *Idx)
 }
 
 /*
+ * A conversion function to obtain an index value correlating to a LibID
+ * This is a zero based value that can be used for indexing into a table.
+ */
+int32 CFE_ES_LibID_ToIndex(uint32 LibId, uint32 *Idx)
+{
+    if (LibId >= CFE_PLATFORM_ES_MAX_LIBRARIES)
+    {
+        return CFE_ES_BAD_ARGUMENT; /* these do not have a dedicated error */
+    }
+
+    /*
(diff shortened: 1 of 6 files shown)


jphickey:
Also note that this cleans up a previously leftover debug `#ifdef` in UT code while also moving the code to a separate function.

CDKnightNASA (approved):
I will create a ticket to define typedefs for ID's and return status codes. For now they can be simple typedefs to the uint32 type but eventually should be wrapped or otherwise we should have a mechanism to ensure that they're not misused.

jphickey:
> I will create a ticket to define typedefs for ID's and return status codes. For now they can be simple typedefs to the uint32 type but eventually should be wrapped or otherwise we should have a mechanism to ensure that they're not misused.

Sounds good - this was on my eventual plans too.  If you write tickets I only request they are separate ones (typedef for ids, typedef for returns) as each is a separate dev effort to do.

astrogeco:
[CCB 2020-09-16](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-16-2020) APPROVED

PR #889, opened 2020-09-16
strawman for #888 - return code for cFE

### LINKED ISSUE #888: typedef for status return values
Currently most CFE API's return a status code (the return type defines as int32.) The numeric return type is somewhat unclear and could result in confusion with other types of information (such as inadvertent confusion with return values from OSAL functions or numeric ID's).

```
typedef int32 CFE_Status_t;

CFE_Status_t CFE_SB_CreatePipe()...
```

In order to facilitate better coding practices, I also propose a utility function:
```
bool CFE_Status_IsSuccess(CFE_Status_t status); // returns true if the status is a success return value, false if not (currently < 0 is an error condition, >= 0 is success with possible further status details.)
```

This will require quite a lot of code change but could be migrated incrementally as creating the typedef to the same type will be compatible. Eventually it could be changed to a struct wrapper or enum or something that will result in compiler errors when the type is considered a numeric type.

**Requester Info**
Christopher.D.Knight@nasa.gov

CDKnightNASA:
closes #888  -- not like this strawman will close it, but just creating the linkage

**Describe the contribution**
This is a strawman for return code typedef for cFE.

**Testing performed**
code builds

**Expected behavior changes**
Returns for cFE functions now typedef'd in headers (.h), not in the source (.c).

Christopher.D.Knight@nasa.gov

### DIFF
--- a/fsw/cfe-core/src/inc/cfe_error.h
+++ b/fsw/cfe-core/src/inc/cfe_error.h
@@ -43,6 +43,19 @@
 #include "osapi.h"
 
 /*
+ * Define a type for readability.
+ */
+
+typedef int32 CFE_Status_t;
+typedef CFE_Status_t CFE_RC_t;
+
+/*
+ * Utility macros.
+ */
+#define CFE_ISSUCCESS(S) ((S) & CFE_SEVERITY_BITMASK != CFE_SEVERITY_ERROR)
+#define CFE_ISINFO(S) ((S) & CFE_SEVERITY_BITMASK == CFE_SEVERITY_INFO)
+
+/*
 **  Status Codes are 32 bit values formatted as follows:
 **
 **   3 3 2 2 2 2 2 2 2 2 2 2 1 1 1 1 1 1 1 1 1 1
--- a/fsw/cfe-core/src/inc/cfe_es.h
+++ b/fsw/cfe-core/src/inc/cfe_es.h
@@ -41,6 +41,7 @@
 #include "cfe_es_extern_typedefs.h"
 #include "cfe_mission_cfg.h"
 #include "cfe_perfids.h"
+#include "cfe_error.h"
 
 /*****************************************************************************/
 
@@ -176,7 +177,7 @@ static inline uint32 CFE_ES_ResourceID_FromInteger(unsigned long Value)
  * @return Execution status, see @ref CFEReturnCodes
  * @retval #CFE_SUCCESS                 @copybrief CFE_SUCCESS
  */
-int32 CFE_ES_AppID_ToIndex(uint32 AppID, uint32 *Idx);
+CFE_RC_t CFE_ES_AppID_ToIndex(uint32 AppID, uint32 *Idx);
 
 /**
  * @brief Obtain an index value correlating to an ES Task ID
@@ -200,7 +201,7 @@ int32 CFE_ES_AppID_ToIndex(uint32 AppID, uint32 *Idx);
  * @return Execution status, see @ref CFEReturnCodes
  * @retval #CFE_SUCCESS                 @copybrief CFE_SUCCESS
  */
-int32 CFE_ES_TaskID_ToIndex(uint32 TaskID, uint32 *Idx);
+CFE_RC_t CFE_ES_TaskID_ToIndex(uint32 TaskID, uint32 *Idx);
 
 /**
  * \brief Application Information
@@ -407,7 +408,7 @@ void CFE_ES_Main(uint32 StartType, uint32 StartSubtype, uint32 ModeId , const ch
 ** \sa #CFE_ES_Main
 **
 ******************************************************************************/
-int32  CFE_ES_ResetCFE(uint32 ResetType);
+CFE_RC_t  CFE_ES_ResetCFE(uint32 ResetType);
 /**@}*/
 
 /** @defgroup CFEAPIESAppControl cFE Application Control APIs
@@ -431,7 +432,7 @@ int32  CFE_ES_ResetCFE(uint32 ResetType);
 ** \sa #CFE_ES_ReloadApp, #CFE_ES_DeleteApp
 **
 ******************************************************************************/
-int32 CFE_ES_RestartApp(uint32 AppID);
+CFE_RC_t CFE_ES_RestartApp(uint32 AppID);
 
 /*****************************************************************************/
 /**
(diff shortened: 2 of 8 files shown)


CDKnightNASA:
note also I did a bit of formatting cleanup (joining multi-line into single line when it fits)

fsw/cfe-core/src/inc/cfe_error.h
```diff
@@ -42,6 +42,19 @@
 /* Include Files */
 #include "osapi.h"
 
+/*
+ * Define a type for readability.
+ */
+
+typedef int32 CFE_Status_t;
+typedef CFE_Status_t CFE_RC_t;
```
jphickey, line 50: Why have two typedefs?

fsw/cfe-core/src/inc/cfe_error.h
```diff
@@ -42,6 +42,19 @@
...
+
+/*
+ * Utility macros.
+ */
+#define CFE_ISSUCCESS(S) ((S) & CFE_SEVERITY_BITMASK != CFE_SEVERITY_ERROR)
+#define CFE_ISINFO(S) ((S) & CFE_SEVERITY_BITMASK == CFE_SEVERITY_INFO)
```
jphickey, line 56: can these be inline functions instead?  These are just as fast but more type safe and don't have the same risks/issues as macros.

jphickey (commented):
Just a quick look through, some more comments inline.

But also note that in order to be really type-agnostic, one must offer an inline function (or macro) to test for any value, not just success.  That's because conditionals like:

    if (Status == CFE_ES_ERR_TASKID)

no longer work if using a type for which a numeric equality test isn't implemented.  In C++ it would be easy to simply provide a `==` operator, but in C we have to provide an explicit `IsEqual` function.

jphickey:
And also, FWIW, I'm still strongly in favor of getting rid of the notion of multiple success variants as indicated in #483.

Functions should have either SUCCESS -- i.e. it worked fully and completely, or it FAILED - nothing in between. 

Having multiple different values for failure is fine - as it can indicate why it failed - but all of them should mean the same thing which is that the function did nothing at all (it failed).  There shouldn't be any cases of "half-worked" -- that means the API is too complex.  Any current API that returns half-worked "sorta success" codes should be split up and simplified into units that can fit the model of all or nothing.

CDKnightNASA:
> And also, FWIW, I'm still strongly in favor of getting rid of the notion of multiple success variants as indicated in #483.
> 
> Functions should have either SUCCESS -- i.e. it worked fully and completely, or it FAILED - nothing in between.
> 
> Having multiple different values for failure is fine - as it can indicate why it failed - but all of them should mean the same thing which is that the function did nothing at all (it failed). There shouldn't be any cases of "half-worked" -- that means the API is too complex. Any current API that returns half-worked "sorta success" codes should be split up and simplified into units that can fit the model of all or nothing.

CFE_TBL is a standout for using non-error returns....Feel free to peruse the .h and consider alternative models? I can file a ticket for a redesign of CFE_TBL to remove the "WARN" and "INFO" return statuses.

fsw/cfe-core/src/inc/cfe_error.h
iamthad: Just as fast if inlining is enabled. If it's not, and these are "real" functions, there's a decent amount of overhead.

If they are declared inline, they should be `static inline` at least, to avoid GOT/PLT lookup and such.

fsw/cfe-core/src/inc/cfe_es.h
```diff
@@ -176,7 +177,7 @@ static inline uint32 CFE_ES_ResourceID_FromInteger(unsigned long Value)
  * @return Execution status, see @ref CFEReturnCodes
  * @retval #CFE_SUCCESS                 @copybrief CFE_SUCCESS
  */
-int32 CFE_ES_AppID_ToIndex(uint32 AppID, uint32 *Idx);
+CFE_RC_t CFE_ES_AppID_ToIndex(uint32 AppID, uint32 *Idx);
```
astrogeco, line 180: Can we use `CFE_ReturnCode_t` ? I like spelling out things as much as possible

astrogeco:
[CCB 2020-09-16](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-16-2020) Discussed benefits and problems with multiple types. Also talked about the problems with multiple "success" and "failure" return codes. We should always ask, what is the purpose of the code? How can it be used?

fsw/cfe-core/src/inc/cfe_es.h
CDKnightNASA: A lot of typing, and causes more line wrapping...I'm open though if that's the consensus.

astrogeco: Ah, I use autocomplete on Atom. I assume this type would be used by app developers so I'd rather be explicit. If it's just internal to cFE then I feel slightly better about abbreviations

jphickey:
> CFE_TBL is a standout for using non-error returns....Feel free to peruse the .h and consider alternative models? I can file a ticket for a redesign of CFE_TBL to remove the "WARN" and "INFO" return statuses.

FYI - see my synopsis here for the various functions that need attention:
https://github.com/nasa/cFE/issues/483#issuecomment-693602105

Probably could have a whole separate discussion on each function.

CDKnightNASA:
One vote for the naming discussion, note that "RC" is a term/acronym used throughout event messages and documentation in `cfe_es_events.h`, such as:

https://github.com/nasa/cFE/blob/60917a2d7e23b32419a9b11b0d2c0bd722d7c35e/fsw/cfe-core/src/inc/cfe_es_events.h#L421

astrogeco:
I'm biased against acronyms and initialisms, especially after reading ["Acronyms Seriuosly Suck"](https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764) could "RtrnCode" be a good compromise?

jphickey:
> I'm biased against acronyms and initialisms, especially after reading ["Acronyms Seriuosly Suck"](https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764). Could "RtrnCode" be a good compromise?

I am NOT a fan of the vowel-removal approach -- code that does this I find really ugly and hard to read.  I also have a slight preference for full words, despite the fact that it makes long lines, I think the result is overall more readable.  I also use an editor with autocomplete so the typing isn't a strong concern.

My preference is `CFE_Status_t`.  Because many times it is referred to as a a status code and stored in a variable called "Status".

I'm also OK with: `CFE_ReturnCode_t`, `CFE_Error_t`, `CFE_RC_t`.

But please - lets steer clear of quasi-abbreviations like "RtrnCode" -- this looks more like a letter jumble.

skliper:
I vote CFE_Status_t for the same reasons @jphickey states.

CDKnightNASA:
consensus is CFE_Status_t, will update PR prior to next CCB.

skliper:
Marked as duplicate of #919 (this PR was the strawman for discussion).

PR #898, opened 2020-09-22
Fix #897, EVS unregistered AppID

### LINKED ISSUE #897: EVS improperly uses the EventCount member for incorrect appID filtering
**Describe the bug**
The `EventCount` member of the EVS internal app data is intended to track the number of events sent by a particular app.

BUT - it is also overloaded to track if the event came from the wrong app, here:
https://github.com/nasa/cFE/blob/983157db90bd205977c52762506ccbf2132837f3/fsw/cfe-core/src/evs/cfe_evs_utils.c#L187-L194

The problem is that this is, by definition, invoked when the calling AppID is _wrong_ (not registered), but it is changing a field that is also potentially used for _valid_ (registered) AppIDs.  

**To Reproduce**
This can theoretically occur if an app calls `CFE_EVS_SendEventWithAppId()` using an old AppID value, for instance if an app was restarted it gets unregistered, and then gets a different AppID but refs to the old value could still exist.  The new AppID doesn't necessarily have the same slot in the table - in fact it shouldn't.  The old table entry might refer to a totally different app.

So if this happens it will corrupt/change the `EventCount` member on an unrelated app data entry.

This can be seen in the EVS telemetry, where if the "unregistered" event occurs it inadvertently creates a nonzero `AppMessageSentCounter` in the TLM data on an unrelated app that happens to share that slot in the table.

Also, if the counter was already nonzero because the table entry is in use by another (registered) app then this prevents the notification about the unregistered app from appearing at all.

**Expected behavior**
Should not overload the EventCounter to track a basically unrelated item - which is whether or not an "unregistered" event occurred on a different app that happened to map to the same entry.

Simple fix would be to just introduce a separate field to track this.

**System observed on:**
Ubuntu 20.04

**Reporter Info**

jphickey:
**Describe the contribution**
Do not overload the `EventCount` field to track whether an "unregistered" event was generated, which by definition comes from a different app than the one that might be actively using this same table entry.

This just makes a separate field to track that state, leaving the `EventCount` to serve its primary purpose of counting actual events generated from a valid/registered AppID.

Fixes #897 

**Testing performed**
Temporarily modified/hacked "sample_app" to send several events _before_ calling `CFE_EVS_Register()`.
Confirmed that one (and only one) "unregistered" event is reported e.g.:

    EVS Port1 42/1/CFE_EVS 41: App SAMPLE_APP not registered with Event Services. Unable to send event.

Also confirmed that this "unregistered" event did not change the valid `EventCount` in the EVS telemetry data.

**Expected behavior changes**
No more incrementing a counter for another app.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/fsw/cfe-core/src/evs/cfe_evs_task.h
+++ b/fsw/cfe-core/src/evs/cfe_evs_task.h
@@ -89,6 +89,8 @@ typedef struct
 typedef struct
 {
     CFE_ES_ResourceID_t AppID;
+    CFE_ES_ResourceID_t UnregAppID;
+
     EVS_BinFilter_t    BinFilters[CFE_PLATFORM_EVS_MAX_EVENT_FILTERS];  /* Array of binary filters */
 
     uint8              ActiveFlag;             /* Application event service active flag */
--- a/fsw/cfe-core/src/evs/cfe_evs_utils.c
+++ b/fsw/cfe-core/src/evs/cfe_evs_utils.c
@@ -185,13 +185,13 @@ int32 EVS_NotRegistered (EVS_AppData_t *AppDataPtr, CFE_ES_ResourceID_t CallerID
    char AppName[OS_MAX_API_NAME];
 
    /* Send only one "not registered" event per application */
-   if (AppDataPtr->EventCount == 0) 
+   if ( !CFE_ES_ResourceID_Equal(AppDataPtr->UnregAppID, CallerID) )
    { 
       /* Increment count of "not registered" applications */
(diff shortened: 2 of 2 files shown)


jphickey:
Rebased to current main branch

astrogeco:
[CCB 2020-09-30](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-30-2020): APPROVED

astrogeco:
@jphickey do we need to document this new Unregistered "counter" somewhere?

PR #899, opened 2020-09-22
Fix #705, Use cFE defines to size arrays

### LINKED ISSUE #705: CFE_ES_OneAppTlm_Payload_t telemetry struct contains non mission-scoped-sized array
Unsure if this is intentional, but 
`CFE_ES_AppInfo_t` is directly used by `CFE_ES_OneAppTlm_Payload_t` which is ground visible. `CFE_ES_AppInfo_t` uses `OS_MAX_*`, whereas all other ground-visible telemetry uses the mission-scoped `CFE_MISSION_MAX_*`.

`CFE_ES_OneAppTlm_Payload_t` should be structured such that the size is dependent solely on mission-scoped constants and consistent amongst all CPUs. 

This is an enhancement as opposed to a bug report since in practice both constants are likely to be same. 

None

**Additional context**
None

**Requester Info**
John N Pham, Northrop Grumman

skliper:
**Describe the contribution**
Fixes #705, also a similar issue in TBL
Change from OS_MAX* defines to CFE defines for array sizing

**Testing performed**
Built and ran unit tests, passed.  Visually confirmed sizes matched.

**Expected behavior changes**
No impact as long as the sizes are the same.  Now scoped appropriately such that OS's can be configured differently and won't break cFE (as long as the cFE defines are max of the set)

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: bundle main + this commit

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_apps.c
+++ b/fsw/cfe-core/src/es/cfe_es_apps.c
@@ -1531,9 +1531,9 @@ int32 CFE_ES_GetTaskInfoInternal(CFE_ES_TaskRecord_t *TaskRecPtr, CFE_ES_TaskInf
       ** Get the Application ID and Task Name
       */
       TaskInfoPtr->AppId = TaskRecPtr->AppId;
-      strncpy((char *)TaskInfoPtr->TaskName,
-              (char *)TaskRecPtr->TaskName,OS_MAX_API_NAME);
-      TaskInfoPtr->TaskName[OS_MAX_API_NAME - 1] = '\0';
+      strncpy((char*)TaskInfoPtr->TaskName, TaskRecPtr->TaskName,
+              sizeof(TaskInfoPtr->TaskName)-1);
+      TaskInfoPtr->TaskName[sizeof(TaskInfoPtr->TaskName)-1] = '\0';
 
       /*
       ** Store away the Task ID ( for the QueryAllTasks Cmd )
@@ -1551,9 +1551,9 @@ int32 CFE_ES_GetTaskInfoInternal(CFE_ES_TaskRecord_t *TaskRecPtr, CFE_ES_TaskInf
       AppRecPtr = CFE_ES_LocateAppRecordByID(TaskRecPtr->AppId);
       if (CFE_ES_AppRecordIsMatch(AppRecPtr, TaskRecPtr->AppId))
       {
-         strncpy((char*)TaskInfoPtr->AppName,
-                 (char*)AppRecPtr->StartParams.Name,
+         strncpy((char*)TaskInfoPtr->AppName, AppRecPtr->StartParams.Name,
                  sizeof(TaskInfoPtr->AppName)-1);
+         TaskInfoPtr->AppName[sizeof(TaskInfoPtr->AppName)-1] = '\0';
          ReturnCode = CFE_SUCCESS;
       }
       else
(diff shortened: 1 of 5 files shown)


jphickey (changes requested):
Seems like there is more work to do here....  ?

- As the structures (`CFE_ES_AppInfo_t`) are actually part of the IPC interface (msg), the definition of these shouldn't be in `cfe_es.h` but rather in `cfe_es_msg.h` where all other structures used within messages are defined.
- Anywhere that the `Name`, `EntryPoint`, or `FileName` members are filled should use `CFE_SB_MessageStringSet`- which is similar to strncpy but takes both a source and dest length.  It is intended for this use case where the local/platform and global/mission sizes might be different and this handles padding/termination issues.

skliper:
> Seems like there is more work...

I suggest new issue(s) since at least the second change sounds like a useful one to get documented separately, the old one just covered the array sizing.

CDKnightNASA (approved):

CDKnightNASA:
FYI given this development, I'm wondering of the wisdom of having return status int32's intermixing CFE and OSAL return status...I realize that'll mean a lot of duplication where we remap OSAL status to CFE status with the same name, but in a lot of cases I'm guessing CFE can simplify the return status.

astrogeco:
[CCB 2020-09-23](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-23-2020) write up extra issues based on conversation. @jphickey to update function calls.

skliper:
> FYI given this development, I'm wondering of the wisdom of having return status int32's intermixing CFE and OSAL return status...I realize that'll mean a lot of duplication where we remap OSAL status to CFE status with the same name, but in a lot of cases I'm guessing CFE can simplify the return status.

Is this in reference to a different issue?  This just changed array sizes...

CDKnightNASA:
> > FYI given this development, I'm wondering of the wisdom of having return status int32's intermixing CFE and OSAL return status...I realize that'll mean a lot of duplication where we remap OSAL status to CFE status with the same name, but in a lot of cases I'm guessing CFE can simplify the return status.
> 
> Is this in reference to a different issue? This just changed array sizes...

Just saying, philosophically we are "divorcing" array sizes from OSAL defs, while we continue to be "married" to the OSAL return codes. :) A general philosophical matter but one I think is a bit of conflict.

jphickey:
@skliper - My suggested update to fix the output sizes is here: https://github.com/skliper/cFE/compare/fix705-rm-os-dependence...jphickey:fix-705-outarray-size

If you want, you can either merge this into your PR or I can submit a new PR that has both.

To summarize, I decided against using `CFE_SB_MessageStringSet()` because it (intentionally) fills the output buffer completely, foregoing a null terminator if it doesn't fit.  But due to the usage patterns of these structs we should always ensure null termination.  So continuing to use the C library `strncpy()` routine is no worse, we just have to make sure to use `sizeof()` rather than assume a specific macro/symbol.

astrogeco:
[CCB 2020-09-30](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-30-2020): 
- There are some cases where we'll get unit test failures, changeing cFE defined OS_size
  - @skliper open new ticket to fix

APPROVED

astrogeco:
@skliper can you clear the conflicts?

skliper:
@astrogeco conflicts resolved.

PR #905, opened 2020-09-23
Fix #903, Add CFE_SB_GetUserData padding check

### LINKED ISSUE #903: Update CFE_SB_GetUserData unit test to catch padding differences between header/payload
CFE_SB_GetUserData returns the location right after the header, which isn't always the start of the payload if padding gets inserted.

Add generic packets with all width types, and ensure the correct location is returned.

None

**Additional context**
None

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
**Describe the contribution**
Fix #903 - this adds checks to see that CFE_SB_GetUserData works with all payload data types.

Illustrates the issue in #488 - As expected it fails for 64 bit payload in a telemetry packet because of implicit padding added between the header (12 bytes) and the 64bit payload (requires 64bit alignment.. so pads to 16 bytes).

Cmds are only OK because the default header is 8 bytes.

Also note - I expect these tests would also throw cast-align errors for an alignment sensitive build, since there's a cast from a structure with 64bit alignment to 32bit alignment.

**Testing performed**
Build and run unit tests... currently failing due to implicit padding.

**Expected behavior changes**
None, except shows the current issue with our header definitions.

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: main bundle + this commit

**Additional context**
Depends on nasa/osal#605 (address equal assert)

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/unit-test/sb_UT.c
+++ b/fsw/cfe-core/unit-test/sb_UT.c
@@ -3661,9 +3661,49 @@ void Test_CFE_SB_MsgHdrSize(void)
 void Test_CFE_SB_GetUserData(void)
 {
     CFE_MSG_Message_t msg;
-    uint8            *ExpAdrReturned;
+    uint8            *expected;
     bool              hassec;
     CFE_MSG_Type_t    type = CFE_MSG_Type_Invalid;
+    struct
+    {
+        CFE_MSG_CommandHeader_t cmd;
+        uint8                   payload;
+    } cmd_uint8;
+    struct
+    {
+        CFE_MSG_CommandHeader_t cmd;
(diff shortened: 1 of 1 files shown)


skliper:
This breaks the build (will fail unit tests), don't merge until #488 is fixed.

skliper:
Depends on #1077 to pass

astrogeco:
[CCB:2021-01-13](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-13-2021) APPROVED

Issue #854, opened 2020-08-31
Cast align issue in cfe_es_task.c - CFE_ES_TaskPipe


skliper:
**Describe the bug**
```
/home/jhageman/cFS-GitHub/cfe/fsw/cfe-core/src/es/cfe_es_task.c: In function ‘CFE_ES_TaskPipe’:
/home/jhageman/cFS-GitHub/cfe/fsw/cfe-core/src/es/cfe_es_task.c:635:52: warning: cast increases required alignment of target type [-Wcast-align]
  635 |                         CFE_ES_SendMemPoolStatsCmd((CFE_ES_SendMemPoolStats_t*)Msg);
```

**To Reproduce**
Build on gcc 8+ with cast-align=strict

**Expected behavior**
Clean build

**Code snips**
https://github.com/nasa/cFE/blob/06c08268f2d7a7fe72aceec8c33201ce6ccba2f4/fsw/cfe-core/src/es/cfe_es_task.c#L632-L637

**System observed on:**
 - Hardware: arm64
 - OS: Ubuntu 20.04
 - Versions bundle + mods to cast-align=strict

**Additional context**
#851, #852 

**Reporter Info**
Jacob Hageman - NASA/GSFC

skliper:
@jphickey are you fixing this with the other mempool command issues (reference instead of raw address)?

jphickey:
Yes, the reason this message has increased alignment needs is because it contains a `cpuaddr` type.  When replaced with an abstract pool handle (which can be `uint32`) the alignment issue goes away.

skliper:
Confirmed this has been resolved (via #917).  Closing

Issue #56, opened 2019-09-30
Consolidate CDS and generic/ram mempool code into single implementation


skliper:
Currently there are two memory pool implementations in ES:

* cfe_es_cds_mempool.c
* cfe_esmempool.c

These two are very similar except that the CDS uses offsets rather than direct pointers, and it calls into the PSP to perform actual read/write functions.  It would not be very hard to consolidate these into a single implementation.

Each implementation consumes about 4-5kB of code/data space so this consolidation would make ES a little smaller in addition to making it cleaner.

skliper:
_Imported from trac issue 25.  Created by jphickey on 2015-01-21T13:33:26, last modified: 2019-03-01T15:27:58_

skliper:
_Trac comment by jphickey on 2015-05-19 09:39:35:_

This is an old enhancement that is yet to be merged in.  Higher priority items taking precedence right now.

jphickey:
With recent addition of a generic (offset-based) memory pool implementation that works with the abstract resource IDs in #917.  I have also tested that this works for CDS blocks too.  Discussed at 2020-09-30 CCB that this would be a good time to finally do this and make one mempool implementation.

PR #917, opened 2020-09-29
Fix #651, Use resource ID for memory pools

### LINKED ISSUE #651: CFE_ES_SEND_MEM_POOL_STATS_CC Issues
**Describe the bug**
There are multiple issues with the CFE_ES_SEND_MEM_POOL_STATS_CC command:
1.	The command/function(s) used are fundamentally flawed in that a user can send a seemingly benign command that can result in a segmentation fault if one of the command parameters is incorrect.  For example, if the PoolHandle parameter is set to zero a segmentation fault will result (pretty much any value below xFFFFFE28 faults on my machine).  This occurs when handle is validated via the CFE_ES_ValidateHandle function.  Should consider modifying the CFE_ES_ValidateHandle function + updating the unit test to test command on boundary/extreme conditions.
2.	On a 64-bit machine if the PoolHandle parameter is set to a valid value the function will fail.  This is because CFE_PSP_MemValidateRange – which is called via the CFE_ES_ValidateHandle function as part of the validation process, limits the max memory range of the handle to xFFFFFFFE.
3.	On a 64-bit machine - compiler added padding will be applied to the command.  The order of variable declaration should ideally be descending in size to avoid future conflicts.

**Expected behavior**
Command works nominally and is vetted via combination of unit/functional tests.

**System observed on:**
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cFE 6.7.13.0, OSAL 5.0.12.0, PSP 1.4.9.0

**Reporter Info**
Dan Knutsen
NASA/Goddard

jphickey:
**Describe the contribution**

Instead of identifying a memory pool by its memory address, use a resource ID.  IDs are a constant size, regardless of
whether the host machine is 32 or 64 bits.

- IDs can be put into commands/telemetry and maintain a more consistent format with consistent alignment requirements.
- IDs can be independently verified without dereferencing memory.  Previously the only way to validate a memory pool was to read the address pointed to, which results in a segfault if the address was bad.

Fixes #651 

**Testing performed**
Build and sanity test CFE, run all unit tests
Confirm code coverage up to par.

**Expected behavior changes**
The `CFE_ES_MemHandle_t` type is no longer a direct CPU address.  Instead it is an abstract resource identifier.
This should fix issues with unexpected padding of CMD/TLM messages that contain memory pool handles.

**System(s) tested on**
Ubuntu 20.04
RTEMS 4.11

**Additional context**
Fairly extensive changes in the memory pool implementation to support abstract handles, but the API exposed to applications should be backward compatible, so long as apps did not rely on the specific pool layout.

The downside is that there is a limit to the total number of abstract identifiers that can exist, which is a new platform config limit.


jphickey:
Confirmed this also fixes #854

astrogeco:
[CCB 2020-09-30](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-30-2020):
- Creates a memory pool implementation that mirrors how we handle other abstract resources 
APPROVED

astrogeco:
@jphickey is #854 the issue we discussed opening at the CCB?

jphickey:
> @jphickey is #854 the issue we discussed opening at the CCB?

No, the old issue about CDS mempool is in #56 ... it has been on the to-do list for about 5 years.

PR #925, opened 2020-09-30
Fix #923, Perform appid lookup early

### LINKED ISSUE #923: Repetitive AppID lookups in CFE_SB_SendMsgFull
**Describe the bug**
SB has some code to filter an app's own messages via `CFE_SB_PIPEOPTS_IGNOREMINE`.   This gets the appid and compares to the AppID of the pipe creator, and skips the destination if its a match:

https://github.com/nasa/cFE/blob/8a7dc8f2e434889287973c41d4a5cb4a48f48b44/fsw/cfe-core/src/sb/cfe_sb_api.c#L1359-L1369

The problem is that this is "inside the loop" of all destinations in the routing entry.  So it will (potentially) call `CFE_ES_GetAppId()` multiple times.  This also creates a double locking situation, because the SB lock is being held at the time this executes, and the ES lock needs to be acquired by `CFE_ES_GetAppId()`

**Expected behavior**
Code should query the caller AppID early, before taking the SB lock.

**Additional context**
The code works but is inefficient, and double locking is a potential deadlock.

**Reporter Info**

jphickey:
**Describe the contribution**

Move the AppID lookup to be early in the CFE_SB_SendMsgFull implementation. Avoids double locking between SB+ES and avoids a block-scope local variable.

Fixes #923 

**Testing performed**
Build and sanity check CFE
Run unit tests

**Expected behavior changes**
None.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -1197,6 +1197,7 @@ int32  CFE_SB_SendMsgFull(CFE_SB_Msg_t    *MsgPtr,
     CFE_SB_BufferD_t        *BufDscPtr;
     uint16                  TotalMsgSize;
     CFE_SB_MsgRouteIdx_t    RtgTblIdx;
+    CFE_ES_ResourceID_t     AppId;
     CFE_ES_ResourceID_t     TskId;
     uint32                  i;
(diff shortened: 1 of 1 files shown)


skliper (approved):

skliper:
@astrogeco @yammajamma - requesting fast-track.

CDKnightNASA (approved):

PR #934, opened 2020-10-01
"SenderReporting" from SB private "global" - removed

MartaTu:
The global variable removed because is only set once and never used.


zanzaben (changes requested):
Other references need to be removed

https://github.com/nasa/cFE/blob/307725172952e35953f7a2815d664602afcb4709/fsw/cfe-core/src/sb/cfe_sb_init.c#L102-L103

https://github.com/nasa/cFE/blob/307725172952e35953f7a2815d664602afcb4709/fsw/cfe-core/unit-test/sb_UT.c#L4099

skliper:
Closing as duplicate of #1099

PR #919, opened 2020-09-30
Fix #888, Add typedef for function return status codes

CDKnightNASA:
Closes #888 

**Describe the contribution**
This adds a typedef for return status codes for functions. I've also added a brief comment to CFE_TBL_GetStatus as it's behavior will likely change in the future in the hopes of not having a non-zero "info" status.

**Testing performed**
make install

**Expected behavior changes**
no impact to behavior

**System(s) tested on**
Ubuntu 20.04

**Contributor Info - All information REQUIRED for consideration of pull request**
Christopher.D.Knight@nasa.gov

PR #936, opened 2020-10-01
Integration Candidate 2020-10-07

### LINKED ISSUE #651: CFE_ES_SEND_MEM_POOL_STATS_CC Issues

### LINKED ISSUE #705: CFE_ES_OneAppTlm_Payload_t telemetry struct contains non mission-scoped-sized array

### LINKED ISSUE #888: typedef for status return values

### LINKED ISSUE #897: EVS improperly uses the EventCount member for incorrect appID filtering

### LINKED ISSUE #923: Repetitive AppID lookups in CFE_SB_SendMsgFull

astrogeco:
**Describe the contribution**
Fix #923 
Fix #897  
Fix #651
Fix #923
Fix #705
Fix #888

**Testing performed**
See Bundle: https://github.com/nasa/cFS/pull/145/checks

**Expected behavior changes**

PR #898 - Adds the field `UnregAppID` to track whether an "unregistered" event was generated, un-overloading the `EventCount` field to serve its primary purpose of counting actual events generated from a valid/registered AppID.

PR #925 - Move the AppID lookup execution to be early in the CFE_SB_SendMsgFull implementation. This avoids double locking between SB+ES and avoids a block-scope local variable.

PR #917 - Instead of identifying a memory pool by its memory address, use a resource ID. IDs are a constant size, regardless of whether the host machine is 32 or 64 bits.
  - IDs can be put into commands/telemetry and maintain a more consistent format with consistent alignment requirements.
  - IDs can be independently verified without dereferencing memory. Previously the only way to validate a memory pool was to read the address pointed to, which results in a segfault if the address was bad.

PR #899 - Change from `OS_MAX*` defines to appropriately-scoped CFE defines for array sizing

PR #919 - This creates the new `CFE_Status_t` typedef for function's return status codes. Also adds a note to `CFE_TBL_GetStatus` since its behavior will likely change in the future in the hopes of not having a non-zero "info" status.

**System(s) tested on**
Bundle CI - https://github.com/nasa/cFS/pull/145/checks

**Additional context**
Part of https://github.com/nasa/cFS/pull/145

**Third party code**
None

@jphickey 
@skliper 
@CDKnightNASA


astrogeco:
@jphickey it looks like #917 broke the build. Thankfully (maybe?) its doxygen stuff.

```
/home/travis/build/nasa/cFS/cfe/cmake/sample_defs/cpu1_platform_cfg.h:283: warning: explicit link request to 'CFE_ES_MAX_MEMPOOL_BLOCK_SIZES' could not be resolved

/home/travis/build/nasa/cFS/cfe/docs/src/cfe_es.dox:763: warning: explicit link request to 'CFE_ES_MAX_MEMPOOL_BLOCK_SIZES' could not be resolved

/home/travis/build/nasa/cFS/cfe/docs/src/cfe_es.dox:763: warning: explicit link request to 'CFE_ES_MAX_MEMPOOL_BLOCK_SIZES' could not be resolved
```

jphickey:
@astrogeco Fixed in commit ae0c043, doxygen looks clean now.

PR #939, opened 2020-10-07
Fix #56, Refactor CDS to use generic pool implementation

### LINKED ISSUE #56: Consolidate CDS and generic/ram mempool code into single implementation

jphickey:
**Describe the contribution**

Rather than having a second pool implementation only for CDS, use the generic pool implementation.  This also uses the abstract resource identifiers to identify CDS blocks, rather than a direct reference.

Fixes #56 

**Testing performed**
Build and sanity test CFE
Confirm all unit tests working
Also tested/Confirmed that the CFE TBL critical table registry is correctly restored when booting in a processor reset mode.  In this case the data is successfully restored from CDS.

**Expected behavior changes**
No impact to behavior.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This consolidates and simplifies a bunch of code in the CDS access area.
Note that previously there was a separate mutex for the CDS pool and CDS registry.  However almost all accesses needed both, because pool access and registry access go together.  So this is simplified to one mutex now.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_api.c
+++ b/fsw/cfe-core/src/es/cfe_es_api.c
@@ -1397,7 +1397,7 @@ uint32 CFE_ES_CalculateCRC(const void *DataPtr, uint32 DataLength, uint32 InputC
 ** Purpose:  Allocate a data block for a Critical Data Store.
 **
 */
-int32 CFE_ES_RegisterCDS(CFE_ES_CDSHandle_t *CDSHandlePtr, int32 BlockSize, const char *Name)
+int32 CFE_ES_RegisterCDS(CFE_ES_CDSHandle_t *CDSHandlePtr, CFE_ES_CDS_Offset_t BlockSize, const char *Name)
 {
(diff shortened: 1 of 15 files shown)


astrogeco:
[CCB 2020-10-06](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-October-06-2020) APPROVED
- Maybe transform the sample_app table into a critical table for testing and demonstration
- Might want to set up CI to do power-on and processor resets

acudmore (commented):
Looks good. I like the consolidation into a single memory pool implementation.

Issue #611, opened 2020-04-14
CFE_TBL_LoadFromFile() should automagically add extension


CDKnightNASA:
One issue that is quite an annoyance with the text-based startup file is that it lists the fully-qualified file name, including extension.  The issue is that the extension varies from system to system, so when I build with e.g. `SIMULATION=native` the extensions need to be `.so`, but when building for RTEMS it needs to be `.obj`.  

Ideally I'd like to see a solution that only lists the basename of the app file in the user-maintained source file, and the extra info, in particular the extension, is attached by the build system based on the file type of loadable modules for the particular platform you are building.

However the first step to any of this would be to get this table-based change in place, then smarter translation on the build side could be a follow-on.

_Originally posted by @jphickey in https://github.com/nasa/cFE/pull/588#issuecomment-613030439_

CDKnightNASA:
Note also Joe reported that CMake provides the ${CMAKE_SHARED_LIBRARY_SUFFIX} variable.

CDKnightNASA:
should wait for #503 to be merged into master.

PR #941, opened 2020-10-07
Fix #940, Add module suffix and core name to configdata

### LINKED ISSUE #940: Add system-specific module and library suffixes to configdata struct
There are a number of use cases where the FSW could be improved if it knows the proper system-specific extension to use for modules and executable files.

See existing issues #611 and nasa/PSP#111 

Should add this information to the `GLOBAL_CONFIGDATA` structure, to make it available for PSP/CFE purposes.

Could also be provided in the `GLOBAL_PSP_CONFIGDATA` structure that resides in the PSP library, but this would require explicitly setting in each PSP.  

Because CMake already knows this information and its readily available in CMake variables, its simpler and probably more reliable to just put it in the top level config struct based on the CMake value.

**Requester Info**
Joseph Hickey, Vantage Systems, Inc., (based off previous requests in other tickets)

jphickey:
**Describe the contribution**
Add the system-specific module suffix (.o, .so, .obj, etc) and the default CFE core executable name to the configdata structure.

This information has several useful purposes.

Fixes #940, but related to previous issues #611 and nasa/psp#111

**Testing performed**
Build and sanity check CFE on native/pc-linux, RTEMS 4.11, and VxWorks 6.9
Confirm using inspection of command line that the correct data is used for the new fields.

On linux:

    -DCFE_DEFAULT_CORE_FILENAME=\"core-cpu1\" -DCFE_DEFAULT_MODULE_EXTENSION=\".so\"

On RTEMS:

    -DCFE_DEFAULT_CORE_FILENAME=\"core-cpu1.exe\" -DCFE_DEFAULT_MODULE_EXTENSION=\".obj\"

On VxWorks:

    -DCFE_DEFAULT_CORE_FILENAME=\"core-cpu1.exe\" -DCFE_DEFAULT_MODULE_EXTENSION=\".o\"

Nothing uses this data at runtime (yet) but I also confirmed with GDB that the "GLOBAL_CONFIGDATA" indeed has the correct info.

**Expected behavior changes**
None.  Just making new information available.

**System(s) tested on**
Ubuntu 20.04 (native), RTEMS 4.11, VxWorks 6.9


jphickey:
Getting this info into the basic runtime config structure is a prerequisite to the PSP change to fix nasa/psp#111, so this should be merged first.

skliper:
Don't you also need CMAKE_EXECUTABLE_SUFFIX and CMAKE_SHARED_LIBRARY_SUFFIX?

https://github.com/nasa/PSP/blob/b383f7c657f97c997c8129a8853a5557eddef517/cmake/Modules/Platform/RTEMS.cmake#L12-L14
https://github.com/nasa/PSP/blob/b383f7c657f97c997c8129a8853a5557eddef517/cmake/Modules/Platform/VxWorks-CFE.cmake#L12-L14

jphickey:
For executables, the only executable that's relevant here is cfe core itself.  So I decided to just store the whole target filename, that saves the FSW from having to reassemble it, along with the assumptions thereof.  Also if we ever change the filename in CMake it should still just work.

All CFE apps/libs are built as modules i.e. add_library(... MODULE ...) so this is the only extension that should appear in load commands/startup scripts.  Furthermore I've never seen them be different - although CMake does differentiate between shared libs and modules, I don't know of any system where they are actually different file types.

IMO I didn't want to throw in every possible item, only the ones we actually know we care about.

skliper (approved):

skliper:
@astrogeco requesting fast track into the current IC, it's needed to fix broken behavior in PSP

jphickey:
@skliper FYI see also commit 655a66c, this prevents editors (or at least my editor) from complaining about undefined symbols.

skliper (approved):

skliper:
CCB: 2020-10-14: Approved.

PR #946, opened 2020-10-13
Fix #943, #924 and #502, improve resource ID management

### LINKED ISSUE #502: CFE ES unnecessarily keeping multiple copies of task/app names
When reviewing the changes for other tickets, noted that the CFE ES is storing the app name in the `AppTable` twice.  It is stored in the `StartParams.Name` subfield, as well as the `TaskInfo.MainTaskName` sub field.  These appear to be always set together, to the same value, such as in CFE_ES_AppCreate for example:

https://github.com/nasa/cFE/blob/3356da650f04b60190642be3efece47afa66b61a/fsw/cfe-core/src/es/cfe_es_apps.c#L554
https://github.com/nasa/cFE/blob/3356da650f04b60190642be3efece47afa66b61a/fsw/cfe-core/src/es/cfe_es_apps.c#L573

It then goes on to store the same string a third time in the TaskTable[x].TaskName field for the task itself:

https://github.com/nasa/cFE/blob/3356da650f04b60190642be3efece47afa66b61a/fsw/cfe-core/src/es/cfe_es_apps.c#L623-L624

The name is a string value and therefore takes a fair bit of memory to store.  In the "stock" example config this bloats the size of the ES data structures by about 2500 bytes, but could easily be much more in a real deployment if OS_MAX_API_NAME is set longer and/or the max number of apps/tasks is larger.

Should store _at most_ one copy of the name in the `AppTable`, but even that might not be needed if it is always the same as the main task name (which it appears to be).

For tasks, OSAL already stores the task name.  For the ES API calls that need to get the name, it should just lookup the name from OSAL, just like we are doing for SB in #404 

**Requester Info**

### LINKED ISSUE #924: Improve uniqueness in resource IDs
In the event that a resource "A" such as an app/counter/pool etc is created, then deleted, then another resource "B" is created, the code will re-issue the same resource ID for "B" as was used for "A".

This means if a stale reference to "A" is stored somewhere, it silently now becomes an alias to "B".  All validation/checks will pass.

Should not immediately recycle ID values.

**Additional context**
This is the way OSAL already works, and is enabled by using abstract IDs.

With the recently introduced resource ID abstractions, it is no longer necessary to limit to the respective table size(s), because IDs are not directly used as table indices anymore.  This means that a much larger set of potential ID values is available, and thus means that IDs of old/deleted objects can be different than the new objects while keeping table size the same.

**Requester Info**

### LINKED ISSUE #943: Provide consistent name lookup API
Many CFE resources have a name associated with them, but CFE is rather hit or miss on how Resource IDs (apps, tasks, libs, etc) can be looked up by name.

For instance, `CFE_ES_GetAppIDByName()` exists, but `CFE_ES_GetLibIDByName()` does not.  Task IDs can be found directly via OSAL through `OS_TaskGetIdByName()` but there is no CFE version of this despite the fact that ES also tracks tasks.

For resource types that have names associated, there should be a consistent API to find the ID by name, similar to `CFE_ES_GetAppIDByName()`.  The following should be implemented, all with the general API pattern (name in, ID out, status return):

    CFE_ES_GetTaskIDByName()
    CFE_ES_GetLibIDByName()
    CFE_ES_GetGenCounterIDByName()
    CFE_ES_GetCDSBlockIDByName()

Keep things inconsistent.

**Additional context**
This is also a prerequisite to #28 - to have libs be manageable similarly to apps, part of that is being able to find lib ID by name, which is required by several management/query commands.

**Requester Info**

Edit - turns out and `CFE_ES_GetGenCounterIDByName()` does exist already.  But the others do not.

jphickey:
**Describe the contribution**

Fixes #924 
Fixes #943 
Fixes #502

Provides a consistent Name-ID translation API for all resource types.  Enforces consistent argument/name validation on all resource types, and also enforces name uniqueness where relevant.

Also includes an enhancement to use the full 16 bits of resource ID space, which avoids repeating ID values that have already been used.  This significantly decreases the likelihood that a previously deleted ID will alias a newly allocated/valid ID.

**Testing performed**
Build and sanity test CFE
Ensure all unit tests pass
ALSO - run a continuous loop of "Restart" commands, to continually restart the SAMPLE_APP.  Confirmed that after each restart, the app is assigned a new id, different than the old ID, and never aliases any other valid IDs.

**Expected behavior changes**
- Complete/consistent API to translate between names and IDs, for all resource types which have a name
- IDs are not re-issued after deletion, helping ensure safety as previously deleted IDs will not validate.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
These two fixes are combined into a single PR as they have dependencies (would be a merge conflict if pulled separately)


jphickey:
Also adding a commit that fixes #502.  
Again due to dependencies this has to be serialized, can't be a separate PR or merge conflicts will result.

CDKnightNASA (approved):

skliper:
CCB: 2020-10-14 - Approved, pending @acudmore review.

fsw/cfe-core/src/es/cfe_es_resource.c
```diff
@@ -0,0 +1,428 @@
...
+ * For complete API information, see prototype in header
+ */
+int32 CFE_ES_ResourceID_ToIndex_Internal(uint32 Serial, uint32 TableSize, uint32 *Idx)
+{
+    if (Serial > CFE_ES_RESOURCEID_MAX)
+    {
+        return CFE_ES_ERR_RESOURCEID_NOT_VALID;
+    }
+
+    *Idx = Serial % TableSize;
```
acudmore, line 57: The 'Serial' variable is checked, should we also check for a non NULL pointer? I assume the check is not needed because the function is internal.

jphickey: Hmm, that may be a valid point ... In this case this function is internal-only but it is performing the range check on behalf of the various type-specific (public) API calls - i.e. so the same test doesn't need to be repeated 5 times over.

But you are correct that the parent doesn't validate the Idx pointer either.  I can add this.

fsw/cfe-core/src/es/cfe_es_resource.c
```diff
@@ -0,0 +1,428 @@
...
+    ** Use the OS task ID to get the ES task record
+    */
+    TaskID = CFE_ES_ResourceID_FromOSAL(OS_TaskGetId());
+    TaskRecPtr = CFE_ES_LocateTaskRecordByID(TaskID);
+
+    /*
+     * Confirm that the entry is actually a match (this requires/assumes
+     * the global data is locked).
+     *
+     * If not a match, return NULL.
```
acudmore, line 377: Is there a case where the TaskRecordIsMatch will catch an error? If the TaskRecPtr from line 371 is NULL, then the result will be the same. If the lookup in CFE_ES_LocateTaskRecordByID returns a TaskRecPtr, will the CFE_ES_TaskRecordIsMatch always succeed?

fsw/cfe-core/src/es/cfe_es_resource.c
```diff
@@ -0,0 +1,428 @@
...
+        */
+        AppRecPtr = CFE_ES_LocateAppRecordByID(TaskRecPtr->AppId);
+
+        /*
+         * Confirm that the entry is actually a match (this requires/assumes
+         * the global data is locked).
+         *
+         * If not a match, return NULL.
+         */
+        if (!CFE_ES_AppRecordIsMatch(AppRecPtr, TaskRecPtr->AppId))
```
acudmore, line 415: Same here. Will this ever catch a problem? Maybe a corrupted table?

acudmore (approved):
Looks good. The only thought I had was perhaps the resource ID functions could be generalized.. The Get ID functions repeat for each object type in ES, but I don't know if it's worth another layer of ID abstraction at this point. (Generic ES ID -> Library ID, etc ). Regardless, it's better than it was before, and solves problems associated with deleting and re-creating a resource. I also like the error code improvements.

astrogeco:
@jphickey can we resolve Alan's comments or do you need to make some tweaks?

fsw/cfe-core/src/es/cfe_es_resource.c
jphickey: Yes -  The "Locate" function just finds the entry in the table where that ID is _supposed_ to be.  It does not check validity of the entry.  This is by design, it is used for all operations - creation, deletion, lookup, etc. and records may or may not be initialized.  The "Locate" function returns NULL only if the resource ID was a completely bad value/wrong type etc.

The "IsMatch" function is the part that actually confirms/checks if the table pointer is actually correct for the intended ID.  It will check for table entries that are uninitialized or may be in use but for a different/non-matching ID - it returns `false` in that case.

So yes - the "Locate" step can succeed but the "IsMatch" can then fail.

Also note that "IsMatch" also needs to be called when locked, but the "Locate" doesn't need a lock, as it doesn't actually validate the entry.

fsw/cfe-core/src/es/cfe_es_resource.c
jphickey: Yep - same as other spot - "Locate" just finds where it is _supposed_ to be, "IsMatch" actually checks it.

jphickey:
> @jphickey can we resolve Alan's comments or do you need to make some tweaks?

@astrogeco The update in commit 4356693 should be all we need.

fsw/cfe-core/src/es/cfe_es_resource.c
jphickey: Update to check Idx in commit 4356693

jphickey:
> Looks good. The only thought I had was perhaps the resource ID functions could be generalized.. The Get ID functions repeat for each object type in ES, but I don't know if it's worth another layer of ID abstraction at this point. 

Yes, I concur there is some repetition here.  I had considered splitting the structures similar to OSAL - where the name/ID info is in one part (common to all) and the type-specific info is in a separate entry.

But - This would have been an even bigger change to the codebase, and the truth is that most of the time the user intends to operate on a specific resource type.  So even if we did give e.g. a "generic" name lookup API, the user would probably have to follow it up with a type check in many cases.  So I decided it probably wasn't worth it.

acudmore:
> > Looks good. The only thought I had was perhaps the resource ID functions could be generalized.. The Get ID functions repeat for each object type in ES, but I don't know if it's worth another layer of ID abstraction at this point.
> 
> Yes, I concur there is some repetition here. I had considered splitting the structures similar to OSAL - where the name/ID info is in one part (common to all) and the type-specific info is in a separate entry.
> 
> But - This would have been an even bigger change to the codebase, and the truth is that most of the time the user intends to operate on a specific resource type. So even if we did give e.g. a "generic" name lookup API, the user would probably have to follow it up with a type check in many cases. So I decided it probably wasn't worth it.

Another thought I had, but probably for a future round of updates.. your comments mentioned taking advantage of the OSAL resource ID structures while they are compatible, but no guarantee that it will stay that way. Maybe on a future update, the OSAL could provide a resource ID API to create its own resource IDs and allow cFE services such as ES to create and use Resource IDs, unifying the functions around creating, lookup, etc.
But I'm happy to leave that for now, this is good to me.

jphickey:
Yeah, I think the correct answer longer-term is to decouple the OSAL task ID from the CFE task ID.  That way each subsystem can manage their own concept of an "ID" independently.  It's easy to do on the CFE side - just have to translate the ID, which is already being done as a placeholder/passthru.

It is only a pass through because code expects to be able to take the output of e.g. `CFE_ES_CreateChildTask()` and pass that to OSAL functions, and likewise take output of e.g. `OS_TaskGetId()` and pass that to CFE functions.  So to support this in a backward-compatible manner we are keeping all IDs as `uint32` and making this work as it historically has.  

The reason I added wrapper functions like `CFE_ES_GetTaskID()` is to start laying the groundwork to breaking this assumption.  Once this is established then CFE can allocate its own IDs independently of the OSAL IDs and they don't need to worry about keeping the bit fields compatible.

jphickey:
Update - writing the comment above made me realize that a bit of debug test logic was left in the final commit.

I had been specifically testing the case were OSAL task IDs and CFE task IDs were _not_ numerically equivalent in order to confirm that all the abstractions were working (which they are) but this was still left in the final commit.

See Commit 0790e11 which puts this back to its backward-compatible form.   Please include this in the IC.

astrogeco:
> Update - writing the comment above made me realize that a bit of debug test logic was left in the final commit.
> 
> I had been specifically testing the case were OSAL task IDs and CFE task IDs were _not_ numerically equivalent in order to confirm that all the abstractions were working (which they are) but this was still left in the final commit.
> 
> See Commit [0790e11](https://github.com/nasa/cFE/commit/0790e118d8e6642cc4b8362e7d79cac2602ecdce) which puts this back to its backward-compatible form. Please include this in the IC.

One last check to make sure. Can I merge this now?

fsw/cfe-core/src/es/cfe_es_cds.c
```diff
@@ -378,43 +404,24 @@ int32 CFE_ES_RegisterCDSEx(CFE_ES_CDSHandle_t *HandlePtr, CFE_ES_CDS_Offset_t Us
...
-                /* On failure set it free */
-                CFE_ES_CDSBlockRecordSetFree(RegRecPtr);
-            }
+            /* Save CDS Name in Registry */
+            strncpy(RegRecPtr->Name, Name, sizeof(RegRecPtr->Name)-1);
+            RegRecPtr->Name[sizeof(RegRecPtr->Name)-1] = 0;
+            CFE_ES_CDSBlockRecordSetUsed(RegRecPtr, PendingBlockId);
         }
 
-        if (IsNewOffset || IsNewEntry)
```
thnkslprpt, line 399: Link https://github.com/nasa/cFE/issues/1929

Issue #139, opened 2019-09-30
TBL - Dump Table Registry Data Command Can Hog CPU (GSFC DCR 23031)


skliper:
The GSFC NICER FSW team discovered that it takes approximately 1 second to process the Dump Table Registry command when there are 250 RTS tables and that lower priority tasks generate various error events related to not getting an opportunity to run. Adding a platform configuration definition to insert task delays into the process loop will increase the time necessary for CFE_TBL to finish the command but will allow lower priority tasks to run while the command is being processed.

Also, creating some test code that puts performance markers around the file i/o calls within the command processor may identify where the bulk of the command process time is spent. It may be possible to reduce the command execution time by optimizing the command handler code if the bulk of the execution time is not being spent inside the file system.

skliper:
_Imported from trac issue 108.  Created by sstrege on 2015-08-28T16:09:54, last modified: 2019-03-01T15:27:58_

skliper:
_Trac comment by glimes on 2016-10-18 15:07:03:_

Remove my name from many tickets (where it was applied by default), so others can tell that there are tickets that need some attention.

skliper:
The file writing should likely be moved to a low priority child task, same for any other potentially "large" file writes: CFE_SB_SendRtgInfo, CFE_SB_SendPipeInfo, CFE_SB_SendMapInfo

jphickey:
These all sound like great candidates for moving into the ES background job service.

Make a generic file writer job template based on the current perf data dump, then apply it for all these file dump commands.

skliper:
Note - standby on this until #928 is implemented (for SB writes), since they are being refactored.

skliper:
Removed dependency label, this can be resolved now.

PR #947, opened 2020-10-13
Fix #928 and #929 - Modularize software bus routing, add msg map hash

### LINKED ISSUE #928: Move SB route lookup (including insert method) to a module
Various implementations possible for message and route tables, all with associated advantages and issues.

Provide the capability to replace the core implementation.

None.

**Additional context**
For implementations that support large MsgIds, or prefer smaller memory footprint at the cost of performance.  Hashes, searches, etc.

**Requester Info**
Jacob Hageman - NASA/GSFC (per stakeholder request for alternate implementation)

### LINKED ISSUE #929: Add message id to message key hash option
Core implementation is message id == message key, leading to a large, sparse message table to get the route index.

Hash message and deconflict on lookup (confirm route matches MsgId).  Could decrease message table size to a small multiple (2 or 4 maybe) of the routing table size depending on how many conflicts are acceptable.

Ordering the routing table and implement binary search and completely eliminate message table.

**Additional context**
See #928

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
**Describe the contribution**
Fix #928 - refactor SB and move routing to a module
Fix #929 - adds message map hash implementation
Partial #948 - fixed double lock potential

Key concepts:
 - added SBR module
 - Message map and routing table are in SBR, access APIs on SB side
 - SB still owns destination logic
 - no more route index or pointers being passed around, all route id and message id based
 - hash message map "oversized" (4x) to minimize collisions (only ~10% single collisions on the real system with a full routing table)
    - added debug event for collisions during add
 - dropped routing push/pop, dropped "key" in direct implementation
 - around a 10-20% performance hit to hash via rough comparison on a linux box, no memory impact from message id "space", just based on needed routes (with overhead)
 - hash designed for 32 bit, a change in CFE_SB_MsgId_Atom_t size may require implementation updates

Remaining work:
- Unit test stubs (could split into new PR), 2 related TODOs in code

Also 
- Deletes unused code CFE_SB_FindGlobalMsgIdCnt
- Note writes to file not protected by lock (before or after refactor)
- CFE_SB_SendPrevSubs isn't really protected by the lock... might be worth removing the locks completely (they don't really solve anything, double lock potential eliminated
  - could just use the private CFE_SBR_ForEachRouteId without a lock instead of messages and reduce all the extra logic
- Whitespace/alignment is very difficult to work with
- Excessive comments (in my opinion)
- violates declare variables at the start of function coding standard, all observed violations fixed

**Testing performed**
 - All unit tests pass, rough performance testing, spot checked routing table size impacts on core with nominal (256), 64, and 32.  32 causes dropped subscriptions/full table and was useful for observing collisions (~10% one level seen) on a realistic set of message ids.  Everything performed as expected. 
- Passed bundle CI at https://travis-ci.com/github/skliper/cFS/builds/191437587

**Expected behavior changes**
- Individual events for deleting destinations when deleting a pipe removed to avoid a race condition
- It's resource intensive to try to report message ID map in message ID order (either search or brute force) for any significantly sized map on an unsorted routing table, allowing out of order makes this reasonable again and gets rid of the need for ForEachMsgId API. Need to trade out of order vs implementing a search.

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: bundle + this commit

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_priv.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_priv.c
@@ -88,31 +88,6 @@
 #include <string.h>
 
 /******************************************************************************
-**  Function:  CFE_SB_InitIdxStack()
-**
-**  Purpose: Initialize a push/pop stack of routing table indexes.
-**           On init each must be unique. After system initialization SB_Idx_top
-**           will always point/index to the next available routing table index
-**
-**  Arguments:
-**
-**  Return:
-**    None
-*/
-
-void CFE_SB_InitIdxStack(void)
-{
-   uint16 i;
-
-   CFE_SB.RouteIdxTop = 0;
-   for (i=0; i<CFE_PLATFORM_SB_MAX_MSG_IDS; i++)
-    {
-       CFE_SB.RouteIdxStack[i] = CFE_SB_ValueToRouteIdx(i);
-    }
-}
-
-
-/******************************************************************************
 **  Function:  CFE_SB_CleanUpApp()
 **
 **  Purpose:
@@ -174,64 +149,6 @@ CFE_SB_PipeId_t CFE_SB_GetAvailPipeIdx(void){
 
 }/* end CFE_SB_GetAvailPipeIdx */
 
-/******************************************************************************
-**  Function:  CFE_SB_RouteIdxPop_Unsync()
-**
-**  Purpose:
-**    SB internal function to get the next available Routing Table element
-**    (CFE_SB_RouteEntry_t). Typically called when an application subscribes
-**    to a message.
-**
-**  Assumptions, External Events, and Notes:
-**      Calls to this function assumed to be protected by a semaphore
-**  Arguments:
-**      None
-**
-**  Return:
-**    Returns the index of an empty Routing Table element or
-**    CFE_SB_INVALID_ROUTE_IDX if there are no more elements available.
-*/
-CFE_SB_MsgRouteIdx_t CFE_SB_RouteIdxPop_Unsync (void) {
-
-    CFE_SB_MsgRouteIdx_t retValue;
-
-    /* This stack grows from 0 to (CFE_PLATFORM_SB_MAX_MSG_IDS - 1) */
-    if (CFE_SB.RouteIdxTop >= CFE_PLATFORM_SB_MAX_MSG_IDS) {
-        retValue = CFE_SB_INVALID_ROUTE_IDX; /* no more Idx remaining, all used */
-    } else {    
-        retValue = CFE_SB.RouteIdxStack[CFE_SB.RouteIdxTop];
-        ++CFE_SB.RouteIdxTop;
-    }
-
-    return (retValue);
-} /* end CFE_SB_IdxPop_Unsync */
-
-
-/******************************************************************************
-**  Function:  CFE_SB_RouteIdxPush_Unsync()
-**
-**  Purpose:
-**    SB internal function to return a Routing Table element to the available stack
-**    (CFE_SB_RouteEntry_t). Typically called when an application un-subscribes
-**    to a message. 0 is a valid idx.
-**
-**  Assumptions, External Events, and Notes:
-**      Calls to this function assumed to be protected by a semaphore
-** 
-**  Arguments:
-**    None
-**
-**  Return:
-**    None
-*/
-void CFE_SB_RouteIdxPush_Unsync (CFE_SB_MsgRouteIdx_t idx) {
-
-    /* This stack grows from 0 to (CFE_PLATFORM_SB_MAX_MSG_IDS - 1) */
-    if (CFE_SB.RouteIdxTop > 0) {
-        --CFE_SB.RouteIdxTop;
-        CFE_SB.RouteIdxStack[CFE_SB.RouteIdxTop] = idx;
-    }
-} /* end CFE_SB_IdxPush_Unsync */
 
 /******************************************************************************
 **  Function:  CFE_SB_GetPipeIdx()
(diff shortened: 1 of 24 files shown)


skliper:
Ready for proof of concept discussion...

skliper:
CCB 2020-10-14: Delayed discussion, will splinter or discuss next week.

skliper:
Worth noting this is a step in the direction of implementing just the direct lookup (no longer has the extra abstraction layers for other implementations).  The hash implementation will have the extra hash layer, but no need to maintain it in the direct lookup implementation.  This results in simplification of the direct lookup implementation.

fsw/cfe-core/src/sb/cfe_sbr.h
```diff
@@ -0,0 +1,236 @@
...
+ * \brief Call the supplied callback function for all routes
+ *
+ * For certain routing implementations this is faster than CFE_SBR_ForEachMsgId,
+ * but should only be used when MsgId order doesn't matter. Invokes callback
+ * for each route in the table.
+ *
+ * \param[in]  CallbackPtr Function to invoke for each matching ID
+ * \param[in]  ArgPtr      Opaque argument to pass to callback function
+ */
+void CFE_SBR_ForEachRouteId(CFE_SBR_CallbackPtr_t CallbackPtr, void *ArgPtr);
```
jphickey, line 185: These "for each" routines are somewhat similar to what is needed for #139 ... but in order to throttle we'll need some way to limit the number - i.e. only do a partial range - then resume where you left off on the next call(s).

fsw/cfe-core/src/sb/cfe_sbr.h
```diff
@@ -0,0 +1,236 @@
...
+ * \brief Set the destination list head pointer for given route id
+ *
+ * \param[in] RouteId Route Id
+ * \param[in] DestPtr Destination list head pointer
+ */
+void CFE_SBR_SetDestListHeadPtr(CFE_SBR_RouteId_t RouteId, CFE_SB_DestinationD_t *DestPtr);
+
+/* TODO add description */
+void CFE_SBR_IncrementSequenceCounter(CFE_SBR_RouteId_t RouteId);
+uint32 CFE_SBR_GetSequenceCounter(CFE_SBR_RouteId_t RouteId);
```
jphickey, line 161: Should we typedef the sequence count, for consistency?
(could argue either way, hard to fathom needing more than 32 bits, but still...)

fsw/cfe-core/src/sb/cfe_sb_priv.c
```diff
@@ -736,99 +515,104 @@ void CFE_SB_FinishSendEvent(CFE_ES_ResourceID_t TaskId, uint32 Bit){
 **  Return:
 **
 */
-int32 CFE_SB_AddDest(CFE_SB_RouteEntry_t *RouteEntry, CFE_SB_DestinationD_t *NewNode){
+int32 CFE_SB_AddDestNode(CFE_SBR_RouteId_t RouteId, CFE_SB_DestinationD_t *NewNode){
```
jphickey, line 518: I'd consider implementing these functions using a circular list structure rather than a linear structure.

This only requires an extra "head" node which does not store data. An "empty" list only has this head node.  In this state both prev and next point back to itself (not NULL).

The advantage is that in this type of data structure there is no NULL next/prev pointers, and therefore special conditions for first/last go away.  Remove/Insert ops become simple and constant time, with no conditionals.

Iterating the list means checking for arrival at the head node again (rather than NULL) which signifies the end of iteration.

IMO this type of structure keeps things simpler and more reliable, easier to read and maintain.

jphickey (commented):
Overall, looks pretty good ... nice cleanup/simplification to SB.

Most critical is really to get the API right, then the rest of the implementation can fall from that.

I'd also like to seriously consider changing the list structure to be circular in nature (where the end of the list points back to the beginning again, rather than NULL) as this makes the code much simpler.

fsw/cfe-core/src/sb/cfe_sb_priv.c
skliper: Good idea, but I'd prefer if that was a new issue/PR.  The only thing the routing does with destinations is set/get the head pointer, so it's all on the SB side (vs SBR).  I mucked minimally with naming and where I couldn't stop myself, but I am trying to limit my changes in this pr.

skliper:
Trade removing the individual unsubscribe events when deleting a pipe to avoid having to unlock which introduces a possible race (old implementation unlocked so could race).  Would be messy to store all the messages up to send at the end after the unlock, other suggestions welcome.

skliper:
Implementation/testing update - add a debug event for collisions (on subscribe), added hash implementation.

Enabled hash and there were no collisions with the nominal routing table size.  Reduced table size to 32 and there were only 4 single level collisions so the factor of 4 on the hash table seems very reasonable for handing a full routing table (and the hash performed well).  The messages past 32 (there's more like 40 or so in the framework) got rejected as expected.

Performing minor cleanup on the test cases (fails for size 32 routing table).  Some event counts could theoretically change due to the new debug message but none seen in practice.  Could switch subscribe test event count asserts to be >= to handle it, but as long as the hash function isn't changed and the routing table isn't incredibly small (16 or less) they will work fine.

fsw/cfe-core/src/sb/cfe_sbr.h
skliper: Now using the cfe message sequence counter typedef.

fsw/cfe-core/src/sb/cfe_sb_priv.c
skliper: See #953

fsw/cfe-core/src/sb/cfe_sbr.h
skliper: Added in fa8a216

skliper:
Brute force ForEachMsgId is unreasonable for any significantly large MsgId space.  Search would also be resource intensive for larger routing tables... only behavior difference from ForEachRouteId is order, propose removing ForEachMsgId (only used for message map reporting, could be sorted offline if needed).

excaliburtb:
tagging self

skliper:
Refactored commits for review.  Ready.

skliper:
Squashed cppcheck warning.

astrogeco:
[CCB 2020-10-21](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-October-21-2020) 

Conversation about collisions. System can be sized appropriately to avoid collisions. Users can deterministically predict the bound by appropriately choosing the MsgIDs.  There's a debug event upon subscription which alerts about collisions. 

Open issue to add documentation about ideal usage.

skliper:
Issues submitted, see #961 and #962.

modules/sbr/src/cfe_sbr_map_hash.c
```diff
@@ -0,0 +1,163 @@
...
+/** \brief Message map shared data */
+CFE_SBR_RouteId_t CFE_SBR_MSGMAP[CFE_SBR_MSG_MAP_SIZE];
+
+/******************************************************************************
+ * Internal helper function to hash the message id
+ *
+ * Note: algorithm designed for a 32 bit int, changing the size of
+ * CFE_SB_MsgId_Atom_t may require an update to this impelementation
+ */
+CFE_SB_MsgId_Atom_t CFE_SBR_MsgIdHash(CFE_SB_MsgId_t MsgId)
```
acudmore, line 81: Should an internal helper function be static?

acudmore (commented):
Code looks good.
I'm a little less clear about the organization. We have the fsw/cfe-core directory with the cFE subsystems, then we have the modules directory that has the msg, sbr, cfe_assert, cfe_testcase, and cfe_testrunner. 

In the mission_defaults.cmake we have cfe-core, osal, PSP, msg, and sbr. Seems like msg and sbr are peers to the cfe-core subsystems, (sb, es, tbl, eve, time) or maybe even subsystems of SB. I know mission_defaults.cmake does not define the architecture, but it leaves the impression to me that these modules are peers to osal, psp, etc. 

The routing module has a common file and APIs, then it has optional (hash or direct) modules. Is the intention to allow the replacement of everything with a different implementation? Or would the common files (in unsorted.c) always be used? If these common functions in unsorted will always be used, I would argue for leaving them in the SB directory and having modules for hash and direct routing.

In the end, I think we need a big reorganization of the code to reflect the evolving nature of how things go together. But at a minimum I think it would be nice at least put the sbr and msg modules in the "fsw" directory next to "cfe-core":
Then we would have:
fsw/
fsw/cfe-core
fsw/modules
fsw/modules/sbr
fsw/modules/msg

or maybe even fsw/cfe-core/modules/...

skliper:
> The routing module has a common file and APIs, then it has optional (hash or direct) modules. Is the intention to allow the replacement of everything with a different implementation?

Yes.  That's the real difference from the core, the module can be completely replaced with a custom implementation.

> Or would the common files (in unsorted.c) always be used?

Not if someone implements a sorted routing table with a search instead of a message map.

> In the end, I think we need a big reorganization of the code to reflect the evolving nature of how things go together.

Agreed! opened #972 to discuss.  I like the msg/sbr modules and core being peers, and eventually all the cfe elements could become modules (someone could entirely replace table services or time services for example).  It's been an as-needed process so far though.

modules/sbr/src/cfe_sbr_map_hash.c
skliper: Certainly could be.  I'm not consistently strict about static since they are occasionally useful from a coverage test point of view, but it's not necessary in this case.  Happy to change if you want to put in an issue.

PR #959, opened 2020-10-21
Integration Candidate: 2020-10-20

### LINKED ISSUE #502: CFE ES unnecessarily keeping multiple copies of task/app names

### LINKED ISSUE #924: Improve uniqueness in resource IDs

### LINKED ISSUE #943: Provide consistent name lookup API

astrogeco:
**Describe the contribution**
Fix #943 
Fix #924
Fix #502 

**Testing performed**
Bundle CI: https://github.com/nasa/cFS/pull/152/checks

**Expected behavior changes**
PR #946 

**System(s) tested on**
Ubuntu

**Additional context**
Part of <https://github.com/nasa/cFS/pull/152>

**Third party code**
None

@jphickey


astrogeco:
@jphickey we have a problem with the merge from #946, 

```
1980-012-14:03:20.00039 EVS:Call to CFE_ES_GetAppID Failed:RC=0xC4000001
1980-012-14:03:20.00041 EVS:Application Init Failed,RC=0xC4000001
```

jphickey:
Fixed - this was some sort of merge/rebase issue, the line that sets the ID got moved but the refs before it were using the wrong (reserved) value not the correct (pending) value.

PR #960, opened 2020-10-21
Fix #28, provide library API

### LINKED ISSUE #28: API for querying libraries (like CFE ES QueryApp)
Unable to verify library image integrity (CS cannot scrub library code space)

Implement API for querying library information, such that an app could scrub

None

**Additional context**
None

**Requester Info**
Jacob Hageman/NASA-GSFC
External request from cfs-community email list on 9/3/2019

### LINKED ISSUE #173: Refactor CFE_ES_AppCreate and CFE_ES_LoadLibrary
These functions actually perform 3 major operations:

* Calling `OS_ModuleLoad` to load the module

* Calling `OS_SymbolLookup` to find the entry point / init function and calling that entry point / init function.  (For apps this also involves `OS_TaskCreate` here whereas a library init function is called directly)

* Manipulate the internal CFE ES Global tables to store the data related to the application/library

These major functional items should be broken into separate functions.  This will make `CFE_ES_AppCreate` and `CFE_ES_LoadLibrary` much cleaner looking and remove the overloads needed to support static loading as in ticket #115.

### LINKED ISSUE #950: Race condition in control requests
**Describe the bug**
Due to the order of operations in clean up, the ES global lock is given up and then re-acquired:

https://github.com/nasa/cFE/blob/dc3d62b37d2bacd8fe23f66ff7b64025ac7d0751/fsw/cfe-core/src/es/cfe_es_apps.c#L859-L861

The problem is that this provides a window of opportunity for the underlying state to change externally while the global data is unlocked.

**To Reproduce**
This can happen, for instance, if the task that is being cleaned up calls `CFE_ES_ExitApp()` while this state machine is also cleaning up the app.  
This actually does happen because `CFE_ES_RunLoop()` will return false if there is an exit request pending.  It is just masked by the fact that most apps are pending in a message receive queue, so they don't self exit - they are deleted by ES instead.

I was able to get CFE to segfault/crash by allowing SAMPLE_APP to exit itself at the very same time that this state machine was also cleaning it up.

**Expected behavior**
No crashes, proper clean up.

**System observed on:**
Ubuntu 20.04

**Additional context**
Due to the ~5 second exit/cleanup delay it is unlikely to occur "in the wild" but it can easily be forced to happen.  In my test I just used a slightly modified `sample_app` that doesn't pend forever on `CFE_SB_RcvMsg`, and also delays itself such that it self-exits at the exact same time that the ES background job is running, which reliably segfaults every time.

**Reporter Info**

jphickey:
**Describe the contribution**

Fixes #28 
Provide Library API similar to App API

Allows the existing `CFE_ES_AppInfo_t` structure to be extended to libraries as well as applications by introducing a new value (3)
for the Type field.

Allows Libraries to be queried via API calls similar to App API.

Also extends the Query All/Query One commands to operate on Libraries or Applications.

Fixes #173 
Break up the monolithic AppCreate and LoadLibrary functions and have these call subroutines that operate on the common components.

Fixes #950 
Fix race conditions in app request processing state machine.

**Testing performed**
Build and sanity check operation of CFE
Confirm able to restart/reload sample_app via command as expected
Run all unit tests

**Expected behavior changes**
Adds Library API.
Query commands that previously worked only on Apps can now be used on Apps or Libs

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_apps.c
+++ b/fsw/cfe-core/src/es/cfe_es_apps.c
@@ -344,6 +344,236 @@ int32 CFE_ES_ParseFileEntry(const char **TokenList, uint32 NumTokens)
 }
 
 /*
+**-------------------------------------------------------------------------------------
+** Name: CFE_ES_LoadModule
+**
+** Helper function to load + configure (but not start) a new app/lib module
+**
+** Loads the module file via OSAL and stores all relevant info in the table entry as necessary.
+**
+**-------------------------------------------------------------------------------------
+*/
+int32 CFE_ES_LoadModule(const CFE_ES_ModuleLoadParams_t* LoadParams, CFE_ES_ModuleLoadStatus_t *LoadStatus)
+{
+    osal_id_t ModuleId;
+    cpuaddr StartAddr;
+    int32 ReturnCode;
+    int32 StatusCode;
+
+    StartAddr = 0;
+    ReturnCode = CFE_SUCCESS;
+
+    if (LoadParams->FileName[0] != 0)
+    {
+        /*
+         ** Load the module via OSAL.
+         */
+        StatusCode = OS_ModuleLoad ( &ModuleId,
+                LoadParams->Name,
+                LoadParams->FileName );
+
+        if (StatusCode != OS_SUCCESS)
+        {
+            CFE_ES_WriteToSysLog("ES Startup: Could not load file:%s. EC = 0x%08X\n",
+                    LoadParams->FileName, (unsigned int)StatusCode);
+            ModuleId = OS_OBJECT_ID_UNDEFINED;
+            ReturnCode = CFE_STATUS_EXTERNAL_RESOURCE_FAIL;
+        }
+    }
+    else
+    {
+        ModuleId = OS_OBJECT_ID_UNDEFINED;
+    }
+
+    /*
+     ** If the Load was OK, then lookup the address of the entry point
+     */
+    if (ReturnCode == CFE_SUCCESS && LoadParams->EntryPoint[0] != 0)
+    {
+        StatusCode = OS_SymbolLookup(&StartAddr, LoadParams->EntryPoint);
+        if (StatusCode != OS_SUCCESS)
+        {
+            CFE_ES_WriteToSysLog("ES Startup: Could not find symbol:%s. EC = 0x%08X\n",
+                    LoadParams->EntryPoint, (unsigned int)StatusCode);
+            ReturnCode = CFE_STATUS_EXTERNAL_RESOURCE_FAIL;
+        }
+    }
+
+    if ( ReturnCode == CFE_SUCCESS )
+    {
+        /* store the data in the app record after successful load+lookup */
+        LoadStatus->ModuleId = ModuleId;
+        LoadStatus->EntryAddress = StartAddr;
+    }
+    else if (OS_ObjectIdDefined(ModuleId))
+    {
+        /* If the module had been successfully loaded, then unload it,
+         * so that it does not consume resources */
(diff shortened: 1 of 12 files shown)


astrogeco:
[CCB 2020-10-21](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-October-21-2020) APPROVED

Undoes a change from about a month ago and does some refactoring. Updated documentation to match.

skliper:
@ejtimmon This will allow CS to checksum libraries

jphickey:
Update - no changes, just rebased to current `main` branch.

fsw/cfe-core/src/es/cfe_es_apps.c
```diff
@@ -343,6 +343,236 @@ int32 CFE_ES_ParseFileEntry(const char **TokenList, uint32 NumTokens)
...
+    if (CFE_ES_GetAppEntryPoint(&RealEntryFunc) == CFE_SUCCESS &&
+            RealEntryFunc != NULL)
+    {
+        (*RealEntryFunc)();
+    }
+}
+
+/*
+**-------------------------------------------------------------------------------------
+** Name: CFE_ES_StartMainTask
```
acudmore, line 496: Name: CFE_ES_StartAppTask

fsw/cfe-core/src/es/cfe_es_apps.c
```diff
@@ -343,6 +343,236 @@ int32 CFE_ES_ParseFileEntry(const char **TokenList, uint32 NumTokens)
...
+** Name: CFE_ES_GetAppEntryPoint
+**
+** Helper function to act as the intermediate entry point of an app
+** This is to support starting apps before having a fully completed entry in the
+** global app table.  The app startup will delay until the app creation is completed
+** and verified, then the actual entry point will be determined.
+**
+**-------------------------------------------------------------------------------------
+*/
+int32 CFE_ES_GetAppEntryPoint(osal_task_entry *FuncPtr)
```
acudmore, line 434: Could this logic go in CFE_ES_RegisterApp to avoid having the entry point redirection?
What happens if the entry point cannot be found? Since this is running in the context of the new task, will it just end since it does not call the actual app entry point?
If so, should there be an else clause in CFE_ES_AppEntryPoint to clean up anything?

acudmore (approved):
I like the clean up and continued modularization, and I think the ability to treat loadable libraries like apps is important.

I think the app startup should be fine as is, but I would prefer if the app handled this in CFE_ES_RegisterApp if possible. Not sure if the context exists to do this from CFE_ES_RegisterApp though.

fsw/cfe-core/src/es/cfe_es_apps.c
jphickey: Unfortunately that gives a chicken-egg situation.  `CFE_ES_RegisterApp()` is invoked by the app's init code, which is inside the entry point.  So this can't be used to find the entry point.

The reason for this is to impose a delay until the global table entries (app + task) are fully set up.  It just wants to see a coherent pair - a task entry which matches the running task context, which in turn refers to an app entry which refers back to the same task in its MainTaskID field.  Only once this positive-match criteria is all met will we "believe" the entry point field and call it.

Truth is - we should probably put the entry point on the task table, not the app table.  It would make this simpler because it only has to check coherency of the task table entry - although it couldn't be removed entirely, as the same fundamental issue would still exist. (i.e. can't set TaskID in global table until you know the ID, and you don't know the ID until the task is already running).

jphickey: And to answer your question - if the criteria is never met and an entry point is not known, the task will exit without starting anything.  If that ever happens though, it means there is a bug somewhere.

Issue #777, opened 2020-07-14
Deprecate SB elements relative to msg module adaption


skliper:
**Is your feature request related to a problem? Please describe.**
Many APIs will be updated for consistency relative to the MSG module, also element scoping improvements (SB doesn't actually care about the header, it just needs to route).

**Describe the solution you'd like**
Per 2020-07-28 discussion SB for once #726 is in.

Deprecating:
- [x] CFE_SB_PKTTYPE_* -> CFE_MSG_Type_t
- [x] CFE_SB_MsgPtr_t -> CFE_MSG_Message_t *
- [x] CFE_SB_Msg_t -> CFE_MSG_Message_t
- [x] CFE_SB_MsgPayloadPtr_t (use pointer to payload in structure)
- [x] CFE_SB_InitMsg -> CFE_MSG_Init
- [x] CFE_SB_GetTotalMsgLength -> CFE_MSG_GetSize
- [x] CFE_SB_SetTotalMsgLength -> CFE_MSG_SetSize
- [x] CFE_SB_GetMsgTime -> CFE_MSG_GetMsgTime (this gets rid of structure return, similar to #45 issue)
- [x] CFE_SB_SetMsgTime -> CFE_MSG_SetMsgTime
- [x] CFE_SB_GetCmdCode -> CFE_MSG_GetFcnCode
- [x] CFE_SB_SetCmdCode -> CFE_MSG_SetFcnCode
- [x] CFE_SB_GetChecksum (no use case defined, what do you need it for?)
- [x] CFE_SB_GenerateChecksum -> CFE_MSG_GenerateChecksum
- [x] CFE_SB_ValidateChecksum -> CFE_MSG_ValidateChecksum
- [x] CFE_SB_GetMsgId -> CFE_MSG_GetMsgId
- [x] CFE_SB_SetMsgId -> CFE_MSG_SetMsgId
- [x] CFE_SB_GetPktType -> CFE_MSG_GetTypeFromMsgId
- [x] CFE_SB_SetMsgSeqCnt -> CFE_MSG_SetSequenceCount

NOT deprecating, but will note in API that these are fragile (guesses based on assumptions).  Future implementation could be replaced via data dictionary sort of access or standards based header definitions (including secondary header, and flags or extra internal data to mange the real sizes):
- CFE_SB_MsgHdrSize: use actual message structure where possible
- CFE_SB_GetUserData: use actual message structure where possible
- CFE_SB_GetUserDataLength: use actual message structure where possible
[...]

Issue #962, opened 2020-10-21
Add stubs for software bus routing module APIs and update SB unit tests


skliper:
**Is your feature request related to a problem? Please describe.**
Stubs not implemented yet, directly linking module to SB for first stage of testing.

**Describe the solution you'd like**
Add stubs, update SB to use them

**Describe alternatives you've considered**
None.

**Additional context**
#947

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
Closed to #1462 as duplicate.

PR #1015, opened 2020-11-13
Fix# 1009, Apply consistent alignment pattern

### LINKED ISSUE #1009: Refactor message header alignment and "raw" types to fit a clear pattern
Aligned version of message headers currently in SB, shows different handling of the base type.
https://github.com/nasa/cFE/blob/9804b59eb34154b10bb2946bde7654f3d27f4b85/fsw/cfe-core/src/inc/cfe_sb.h#L150-L163

See discussion below.

None.

**Additional context**
Brought up as part of #777/#998 review.

**Requester Info**
Jacob Hageman - NASA/GSFC (from CCB discussion)

skliper:
**Describe the contribution**
Fix #1009 
Partially addresses #1019 (adds CFE_SB_TransmitMsg, CFE_SB_TransmitBuffer, CFE_SB_ReceiveBuffer)

 - Main change is to utilize CFE_SB_Buffer_t and CFE_MSG_Message_t in a consistent manner to facilitate alignment
 - Deprecates CFE_SB_SendMsg, CFE_SB_PassMsg, CFE_SB_RcvMsg, CFE_SB_ZeroCopyPass, CFE_SB_ZeroCopySend
 - Deprecates CFE_SB_TlmHdr_t, CFE_SB_CmdHdr_t
 - Deprecates CFE_SB_CMD_HDR_SIZE and CFE_SB_TLM_HDR_SIZE
 - Redefines CFE_MSG_Size_t as size_t to minimize duplicated work and facilitate transition to just size_t

See also details in the individual commits.

Why is this necessary?
  - The former implementation wasn't clear on the use of CFE_SB_Msg_t vs CFE_MSG_Message_t, CFE_SB_TlmHdr_t vs CFE_MSG_TelemetryHeader_t, CFE_SB_CmdHdr_t vs CFE_SB_CommandHeader_t
  - Worst case alignment was enforce at the message level, making it impossible to use the message types in cmds/tlm without impacting the sizes of some of the cmds/tlm (they would get rounded up)
  - Still couldn't cast to a command type that contained anything that required more than 32 bit alignment

Now
  - CFE_SB_Buffer_t is aligned for up to a long double, so now for command processing cast alignment warnings are all resolved
  - Clear use of CFE_SB_Buffer_t and removal of duplicated/confusing terms
  - CFE_MSG_Message_t no longer requires any "extra" alignment and is available in the CFE_MSG_TelemetryHeader_t and CFE_MSG_CommandHeader_t structures so no cast is required to use the MSG APIs (just pass in the msg)
  - CFE_MSG_TelemetryHeader_t and CFE_MSG_CommandHeader_t can now be used in the definition of all cmd/tlm structures and avoid casts (no more uint8 header of size *_HDR_SIZE)

**Testing performed**
Bundle passed CI, unit tests pass.

**Expected behavior changes**
None, pattern change.

**System(s) tested on**
 - Hardware: cFS Dev server
 - OS: Ubuntu 18.04
 - Versions: Bundle main + #998, although #1009 is the subject of this PR

**Additional context**
#777, #998 

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/docs/cFE Application Developers Guide.md
+++ b/docs/cFE Application Developers Guide.md
@@ -1080,21 +1080,17 @@ void SAMPLE_TaskMain(void)
         /*
         ** Wait for the next Software Bus message
         */
-        Status = CFE_SB_RcvMsg( &SAMPLE_TaskData.MsgPtr,
-                                SAMPLE_TaskData.CmdPipe,
-                                CFE_SB_PEND_FOREVER );
+        Status = CFE_SB_ReceiveBuffer(&SBBufPtr, SAMPLE_TaskData.CmdPipe, CFE_SB_PEND_FOREVER);
 
         if (Status == CFE_SUCCESS)
         {
             /*
             ** Process Software Bus message
             */
-            SAMPLE_TaskPipe(CFE_TBL_TaskData.MsgPtr);
+            SAMPLE_TaskPipe(SBBufPtr);
 
             /* Update Critical Data Store */
-  	        CFE_ES_CopyToCDS(SAMPLE_TaskData.MyCDSHandle,
-                             &SAMPLE_MyCDSDataType_t);
-
+            CFE_ES_CopyToCDS(SAMPLE_TaskData.MyCDSHandle, &SAMPLE_MyCDSDataType_t);
         }
     }
     CFE_EVS_SendEvent(CFE_TBL_EXIT_ERR_EID, CFE_EVS_ERROR,
@@ -1216,7 +1212,7 @@ FILE: xx_app.c
 void XX_AppMain(void)
 {
     uint32 RunStatus = CFE_ES_RunStatus_APP_RUN;
-    CFE_MSG_Message_t *MsgPtr;
+    CFE_SB_Buffer_t *SBBufPtr;
     int32  Result = CFE_SUCCESS;
 
     /* Register application */
@@ -1250,7 +1246,7 @@ void XX_AppMain(void)
         CFE_ES_PerfLogExit(XX_APPMAIN_PERF_ID);
 
         /* Wait for the next Software Bus message */
-        Result = CFE_SB_RcvMsg(&MsgPtr, XX_GlobalData.CmdPipe, CFE_SB_PEND_FOREVER);
+        Result = CFE_SB_ReceiveBuffer(&SBBufPtr, XX_GlobalData.CmdPipe, CFE_SB_PEND_FOREVER);
 
         /* Performance Log (start time counter) */
         CFE_ES_PerfLogEntry(XX_APPMAIN_PERF_ID);
@@ -1258,7 +1254,7 @@ void XX_AppMain(void)
         if (Result == CFE_SUCCESS)
         {
             /* Process Software Bus message */
-            XX_ProcessPkt(MsgPtr);
+            XX_ProcessPkt(SBBufPtr);
         }
         else
         {
@@ -1612,7 +1608,7 @@ FILE: sample_app.h
 */
 typedef struct
 {
-  uint8          TlmHeader[CFE_SB_TLM_HDR_SIZE];
+  CFE_MSG_TelemetryHeader_t TlmHeader;
 
   /*
   ** Task command interface counters...
@@ -1646,17 +1642,17 @@ SAMPLE_AppData_t  SAMPLE_AppData;  /* Instantiate Task Data */
    int32 Status;
 
    ...
-   Status = CFE_MSG_Init(&SAMPLE_AppData.HkPacket,    /* Address of SB Message Data Buffer */
-                          SAMPLE_HK_TLM_MID,          /* SB Message ID associated with Data */
-                          sizeof(SAMPLE_HkPacket_t)); /* Size of Buffer */
+   Status = CFE_MSG_Init(&SAMPLE_AppData.HkPacket.TlmHeader.Msg, /* Address of SB Message Data Buffer */
+                          SAMPLE_HK_TLM_MID,                     /* SB Message ID associated with Data */
+                          sizeof(SAMPLE_AppData.HkPacket));      /* Size of Buffer */
    ...
 }
 ```
 
 In this example, the Developer has allocated space for the SB Message
-header in their structure using the CFE_SB_TLM_HDR_SIZE macro. If
+header in their structure using the CFE_MSG_TelemetryHeader_t type. If
 the SB Message was to be a command message, it would have been important
-for the Developer to have used the CFE_SB_CMD_HDR_SIZE macro
+for the Developer to have used the CFE_MSG_CommandHeader_t macro
 instead.
 
 The CFE_MSG_Init API call formats the Message Header
@@ -1671,38 +1667,44 @@ of the Application's interface. This makes it much simpler to port the
 Application to another mission where SB Message IDs may need to be
 renumbered.
 
-##### 6.5.1 Software Bus Message Header Types
+##### 6.5.1 Message Header Types
 
-The SB support two main types of headers: command headers and telemetry
-headers.  In the CCSDS implementation of the SB, the command and telemetry
+The Message module supports two main types of headers: command headers and telemetry
+headers.  In the CCSDS implementation, the command and telemetry
 headers share the same CCSDS primary header structure but have different
 secondary header structures. The secondary header structures are
-shown below.  Note that all SB messages must have a secondary header.
+shown below.  Note that all messages must have a secondary header, a message
+containing just a CFE_MSG_Message_t is invalid per the CCSDS standard.
 
 ```
-typedef struct {
+typedef struct
+{
 
-   uint16  Command;      /* command secondary header */
-      /*  bits  shift   ------------ description ---------------- */
-      /* 0x00FF    0  : checksum, calculated by ground system     */
-      /* 0x7F00    8  : command function code                     */
-      /* 0x8000   15  : reserved, set to 0                        */
+    uint8 FunctionCode; /**< \brief Command Function Code */
+                        /* bits shift ---------description-------- */
+                        /* 0x7F  0    Command function code        */
+                        /* 0x80  7    Reserved                     */
 
-} CCSDS_CmdSecHdr_t;
+    uint8 Checksum; /**< \brief Command checksum  (all bits, 0xFF)      */
 
-typedef struct {
+} CFE_MSG_CommandSecondaryHeader_t;
 
-   uint8  Time[CCSDS_TIME_SIZE];
+/**
+ * \brief cFS telemetry secondary header
+ */
+typedef struct
+{
 
-} CCSDS_TlmSecHdr_t;
+    uint8 Time[6]; /**< \brief Time, big endian: 4 byte seconds, 2 byte subseconds */
 
+} CFE_MSG_TelemetrySecondaryHeader_t;
 ```
 
-The sizes of command and telemetry message headers are defined by  
-CFE_SB_CMD_HDR_SIZE and CFE_SB_TLM_HDR_SIZE respectively.  
+The sizes of command and telemetry message headers can be calculated using
+sizeof(CFE_MSG_CommandHeader_t) and sizeof(CFE_MSG_TelemetryHeader_t) respectively.
 
-It is important to note that some SB API calls assume the presence of a
-particular header type and will not work properly if the other header type
+It is important to note that some API calls require the presence of a
+particular header type and will return an error if the other header type
 is present instead.  The following section provides more detail.  
 
 ##### 6.5.2 Setting Message Header Information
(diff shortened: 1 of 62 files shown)


skliper:
How about renaming CFE_SB_Msg_t to CFE_SB_Buffer_t (since it's really a buffer that contains a message), and SendMsg could just take the "raw" message since it does a copy (no need to union the buffer on a normal send).  The only place that needs to locally handle the buffer is for zero copy, and when receiving the message (update to ES coming that does the receive more appropriately).

skliper:
6a7e6c3 - implements the latest strawman for ES
 - CFE_SB_SendMsg uses CFE_MSG_Message_t (apps no longer need alignment union)
 - Didn't convert MemStatsPacket or OneAppPacket to use zero copy, but these would be ideal candidates

docs/src/cfe_sb.dox
```diff
@@ -508,31 +508,29 @@
...
         </B><TR><TD WIDTH="5%"> &nbsp; <TD WIDTH="95%">
            There is a single copy of the message performed during a typical delivery.
-           During the #CFE_SB_SendMsg API, the software bus copies the message from the
-           callers memory space to the software bus memory space. The #CFE_SB_RcvMsg API
-           gives the user a pointer to the message in the software bus memory space. This
+           When transmitting a message, the software bus copies the message from the
+           callers memory space into a buffer in the software bus memory space.
+           The #CFE_SB_ReceiveBuffer API gives the user back a pointer to the buffer. This
            is equivalent to the copy mode send and pointer mode receive in the heritage
            software bus used on WMAP, ST5, SDO etc.
```
ejtimmon, line 515: Suggest updating the last sentence of this answer to compare the new approach to previous versions of cFE instead of "WMAP, ST5, SDO, etc." -> I think more users would be familiar with older versions of cFE than with those specific missions.

docs/src/cfe_sb.dox
```diff
@@ -292,20 +292,21 @@
...
-  value of zero for one instance. The sequence counter is incremented in the #CFE_SB_SendMsg
-  API after all the checks have passed prior to the actual sending of the message. This
-  includes the parameter checks and the memory allocation check.
-  Note: The count is incremented regardless of whether there are any subscribers.
-
-  For telemetry messages sent with the #CFE_SB_PassMsg API the sequence counter is not incremented.
-  This method of message delivery is recommended for situations
-  where the sender did not generate the packet, such as a network interface application
-  passing a packet from a remote system to the local software bus.
+  For a telemetry message, the behavior is controlled via input parameters or API selection
```
ejtimmon, line 295: What "API selection" is used for this?

skliper: It used to be via CFE_SB_SendMsg and CFE_SB_PassMsg.  CFE_SB_TransmitMsg just has an input parameter to control it.

acudmore (approved):
I like the API simplification, and it solves the alignment issue, which is essential.

Review Notes:
1. I'm not crazy about using TransmitMsg instead of SendMsg - But I understand that a different API is necessary to keep the deprecated APIs in the build.

2. I had a hard time with the naming of the primary header fields in CFE_MSG_CommandHeader_t and CFE_MSG_TelemetryHeader_t. When reading some of the code/structures it looks like the "message" is in the header. Like this:
 CFE_SB_TransmitMsg(&CFE_EVS_GlobalData.EVS_TlmPkt.TlmHeader.Msg, true);
The more code I reviewed, the more I got used to the pattern, but I think it will be confusing new users especially.
I would prefer a name to indicate it is a primary header or message header, but then it might make other uses of it when calling APIs confusing. I don't have a better suggestion right now.

3. Related to item 2, when a user calls TransmitMsg, the pointer to the message header is passed in, it seems counter intuitive to not pass a pointer to an entire message structure that contains the header, secondary header, and payload. I can see a mistake where a user will declare the primary message header and pass only that to TransmitMsg. But again, the cast issue is solved by doing this.
So I think my issue is with the overload of the term "message" which I always took to mean the entire message (header, sec header, payload) and not just the primary header.

4. I was originally confused with the use of CFE_SB_Buffer_t vs CFE_MSG_Message_t, but I think I understand the difference now. CFE_SB_Buffer_t is the aligned wrapper for CFE_MSG_Message_t. 

Overall, the technical implementation is fine, and there is a lot of clean up that looks good ( size_t, etc) The Zero copy is cleaner. I think we will have to take extra time in documentation and training to make sure that users know the distinctions between all of the these structures and APIs are used:
- TransmitMsg vs TransmitBuffer
- Use of CFE_SB_Buffer_t vs CFE_MSG_Message_t
- What part of the message to pass to TransmitMessage, etc.
New users will definitely have to start with a sample app to get things right.
( note, that this was probably always like this with many of our cFE APIs, but that was my initial impression with looking at the new APIs )

astrogeco:
[CCB 2020-12-02](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-02-2020) APPROVED

docs/src/cfe_sb.dox
skliper: Updated, see 76a0951.

docs/src/cfe_sb.dox
skliper: Updated, see 76a0951

fsw/cfe-core/src/inc/cfe_sb.h
```diff
@@ -513,7 +514,40 @@ CFE_Status_t CFE_SB_UnsubscribeLocal(CFE_SB_MsgId_t MsgId, CFE_SB_PipeId_t PipeI
...
+**                                   sequence count and update the message if the
+**                                   buffer contains a telemetry message
+**
+** \return Execution status, see \ref CFEReturnCodes
+** \retval #CFE_SUCCESS         \copybrief CFE_SUCCESS
+** \retval #CFE_SB_BAD_ARGUMENT \copybrief CFE_SB_BAD_ARGUMENT
+** \retval #CFE_SB_MSG_TOO_BIG  \copybrief CFE_SB_MSG_TOO_BIG
+** \retval #CFE_SB_BUF_ALOC_ERR \copybrief CFE_SB_BUF_ALOC_ERR
+**/
+CFE_Status_t  CFE_SB_TransmitMsg(CFE_MSG_Message_t *MsgPtr, bool IncrementSequenceCount);
```
jphickey, line 544: I'm slightly concerned about the extra boolean on this API.  I would prefer something that can be a drop-in replacement for `CFE_SB_SendMsg`.  Given that probably 99% of users will be passing `true` for the sequence count, and passing `false` is probably only a special use case that only SBN needs.  It might be confusing to users why this extra `true` argument needs to be passed and open up the possibility that they make mistakes in doing so, then wonder why their sequence numbers are wrong.

For consideration - can we make the "increment sequence count" to be a property of the _buffer itself_ instead of the API call?  That is, the buffer indicates if it is a truly a "new" message or if it is simply re-encapsulating a message that originated elsewhere - and therefore doesn't get a new sequence#.

This means that `CFE_SB_TransmitMessage` would always increment sequence (no extra boolean parameter)

For the SBN use case, it could request a pass-through buffer and use the `CFE_SB_TransmitBuffer` to pass a message without incrementing the sequence count.

To put it simply - I think we should strive to keep the common use case (TransmitMsg) to be as simple of an API as possible, minimizing the number of ways they can go wrong.

Users who need something more complex or involving special treatment should use the buffer-based/zero copy pattern (i.e. get a buffer, indicate that the _buffer_ represents a pass-through message, and then send the buffer).

Users who are simply sending a new message from their app (most common use case) should get the simplest API which does exactly what they want and nothing else.

fsw/cfe-core/src/inc/cfe_sb.h
```diff
@@ -569,10 +601,9 @@ CFE_Status_t  CFE_SB_SendMsg(CFE_MSG_Message_t *MsgPtr);
 ** \retval #CFE_SB_BAD_ARGUMENT \copybrief CFE_SB_BAD_ARGUMENT
 ** \retval #CFE_SB_MSG_TOO_BIG  \copybrief CFE_SB_MSG_TOO_BIG
 ** \retval #CFE_SB_BUF_ALOC_ERR \copybrief CFE_SB_BUF_ALOC_ERR
-**
-** \sa #CFE_SB_RcvMsg, #CFE_SB_ZeroCopySend, #CFE_SB_SendMsg
 **/
 CFE_Status_t  CFE_SB_PassMsg(CFE_MSG_Message_t *MsgPtr);
+#endif /* CFE_OMIT_DEPRECATED_6_8 */
```
jphickey, line 606: In line with my previous comment - it might be worth _not_ deprecating CFE_SB_PassMsg just yet.  I agree it should go way but I think it should be melded into the buffer-based/zero-copy approach rather than the typical transmit message approach - which should be kept as simple and straightforward as possible.

jphickey (changes requested):
This looks good!

As noted in my comments, the only thing that concerns me is the fact that we are making one of the most commonly used functions in the system - `CFE_SB_SendMsg()` aka `CFE_SB_TransmitMessage()` to be more complicated than it was before, forcing that complexity on all users.  (In an ideal world I'd also push to to even keep the name the same, to avoid breaking their code, but since the type change is necessary it's kinda moot, code would break either way).

Most users won't know why they need to pass an extra `true` on the new API.  For a typical user that has no knowledge of SBN - and works on a single SB domain - the extra parameter make no sense and there is no reason for typical users to ever pass `false` here.

So I'd like to see the use case for `CFE_SB_TransmitMessage()` be exactly the same as `CFE_SB_SendMsg()` if we can.  Keep it simple - as the vast majority of use cases will be through this API.

This is the only concern that I have.  Everything else looks good.

My other comment about making a local typedef for the "send HK" commands can be done as a follow on.

skliper:
RE; Send message more complicated...

I'm of the opinion we are better off educating the user what the new flag actually means.  CFE_SB_SendMsg vs CFE_SB_PassMsg really seems less informative than an explicit flag to update the sequence counter.  In the old documentation I didn't think it was obvious that the sequence counter in the message isn't getting incremented, it's actually that the sequence counter is getting filled in from the internally tracked SB value.

Normally I could be easily convinced to change, but I can see arguments supporting either approach with no clear winner.  My preference is to leave as is if we can... I'd rather not slow down this change if it can be avoided.

skliper:
Note using the suggested approach of incrementing sequence count for TransmitMsg and not for TransmitBuffer - I think that would open us up to inconsistencies based on which API the app decides to use.  The sequence count only increments when there is a subscriber now, which would no longer be the case for one of the APIs... etc.

fsw/cfe-core/src/es/cfe_es_task.h
```diff
@@ -174,31 +171,31 @@ void  CFE_ES_BackgroundCleanup(void);
...
-int32 CFE_ES_ResetPRCountCmd(const CFE_ES_ResetPRCount_t *data);
-int32 CFE_ES_SetMaxPRCountCmd(const CFE_ES_SetMaxPRCount_t *data);
-int32 CFE_ES_DeleteCDSCmd(const CFE_ES_DeleteCDS_t *data);
-int32 CFE_ES_StartPerfDataCmd(const CFE_ES_StartPerfData_t *data);
-int32 CFE_ES_StopPerfDataCmd(const CFE_ES_StopPerfData_t *data);
-int32 CFE_ES_SetPerfFilterMaskCmd(const CFE_ES_SetPerfFilterMask_t *data);
-int32 CFE_ES_SetPerfTriggerMaskCmd(const CFE_ES_SetPerfTriggerMask_t *data);
-int32 CFE_ES_SendMemPoolStatsCmd(const CFE_ES_SendMemPoolStats_t *data);
-int32 CFE_ES_DumpCDSRegistryCmd(const CFE_ES_DumpCDSRegistry_t *data);
+int32 CFE_ES_HousekeepingCmd(const CFE_MSG_CommandHeader_t *data);
```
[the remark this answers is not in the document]
skliper: I'd prefer this as a new issue.

skliper:
Or maybe we just leave CFE_SB_SendMsg, CFE_SB_PassMsg, CFE_SB_ZeroCopySend, CFE_SB_ZeroCopyPass in and have them wrap the transmit APIs?  Would just be a couple quick additional coverage tests... maybe call them CFE_SB_SendBuffer, PassBuffer instead of zero copy?  I still don't see a big impact here and it works as is, but if the current implementation is a show stopper...

ejtimmon:
FWIW, I think I agree with Jake on the topic of the added flag on CFE_SB_TransmitMsg.  I think it makes it clearer to the user what's actually happening.

jphickey (approved):
OK - I figured I'd bring up the concern about extra API complexity - but if folks are OK with it, then no need to hold this up.

To clarify though - my main suggestion was to reserve the full suite of options to users who choose the full (buffer-based) API - that is - obtain a buffer, fill it, then transmit the buffer.  So the "transmit buffer" API could have this sequence option (that is - put this bool on the TransmitBuffer call only) or it could be a totally separate flag on the buffer itself that says the sequence number is already set and is manipulated by other means.  It would not be different behavior depending on the API you chooise, it would be that the API you choose offers more control or simpler operation.

FWIW - we might need flags like that in the descriptor anyway - for some of the priority stuff that's been proposed, so it makes sense to have a flag in the descriptor that gets set once the sequence# is assigned, which prevents it from assigning a sequence number again.

Has anyone considered the possibility that a user receives a buffer and attempts to transmit it again?  My guess is that isn't supported, but hopefully it doesn't crash and burn if someone does it.

fsw/cfe-core/unit-test/CMakeLists.txt
```diff
@@ -71,7 +71,7 @@ foreach(MODULE ${CFE_CORE_MODULES})
         ${UT_COVERAGE_LINK_FLAGS}
         ut_cfe-core_support
         ut_cfe-core_stubs
-        sbr  # TODO remove this
+        sbr  # Will be removed when sbr stubs are implemented
```
astrogeco, line 74: Do we have an issue number for this TODO?

docs/cFE Application Developers Guide.md
```diff
@@ -1080,21 +1080,17 @@ void SAMPLE_TaskMain(void)
         /*
```
astrogeco, line 1080: Can we rename this file so it does not use spaces?

fsw/cfe-core/src/es/cfe_es_task.h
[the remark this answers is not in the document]
astrogeco: See #1037 @jphickey can you add more details or clean up the issue?

astrogeco:
@skliper what's the path forward on this? There's a lot of comments...

docs/cFE Application Developers Guide.md
skliper: Please do as a separate issue

fsw/cfe-core/unit-test/CMakeLists.txt
skliper: Yes, #962.

skliper:
@astrogeco There's no open issues against merging this that I can tell.  Let me know if I'm mistaken.

fsw/cfe-core/src/es/cfe_es_task.h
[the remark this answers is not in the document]
skliper: See #1039

astrogeco:
> @astrogeco There's no open issues against merging this that I can tell. Let me know if I'm mistaken.

Should have changed the base yesterday... We got conflicts now :/

skliper:
@astrogeco fixed conflict.

PR #1017, opened 2020-11-16
Fix #965, remove OSAL ID from App/LibInfo struct

### LINKED ISSUE #965: Avoid OSAL IDs in messages/data files
OSAL runtime types aren't really defined as part of the external (i.e. mission-scope) interface.  The `osal_id_t` is one such example - it is a local runtime type, not really intended to be saved in data files or command/telemetry messages.

This might seem pedantic because it is simply a `uint32`, but when using CFE with a command/data dictionary tool this becomes apparent that the OSAL types used in telemetry and data files aren't part of the data dictionary.  (OSAL itself doesn't have a cmd/tlm interface so it naturally wouldn't provide any such entity).

Use the `CFE_ES_ResourceID_t` instead.  Call `CFE_ES_ResourceID_FromOSAL()` when writing and `CFE_ES_ResourceID_ToOSAL()` when reading, to do the type conversion.  The underlying value is compatible (i.e. both 32 bit `uint32`, same numbers) so it should be transparent to external tools.

Have OSAL provide a stablized CMD/TLM definition of IDs?  (but CMD/TLM interface is not really part of OSAL's role - it is a runtime library)

**Requester Info**

jphickey:
**Describe the contribution**
Removes the Module ID from the App/Lib info telemetry structures.

Fixes #965 

**Testing performed**
Build and sanity test CFE
Run all unit tests

**Expected behavior changes**
App and Lib info telemetry structures no longer contain the "ModuleId" value from OSAL.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
CFE code should be distinct about the types used in messages vs. the types used at runtime.  They may need to be different depending on mission architecture (i.e. mixing 32 and 64 bits, different local platform-specific sizes of things, etc).  All message definitions should be in the proper app-specific message typedef files and must have mission scope, not platform scope.

The "osal_id_t" type isn't defined in any of the CFE message/interface header files for use within telemetry, but this was included in the AppInfo and LibInfo telemetry structures.

This ID is an ephemeral runtime value and is not relevant to a ground system or anything outside CFE, so it makes most sense to simply remove this ID from the telemetry structure.  Note that all commands are name-based, not ID-based, hence why this ID is not really useful.

The alternative would be to use CFE_ES_ResourceID_t type instead.  This is the same underlying value but defined in ES rather than OSAL, in the typedefs file with all other message types.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_api.c
+++ b/fsw/cfe-core/src/es/cfe_es_api.c
@@ -970,6 +970,7 @@ int32 CFE_ES_GetAppInfo(CFE_ES_AppInfo_t *AppInfo, CFE_ES_ResourceID_t AppId)
    CFE_ES_AppRecord_t *AppRecPtr;
    CFE_ES_TaskRecord_t *TaskRecPtr;
    int32              Status;
+   osal_id_t          ModuleId;
    uint32             i;
 
    if ( AppInfo == NULL )
@@ -979,6 +980,7 @@ int32 CFE_ES_GetAppInfo(CFE_ES_AppInfo_t *AppInfo, CFE_ES_ResourceID_t AppId)
    }
 
    memset(AppInfo, 0, sizeof(*AppInfo));
+   ModuleId = OS_OBJECT_ID_UNDEFINED;
 
    AppRecPtr = CFE_ES_LocateAppRecordByID(AppId);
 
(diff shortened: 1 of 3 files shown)


astrogeco:
- Missing an update from @jphickey local repo

jphickey:
Commit 720525ed is the corrected version (it actually just changes a little bit less than the original with the hope of reducing merge conflicts)

astrogeco:
@jphickey can you clear the conflicts here?

jphickey:
@astrogeco - conflict resolved in the IC branch.

PR #1022, opened 2020-11-17
Fix #1004, default stack size

### LINKED ISSUE #1004: Documentation/usage mismatch in CFE ES "Start App" command
**Describe the bug**
The `CFE_PLATFORM_ES_DEFAULT_STACK_SIZE` is documented as being a default stack size, not a minimum stack size.  But the CFE ES "Start App" command enforces it as a minimum value here:

https://github.com/nasa/cFE/blob/9804b59eb34154b10bb2946bde7654f3d27f4b85/fsw/cfe-core/src/es/cfe_es_task.c#L900-L906

But this is not in agreement with how it is documented:

https://github.com/nasa/cFE/blob/9804b59eb34154b10bb2946bde7654f3d27f4b85/cmake/sample_defs/cpu1_platform_cfg.h#L1186-L1201

**To Reproduce**
N/A

**Expected behavior**
Should not enforce the default as a minimum.  

I don't see any CFE platform definition for an enforced minimum stack size.  If I remember correctly this was discussed once or twice and the agreement was that this is an operational issue - stack size requirements depend on the app stack usage and the memory constraints of the platform - so CFE cannot (and should not) impose some random limitations on it - it should attempt to do what the user requested.

So recommendation would be to remove this check.
One valid possibility is that if the stack size is specified as 0 (which is definitely not valid), to use the default value instead.

**System observed on:**
Ubuntu 20.04

**Reporter Info**

jphickey:
**Describe the contribution**

Do not enforce `CFE_PLATFORM_ES_DEFAULT_STACK_SIZE` as a minimum, it should be a default.

Fixes #1004

**Testing performed**
Build and sanity test CFE.
Run all unit tests.

**Expected behavior changes**
This affects the Start App command.
If stack size is specified as zero, then the default stack size value from platform config is used.  Otherwise the value in the command will be passed through and used as-is.  

**System(s) tested on**
Ubuntu 20.04
RTEMS 4.11

**Additional context**
No specific value is enforced at the ES level.  Many RTOS's will implement any stack size requested - so no need for ES to get in the way and put arbitrary restrictions.  Furthermore the value is documented as a default, not a minimum.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_task.c
+++ b/fsw/cfe-core/src/es/cfe_es_task.c
@@ -861,6 +861,7 @@ int32 CFE_ES_StartAppCmd(const CFE_ES_StartApp_t *data)
     int32                 FilenameLen;
     int32                 AppEntryLen;
     int32                 AppNameLen;
+    size_t                AppStackSize;
     char                  LocalFile[OS_MAX_PATH_LEN];
     char                  LocalEntryPt[OS_MAX_API_NAME];
     char                  LocalAppName[OS_MAX_API_NAME];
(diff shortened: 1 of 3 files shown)


astrogeco:

PR #1025, opened 2020-11-18
Fix #1021, pool buffer type

### LINKED ISSUE #1021: Pool buffers should not use `uint32*` as interface type
The ES memory pool API uses `uint32*` as a buffer pointer - this is the type used in  `CFE_ES_GetPoolBuf()` and `CFE_ES_PutPoolBuf()` among others.

This presents a few usability problems:

- Most often the data being stored is _not_ actually `uint32` - so it generally needs to be type cast by the user.
- Typecasts are ugly and risky
- Specifically - Typecasts to/from `uint32*` might create a warning about alignment on some platforms  (one direction or the other is likely to be seen as an upgrade in alignment requirement)
- If alignment was the goal, 32 bits is still too low for `double` type, or a 64-bit pointer, so it fails at that job.

The API should use `void*`.  

**Additional context**
Although this basically turns off type checking, there was no real type checking here to begin with, and pool buffers _intentionally_ should be convertible to any type, so no need for it here anyway.

Using `void*` will clean up the code substantially, allowing a lot of unnecessary type casts to be removed.

**Requester Info**

jphickey:
**Describe the contribution**

Use a generic `void*` as the interface type for the pool buffer pointers.  This reduces the need for local type casting in the apps and makes it generally easier to use.

Fixes #1021 

**Testing performed**
Build and sanity test CFE
Run all unit tests

**Expected behavior changes**
No impact to behavior. 

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Note that getting a buffer in one step still requires a type cast, as the interface here is really `void**`.  However when putting or querying a buffer no cast is needed.  One can also avoid the need to cast if they get a buffer in two steps (i.e. get as `void*` first, then convert to real type with a simple assignment.


jphickey:
Real commit is in 3e347e55, need to get a new baseline first to avoid conflicts.

This can be reviewed but it will need a rebase before final merging.

astrogeco:
- @jphickey to rebase so we can include it in this merge cycle

jphickey:
Rebased and ready to go.

astrogeco:
[CCB 2020-12-02](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-02-2020) Minor Cleanup

PR #1028, opened 2020-11-23
Fix merge conflicts in PR #998 

astrogeco:
**Describe the contribution**
Took #998 and rebased on top of integration-candidate and attempted to solve merge conflicts

**Testing performed**
Built on docker container. Unit tests pass. 

See https://github.com/astrogeco/cFS/actions for more tests

### DIFF
--- a/docs/cFE Application Developers Guide.md
+++ b/docs/cFE Application Developers Guide.md
@@ -1216,7 +1216,7 @@ FILE: xx_app.c
 void XX_AppMain(void)
 {
     uint32 RunStatus = CFE_ES_RunStatus_APP_RUN;
-    CFE_SB_MsgPtr_t MsgPtr;
+    CFE_MSG_Message_t *MsgPtr;
     int32  Result = CFE_SUCCESS;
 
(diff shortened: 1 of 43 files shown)


astrogeco:
@jphickey and @skliper I tried to resolve the merge conflicts in #998 but I probably messed some things up. 

Some of the size_t changes did not play well with CFE_MSG_Size_t. I think I fixed all of them but it might be good to double check that everything works as intended. 

Feel free to force push to this branch and then we can either merge this one instead of #998 or force push to skliper/fix777-deprecate-sb

jphickey:
Updated with my take - basically I did the same but as a merge (not rebase) and just resolved the conflicts, tested, etc.  Mostly the same as what you had aside from some whitespace diffs and a few size_t/CFE_MSG_Size_t diffs.  

If you concur we should be able to just push (i.e. fast forward) the integration candidate here - don't really need a separate PR, I think.  It's just a merge.

skliper (commented):
I didn't check every line, but looks fine to me.

PR #1035, opened 2020-12-03
Fix 1034, Atomic type for SystemState

### LINKED ISSUE #1034: Use appropriate atomic type for inter-thread sync
The RV tool analysis reported several cases of reading/writing shared memory variables without a lock. 

The intention behind the code was that the data type being read/written simultaneously here was atomic in nature, thus the parallel access would be safe, as it is not possible to catch an atomic value mid-update.

To be more portable the code should use the C99 type `sigatomic_t` to ensure that the data type is in fact atomic on the given platform.  Currently it is using `uint32` which is not guaranteed to be atomic on all platforms.

Calling this "minor" because the `uint32` type will be atomic on all the platforms that the code in question is actually used on.  There is no bug currently here, this is just to prevent a future bug if this code is expanded to smaller CPUs (microcontrollers) where `uint32` is not atomic.

skliper:
**Describe the contribution**
Fix #1034 
- Use volatile sig_atomic_t for system state avoids race issue if uint32 isn't atomic on a system (microcontroller?)

**Testing performed**
Built and ran unit tests, passed.

**Expected behavior changes**
None

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_global.h
+++ b/fsw/cfe-core/src/es/cfe_es_global.h
@@ -52,6 +52,8 @@
 #include "cfe_evs.h"
 #include "cfe_psp.h"
 
+#include <signal.h>
+
 
 /*
 ** Typedefs
@@ -104,7 +106,7 @@ typedef struct
    /*
    ** Startup Sync
    */
-   uint32  SystemState;
+   volatile sig_atomic_t SystemState;
 
    /*
    ** ES Task Table


astrogeco:
[CCB 2020-12-09](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-09-2020) APPROVED
- is sig_atomic a C99 type? Yes: https://www.gnu.org/software/libc/manual/html_node/Atomic-Types.html

PR #1038, opened 2020-12-07
Fix #952, OSAL module flags to permit app reload

### LINKED ISSUE #952: Requirement cES1008 Failure
**Describe the bug**
Requirement cES1008 specifies:  Upon receipt of a Command the cFE shall Reload the Command specified cFE Application from the Command specified cFE Application file.

This is not consistent with how the FSW works.  If a reload command is received, the FSW will restart the Application with the previously loaded version as long as the command specified cFE Application file exists.  The same result is seen if an App is started/stopped - then loaded/started with a new version.  The system/event log entries will indicate that the loading/starting of the new version of the App is successful...however the original version of the App will be the one that will actually be executed.   

**To Reproduce**
1. Start an App
2. Send the reload command with a filename that contains a different version of the App (App name must remain constant)
3. Verify that the original version of the App was actually the one started.

Alternatively:
1. Start an App
2. Stop the App
3. Start a different version of the same App (App name must remain constant)
4. Verify that the original version of the App was actually the one started.

**Expected behavior**
Able to restart a task of the the same name.

**Code snips**
If applicable, add references to the software.

**System observed on:**
OS: ubuntu-19.10
Versions:  cfe:  v6.7.0+dev295; osal: v5.0.0+dev247; psp:  v1.4.14.0

**Reporter Info**
Dan Knutsen

jphickey:
**Describe the contribution**
Set the flags parameter on the `OS_ModuleLoad()` properly to allow an app to be properly unloaded, which in turn allows the reload command to work as expected.

Fixes #952 

**Testing performed**
Build and run all unit tests
Test CFE app reload procedure:
- Run CFE as normal, confirm unmodified SAMPLE_APP is loaded and running 
- Build a modified SAMPLE_APP that has an additional `OS_printf()` call during startup, and install as `sample_new.so`
- Issue "reload" command (7) via cmdUtil, to reload SAMPLE_APP from `/cf/sample_new.so`
- Confirm that the expected printf message is visible (confirms that the new file was actually loaded)

**Expected behavior changes**
Module unload/reload works as expected on Linux.  Previously the unload would not actually unload, so it would end up restarting the same app code over again.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Needs sync delete fix in OSAL (see nasa/osal#642) to avoid task delete race condition.  Without this OSAL change, the task may still be running at the time CFE ES unloads it, which will likely cause a segfault/crash if that happens.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_api.c
+++ b/fsw/cfe-core/src/es/cfe_es_api.c
@@ -1117,12 +1117,9 @@ int32 CFE_ES_GetLibInfo(CFE_ES_AppInfo_t *LibInfo, CFE_ES_ResourceID_t LibId)
 */
 int32 CFE_ES_GetModuleInfo(CFE_ES_AppInfo_t *ModuleInfo, CFE_ES_ResourceID_t ResourceId)
 {
-    uint32 ResourceType;
     int32 Status;
 
-    ResourceType = CFE_ES_ResourceID_ToInteger(ResourceId);
-    ResourceType -= ResourceType & CFE_ES_RESOURCEID_MAX;
-    switch(ResourceType)
+    switch(CFE_ES_ResourceID_GetBase(ResourceId))
     {
     case CFE_ES_APPID_BASE:
         Status = CFE_ES_GetAppInfo(ModuleInfo, ResourceId);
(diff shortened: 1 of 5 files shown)


astrogeco:
[CCB 2020-12-09](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-December-09-2020) APPROVED

Issue #915, opened 2020-09-28
cFE req. cES1007.2: Restart Application - Reject On Missing File


dmknutsen:
**Describe the bug**
cFE requirement cES1007.2 specifies that a user can not restart an Application if the original file has been removed.  If the user attempts to restart an application following removal of file, the cFE shall reject the Command, increment the invalid Command counter, and generate an event message. 

This is not what happens when the restart application command is received.  If a user has deleted the cFE Application file and then sends the restart application command.  The application successfully restarts.

**To Reproduce**
Delete an applications shared library file from the /cf/ location.
Send the associated restart application command.

**System observed on:**
Oracle VM VirtualBox
OS: ubuntu-19.10
Versions: cfe v6.8.0-rc1+dev28, osal v5.1.0-rc1+dev12, psp v1.4.0+dev76

**Reporter Info**
Dan Knutsen
NASA/Goddard

skliper:
Related to #424, which was reported as a requirements failure since the restart didn't work.  Need to resolve if something changed, or a different process is being used, or if it's a system dependency.

skliper:
Now that osal is working as expected, we are still failing this requirement since the command doesn't error (the request to restart is sent correctly so the valid command counter is incremented).  Likely worth just adding the file check so it is even more similar to reload.

skliper:
Actually there's 3 requirements that no longer are met in terms of incrementing the error counter:
cES1007.2 (already documented above)
cES1007.3:  If the cFE Application Restart fails due to a non-parameter error, then the cFE shall delete the Application, **increment the invalid Command counter**, and generate an event message.
cES1008.3: If the cFE Application Reload fails due to a non-parameter error, then the cFE shall delete the Application, **increment the invalid Command counter**, and generate an event message.

Command just does the request, failure to load due to a non-parameter error does not increment the invalid command counter.

skliper:
This issue is partially fixed by #1038, still will fail on missing file though without incrementing the invalid command counter.

PR #1083, opened 2021-01-11
Fix #793, Clarify restart/reload app behavior

### LINKED ISSUE #793: CFE RELOAD and RESTART commands handled (almost) identically
**Describe the bug**
The documentation of the `CFE_ES_RESTART_APP_CC` specifically says here that it is _not_ reloaded from the disk file:

https://github.com/nasa/cFE/blob/5ac3f0a02038b82894c1803339f92b294ae9f0b7/fsw/cfe-core/src/inc/cfe_es_msg.h#L324

But in the way the code has evolved, both RELOAD and RESTART control requests end up going through `CFE_ES_CleanUpApp()` function:

https://github.com/nasa/cFE/blob/5ac3f0a02038b82894c1803339f92b294ae9f0b7/fsw/cfe-core/src/es/cfe_es_apps.c#L961-L968
https://github.com/nasa/cFE/blob/5ac3f0a02038b82894c1803339f92b294ae9f0b7/fsw/cfe-core/src/es/cfe_es_apps.c#L997-L1003

Notably, the `CFE_ES_CleanUpApp()` function will, in fact, unload the module via `OS_ModuleUnload()`, and the subsequent `CFE_ES_AppCreate()` function will load it again from disk.  Also important that the AppID might change too as part of this process, which may or may not be expected?

**Expected behavior**
Should make the documentation and code match one way or another:
- If we want a true "restart" without reload, as the documentation for `CFE_ES_RESTART_APP_CC` says, we need to update this to NOT completely unload the module.
- OR If the current implementation is OK then I'd say it isn't sufficiently different from `CFE_ES_RELOAD_APP_CC` to warrant the existence of a separate command.

**Additional context**
I noticed this inconsistency while doing implementation of #28.  I can put in a fix for this issues as part of the same (upcoming) PR, just need CCB concurrence on which way to go - do we make it work as described, or we describe the way it works.

**Reporter Info**

skliper:
**Describe the contribution**
Fix #793 - Clarifies the difference between restart and reload in API/cmd and user's guide documentation

**Testing performed**
None, comments only

**Expected behavior changes**
None

**System(s) tested on**
N/A

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/docs/src/cfe_es.dox
+++ b/docs/src/cfe_es.dox
@@ -366,9 +366,16 @@
 /**
   \page cfeesugapprestart Restarting an Application
 
-  The  #CFE_ES_RESTART_APP_CC command is used to restart an application. 
-  This command stops and restarts an application using the parameters 
-  defined when the application was originally started, either through 
+  The #CFE_ES_RESTART_APP_CC command is used to restart an application
+  using the same file name as the last start.
+
+  This command checks for file existence, the application is running,
+  and the application is not a core app.  If valid, the application restart
+  is requested.
+
+  When requested, ES stops the application, unloads the object file, loads the object file
+  using the previous file name, and restarts an application using the parameters
+  defined when the application was previously started, either through
   the startup script or by way of the #CFE_ES_START_APP_CC command. 
     
   Next: \ref cfeesugappreload <BR>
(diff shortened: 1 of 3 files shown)


skliper:
After reviewing #915 and testing the current behavior, these documentation updates may be premature if the fix for #915 is to check for file existence as part of command processing.

astrogeco:
[CCB:2021-01-13](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-13-2021) 

Also see https://github.com/nasa/cFE/issues/915 for context

skliper:
Refactored based on #1128 and resolution of #915, ready to go.

Issue #1078, opened 2021-01-08
CFE_SB_ReceiveBuffer() -- switch timeout to int32


CDKnightNASA:
**Describe the bug**
Per discussion at the Jan. 6, 2020 CCB, @jphickey suggested that the timeout parameter should be int32, not uint32. This should be changed.

See also #1063 

**Reporter Info**
Christopher.D.Knight@nasa.gov

jphickey:
CFE_SB_ReceiveBuffer() is already `int32` (see https://github.com/nasa/cFE/blob/1ede295c3874e6571aca965a7850d950cf7aa1a3/fsw/cfe-core/src/sb/cfe_sb_api.c#L1470-L1473)

But `CFE_SB_ReadBuffer` (internal function) is not.

Unless there is a good reason to "hide" OSAL definitions here (and I can't think of a good reason, because everything already depends on osapi.h anyway), I'd rather simplify and just make it so we have:

```
#define CFE_SB_PEND_FOREVER   OS_PEND
#define CFE_SB_POLL           OS_CHECK
```

... rather than redefining these values.

Right now we use the same underlying values but redefine them, so to be correct/safe the ReadQueue needs to translate.  With them defined this way we still have symbol independence (so SB can diverge from OSAL if it needs to, though I don't know why it would).  Then the ReadQueue function doesn't need to translate this value, so long as it shares the same definition.

jphickey:
Actually, just assign this one to me.  As I'm going through code looking for race conditions in issue #1073 the entire design of CFE_SB_ReadQueue() is broken - it is accessing PipeDscPtr without the SB lock.  I will replace this with something that works.

Issue #287, opened 2019-09-30
SB: remove AppName (and PipeName?) from CFE_SB_PipeD_t


skliper:
In looking at SB code, I noticed that CFE_SB_PipeD_t has two character arrays, AppName and PipeName. AppName, particularly, is a waste of memory to have in each pipe structure (it should be maintained in one location) and there is an AppId field in the PipeD_t so it's not needed for finding pipes.

Same may go with PipeName, as the OSAL queue takes the name as a parameter (need to ensure the pipe name doesn't already exist in OSAL.)

skliper:
_Imported from trac issue 256.  Created by cdknight on 2018-10-23T14:24:29, last modified: 2019-03-01T15:27:58_

skliper:
_Trac comment by cdknight on 2018-10-23 22:15:56:_

PipeName moved to a separate ticket as GetPipeName currently returns the char * from the struct. See #288

skliper:
_Trac comment by cdknight on 2018-10-23 22:25:46:_

see [changeset:61178a]

skliper:
_Trac comment by jhageman on 2019-02-25 16:13:25:_

2018/11/13: CCB noted "feedback reviewed; will annotate ticket".  State is waiting for owner to complete this action.

CCB - Should review, not clear what action is.

skliper:
Note - AppName is the only remaining char array in the CFE_SB_PipeD_t structure.  The best I can tell this ticket still needs investigation and/or proposed design update to resolve.

PR #1092, opened 2021-01-13
Fix #1073, software bus locking

### LINKED ISSUE #948: Possible double locks in SB
**Describe the bug**
There's a potential for double locks in the following code (just from code review, haven't tested to confirm):

https://github.com/nasa/cFE/blob/dc3d62b37d2bacd8fe23f66ff7b64025ac7d0751/fsw/cfe-core/src/sb/cfe_sb_api.c#L1110-L1113
https://github.com/nasa/cFE/blob/dc3d62b37d2bacd8fe23f66ff7b64025ac7d0751/fsw/cfe-core/src/sb/cfe_sb_api.c#L1121-L1124
https://github.com/nasa/cFE/blob/dc3d62b37d2bacd8fe23f66ff7b64025ac7d0751/fsw/cfe-core/src/sb/cfe_sb_task.c#L1172-L1173
https://github.com/nasa/cFE/blob/dc3d62b37d2bacd8fe23f66ff7b64025ac7d0751/fsw/cfe-core/src/sb/cfe_sb_task.c#L1200-L1201

Related - the locking in the SendPrevSubs command handling doesn't look like it really helps since it has to unlock to send the message (same issues as the commands to record route/map info to file), typical use case is to enable subscription reporting, then send all previous subscriptions so may make sense to refactor (and possibly throttle).

**To Reproduce**
Haven't tested yet, but could clear filters on the debug messages and trigger, or just subscribe and unsubscribe twice to trigger CFE_SB_UNSUB_NO_SUBS_EID.

**Expected behavior**
Shouldn't double lock.

**Code snips**
See above.

**System observed on:**
NA - haven't tested yet

**Additional context**
From analysis during #928 and #947 

**Reporter Info**
Jacob Hageman - NASA/GSFC

### LINKED ISSUE #985: Use Resource ID type for SB Pipe ID
The CFE SB pipe ID table is not safe from aliasing or other issues.

Redefine the CFE_SB_PipeID_t type to be a form of CFE_ES_ResourceID_t like many other resources have already been converted (mem pool handles, CDS blocks, etc).

Re-Use all the same management patterns of this structure.

**Additional context**
The only potential downside is that Resource IDs are defined as 32 bit values but Pipe IDs were only 8 bits.  So this will make `sizeof(CFE_SB_PipeId_t)` into 4 instead of 1.  But as long as apps are properly using the typedef and not assuming `uint8` or otherwise depending on this being a single byte, this shouldn't be noticeable.

This is somewhat related to previous issue #100 - implementing this would be a good step in the right direction for that issue too.

**Requester Info**

### LINKED ISSUE #1073: CFE intermittently showing invalid message ID errors
**Describe the bug**
When leaving CFE running for long periods of time, I am seeing occasional errors related to bad message IDs/Commands appear.

For instance:
`EVS Port1 66/1/CFE_EVS 5: Invalid command packet, Message ID = 0x00001810`

I was actually able to catch one of these in a core dump, and I can confirm that `SBBufPtr->Msg` looks like a perfectly well-formed message with MsgID of 0x1810.  This particular example was caught at https://github.com/nasa/cFE/blob/56397a3a2c47c674892c14610e78263030441d73/fsw/cfe-core/src/evs/cfe_evs_task.c#L378.

```
(gdb) print /x *((CFE_MSG_CommandHeader_t*)SBBufPtr)
$2 = {Msg = {CCSDS = {Pri = {StreamId = {0x18, 0x10}, Sequence = {0xc0, 0x0}, Length = {0x0, 0x1}}}, Byte = {0x18, 0x10, 0xc0, 0x0, 0x0, 0x1}}, Sec = {FunctionCode = 0x0, Checksum = 0x0}}
```

What is particularly mysterious is that 0x1810 is `CFE_TIME_TONE_CMD_MID` ... These MIDs should be delivered only to CFE_TIME - EVS does not subscribe to these - yet for some reason it appears to be delivered to EVS....

**To Reproduce**
Build CFE in default debug config (SIMULATION=native, etc).  Start up CFE core and just let it run.  It may take several hours for the first error to appear.  CFE appears to continue running normally afterwards though.

**Expected behavior**
EVS (or other apps!) should not randomly see messages it didn't subscribe to.

**System observed on:**
Ubuntu 20.04 (native)

**Additional context**
Not sure when this started because everything seems fine when debugging CFE for shorter time periods.   But over the last couple weeks I've noticed these messages randomly appearing in terminal windows where I've started CFE and forgotten about it (overnight or longer).

Although I cite EVS here (because its the one I actually caught in a core file) this phenomenon occurs in other apps too (SB, TIME, etc) all randomly getting a delivered message ID they didn't subscribe to.

**Reporter Info**

### LINKED ISSUE #1096: Possible race conditions in SB Pipe operations
**Describe the bug**
As part of debugging #1073 recently, it was noted that several historical SB APIs are not adequately locking the global data structures that they access.  In particular during pipe operations and subscription/unsubscription, some functions are accessing the "CFE_SB_PipeD_t" objects (pipe descriptors) in the `CFE_SB.PipeTbl` and routes in the routing table outside any SB shared data lock, or are not maintaining the lock from start to finish.  

It is therefore possible that another task can modify this data while being accessed by another task.

**To Reproduce**
Not actually observed "in the wild" - this was found by code inspection.  To manifest this would require running CFE on a multi-core system where task(s) were actively subscribing/unsubscribing or creating/deleting pipes _at the same time_.  

**Expected behavior**
All SB global data access should be protected against any possible concurrency issues.

**Code snips**
Some examples of potentially affected routines / operations in older CFE versions (these are observed in CFE 6.8, just a quick skim though - this should NOT be considered an exhaustive/complete list):  

- CFE_SB_EnableRouteCmd() / CFE_SB_DisableRouteCmd() - accesses pipes and routes with no lock
- CFE_SB_DeletePipeFull() - locks initially, but releases lock in the middle of "destptr" loop to do unsubscribe, which could possibly also release another task that also accesses the same pipe or change routing (i.e. changing the same list).
- CFE_SB_GetPipeName() - accesses PipeTbl without lock
- CFE_SB_SubscribeFull() - when sending the final "subscription report" telemetry message, this is actually stored a global, filled while locked, but actually _sent_ after unlock.  There is a possibility between the unlock and the send that another higher-priority task will do a subscribe and overwrite the message.

**System observed on:**
N/A

**Additional context**
During "steady state" operations where the routing table is not being changed, there isn't much risk, as normal sending/receiving of messages does not actually modify the data in the routing table.

**Reporter Info**

jphickey:
**Describe the contribution**

The root cause of the mysterious message delivery issues was tracked down to inconsistent locking.  When running on multi core systems a variety of buffer corruption issues were observed.  This fix is really a rework of all CFE SB API implementations toward the goal of more consistent locking patterns to solve this issue.

Because SB APIs make heavy use of event IDs, this presents a challenge because events can only be sent while unlocked.  In order to ensure that all events are generated and all counters are incremented consistently, the functions should run through to the end and not return early, and use a "PendingEventID" register to record what event ID should be sent per the current operation, rather than simply sending the event at the time the condition is identified.  

The general pattern becomes:

1. Initial checks without lock
2. Lock SB
3. Perform query/modification of SB pipe/routing tables as needed while locked
4. Unlock SB
5. Call into other subsystems as needed (e.g. OSAL)
6. (Optional) Re-lock SB global to update based on the result of (5), and unlock again
7. Send pending event ID.

This also employs the `CFE_ES_ResourceID_t` type and related patterns for managing the SB Pipe IDs.  In particular this (intentionally) makes it not possible to use this directly an array index, and will break code which directly accessed these items without going through the lookup function.

Fixes #985 
Fixes #1073 
Fixes #1096

**Testing performed**
Build and sanity check CFE
Run all unit tests
Also Let CFE run for approx 72 hours, and observed no more "invalid msgid" or memory pool errors.

**Expected behavior changes**
- __The `CFE_SB_PipeId_t` type is no longer usable as a direct array index__.  It is also increased in size from 8 bits to 32 bits.   But as a result is now consistent with all other ID types in both behavior and size.  As long as the typedef is used and PipeIds are used only as intended,, this should not be noticeable to apps.
- __The "pipe stats" structure in the Pipe TLM is also changed__.  This structure contained a `CFE_SB_PipeId_t` value, hence why it had to be updated because the type is now bigger.  The spare bytes are also moved to the end of the struct.

Neither of these are expected to be particularly problematic - just stuff to be aware of.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Not sure if the change to `CFE_SB_PipeId_t` size (and the TLM message that contains this type) might require an update to cFS-GroundSystem.

EDIT - Fix #948 - eliminates identified double locks.


jphickey:
Note this also fixes #1078 -as  it removes the problematic routine entirely.

astrogeco:
[CCB:2021-01-13](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-13-2021)

Fixes concurrency issues in SB. 

Will probably follow up with moving this to a module since there's multiple ways to implement the Resource ID functionality.

Accessing tables using PipeID will no longer be available to external access.

ResourceIDs are no longer recycled and removes risk of PipeID aliasing

Update cFS-GroundSystem to reflect changes in structure sizes. 

@jphickey Create "retroactive" issue to document bugs with event and error messages inconsistencies due to lock handling. 

Remove lingering unnecessary debug artifacts

Link this to the "remove pipename from pipe structure" #287

CDKnightNASA:
This is excellent work, @jphickey ! I agree with all of your changes. Given the scope of these changes, how should we test and roll this out? SBN touches a fair bit of SB, and I've been trying to bring SBN up to "caelum". I may try this branch against my code but do we plan to merge into main soon?

(As an aside, I started at NASA as a contractor for a company called Caelum.)

jphickey:
BTW- regarding discussion of old versions that are affected by this bug.... I wanted to add one important detail:

Although some general race conditions certainly have existed across SB for quite some time, this particular issue (or at the least the specific race condition that was actually causing failures described in #1073) was introduced as part of the SB API refactor that was done a month or two ago - that is in the transition to `CFE_SB_TransmitMsg()` instead of `CFE_SB_SendMsg()`.  Specifically the call to `CFE_SB_GetBufferFromPool()` to get the buffer was not serialized as it needed to be.  In the old `CFE_SB_SendMsgFull()` implementation in bootes this actually was inside a locked section, but after migrating to `CFE_SB_TransmitMsg()` it was not invoked from a locked section anymore.  I suspect (though I cannot confirm) that getting the buffer from the pool was the actual problem area causing the bad msg ID errors - because if two tasks call this simultaneously, there is a possibility that they both get the same buffer.

Of course there were a ton of other race conditions as well, such as accessing the pipe/routing table - but these didn't seem to manifest themselves in a noticeable way.

The other race conditions fixed by this PR address problem areas with Creating/Deleting SB pipes and route subscribe/unsubscribe ops.  While Bootes probably has problems there if you push it hard enough, most FSW doesn't do repetitive subscribe/unsubscribe operations enough to expose these problems. 

So in other words - I wouldn't necessarily worry about Bootes for normal steady-state message passing.  The specific issue that likely manifested itself as #1073 was introduced more recently.

jphickey:
@CDKnightNASA -

> how should we test and roll this out? SBN touches a fair bit of

Yes SBN is an important use case that I was not really able to test as part of this - I don't really have any test setups that use SBN.  I'd be hoping that you (or someone!) can pull this into an SBN setup.

Note that aside from the scary-looking refactoring, it actually doesn't change that much in the way of actual _logic_.  It gets rid of some redundant struct members and consolidates access to the structs into a locked section, but tries hard not to actually change the behavior of the way these things actually operated.

jphickey:
Rebased to "main" branch (2021-01-12 baseline) and resolved conflicts

astrogeco:
[CCB:2021-01-21](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-13-2021) APPROVED

PR #1107, opened 2021-01-19
Fix #955, Remove OS_printf stub count checks in UT

### LINKED ISSUE #955: UT checking calls to OS_printf
Many unit tests are checking the calls to `OS_printf()`.  But typically all syslog/printf type messages are not fulfilling any sort of requirements, they are just informational in nature.  

So AFAIK there isn't really a strong justification to specifically check for certain syslog messages, but the fact that the UT does check for these means that any time the syslog messages are changed or refactored in any way, one gets a bunch of nuisance UT failures.  As a result we spend a lot of time fixing UT tests for things that aren't really relevant to the operation of the code.

At least remove checks for number of times `OS_printf()` was called in a given path.  This is really quite irrelevant to pass/fail. 

The `UT_PrintfIsInHistory` tests might be OK to keep, but an actual return code should be preferred.  If there is no return code then this is probably OK - its also less volatile because it just checks format string and confirms that a specific path was taken.

**Requester Info**

skliper:
**Describe the contribution**
Fix #955 - removes checking of OS_printf # of stub calls in unit tests (mostly ES, 2 in EVS).  Checks for specific format string in history instead (added 2 checks where they were missing) to confirm the right path was taken.

**Testing performed**
Built and ran unit tests, passed

**Expected behavior changes**
None, reduces technical debt (no need to fix sub counts all over when implementation changes)

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/unit-test/es_UT.c
+++ b/fsw/cfe-core/unit-test/es_UT.c
@@ -882,8 +882,7 @@ void TestStartupErrorPaths(void)
     CFE_ES_InitializeFileSystems(CFE_PSP_RST_TYPE_POWERON);
     UT_Report(__FILE__, __LINE__,
               UT_PrintfIsInHistory(UT_OSP_MESSAGES[UT_OSP_CREATE_VOLATILE]) &&
-              UT_PrintfIsInHistory(UT_OSP_MESSAGES[UT_OSP_MOUNT_VOLATILE]) && 
-              UT_GetStubCount(UT_KEY(OS_printf)) == 2,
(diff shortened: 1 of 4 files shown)


astrogeco:
[CCB:2021-01-21](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-21-2021) APPROVED

PR #1124, opened 2021-01-25
Fix #1036, Clear task data in EarlyInit and standardize name

### LINKED ISSUE #1036: Clear BSS on App Restart
Under normal conditions, when a task is started, the process of loading
the task will clear the BSS segment (used to store global-scope variables
that are not explicitly initialized).

However, if a task is "restarted" without "reloading" it may find that
its BSS segment still retains the content from the prior run.

We discussed this at the CCB meeting on 2016-Jan-19, and the moderate
concensus was that it might be most useful for the BSP to explicitly clear
the BSS of tasks before they are started. According to my notes, Joe was
going to take a look at this.

skliper:
**Describe the contribution**
Fix #1036 - explicity memset the task data to zero at the start of EarlyInit.  Also standardizes the global typdef/variable names.

Might need a rebase based on the name changes.

**Testing performed**
Build and run unit tests, passed.

**Expected behavior changes**
None.

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04

**Additional context**
TBL global depends on `CFE_TBL_Global.TableTaskAppId` being set, which doesn't happen until the app starts (in `CFE_TBL_InitData`), so replaced where CFE_TBL_EarlyInit was being called to initialize the data with the test initialization function `Test_CFE_TBL_ApiInit`.  In the future these tests could be cleaned up by calling the initialization before each test, and ensuring there are no test configuration dependencies across tests.

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/fsw/cfe-core/src/evs/cfe_evs_log.c
+++ b/fsw/cfe-core/src/evs/cfe_evs_log.c
@@ -54,51 +54,51 @@
 void EVS_AddLog (CFE_EVS_LongEventTlm_t *EVS_PktPtr)
 {
 
-   if (CFE_EVS_GlobalData.EVS_TlmPkt.Payload.LogEnabled == true)
+   if (CFE_EVS_Global.EVS_TlmPkt.Payload.LogEnabled == true)
    {   
       /* Serialize access to event log control variables */
-      OS_MutSemTake(CFE_EVS_GlobalData.EVS_SharedDataMutexID);
+      OS_MutSemTake(CFE_EVS_Global.EVS_SharedDataMutexID);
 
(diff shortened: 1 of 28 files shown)


skliper:
Rebased and updated SB global name (finished work).

astrogeco:
[CCB:2021-01-27](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-27-2021) APPROVED
- There were two services that did the memset zero, this standardizes the practice
- Makes global variable names consistent as: `SERVICENAME_Global`
- Simplifies initialization
- `tbl_UT.c` needs cleanup since they depend on the order of how things are set up

PR #1129, opened 2021-01-25
Fix #609, Remove CFE_PLATFORM_EVS_LOG_ON undefined option to diasble log

### LINKED ISSUE #609: EVS "CFE_PLATFORM_EVS_LOG_ON" option unit test failure
**Describe the bug**
The Event Services subsystem has a broken compile-time platform option called `CFE_PLATFORM_EVS_LOG_ON`.  The description says: "The CFE_PLATFORM_EVS_LOG_ON configuration parameter must be defined to enable EVS event logging"

If UT is disabled, then CFE core itself actually seems to build and run OK.  However, certain risky things are not clear in the code that:

- The `EVS_SharedDataMutexID` will be left uninitialized
- The `EVS_LogPtr` will be left as NULL

The code that accesses these seems to be mostly protected by checking the separate `CFE_EVS_GlobalData.EVS_TlmPkt.Payload.LogEnabled` member boolean in the outgoing telemetry packet.  This seems like a weak design, in particular because the telemetry packet is supposed to be informational, not an active control structure.

**To Reproduce**
Disable the `CFE_PLATFORM_EVS_LOG_ON` option, and build with `ENABLE_UNIT_TESTS=TRUE`.  CFE EVS unit test fails to build with a compiler error.

**System observed on:**
Ubuntu 18.04 LTS 64 bit.

**Additional context**
Unless there is a specific requirement for `CFE_PLATFORM_EVS_LOG_ON` as it stands today, my recommendation would be to deprecate this option and keep it always on, which reduces the testing matrix, and makes the FSW code more consistent.  Platform config options that actually add/remove `#ifdef` code should be avoided, as this has proven to be a testing/support issue time and time again.

In this case, only the code that initializes the structures is compiled out.  All the code that reads/writes to it is still compiled in, but skipped via a runtime test.  So this isn't saving much in the way of code/text space.

If the goal of this option is to save data space memory, then _mostly_ the same effect can be achieved by keeping the log very small, by setting `CFE_PLATFORM_EVS_LOG_MAX` to a very low number, such as 1.  In this mode the log structure uses only 176 bytes of memory on an x86-64 machine, down from 3368 bytes with the default size of 20.  

And the unit tests still build and pass with the max set to 1, and it reduces the amount of conditionally-compiled code and variances on the FSW side.

**Reporter Info**

skliper:
**Describe the contribution**
Fix #609 - removes the ability to disable the log by not defining `CFE_PLATFORM_EVS_LOG_ON`

**Testing performed**
Build and run unit tests, passed

**Expected behavior changes**
No longer able to disable log completely, for minimum memory use define `CFE_PLATFORM_EVS_LOG_MAX` as 1

**System(s) tested on**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04

**Additional context**
Could remove control based on LogEnabled, panic on reset area fail and limp along if sem create fails.
Needs requirements updates - #1131

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/cmake/sample_defs/cpu1_platform_cfg.h
+++ b/cmake/sample_defs/cpu1_platform_cfg.h
@@ -1456,20 +1456,6 @@
 
 
 /**
-**  \cfeevscfg Enable or Disable EVS Local Event Log
-**
-**  \par Description:
-**       The CFE_PLATFORM_EVS_LOG_ON configuration parameter must be defined to enable EVS
-**       event logging. In order to disable the local event log this definition needs
-**       to be commented out.
-**
-**  \par Limits
-**       Not Applicable
-*/
-#define CFE_PLATFORM_EVS_LOG_ON
-
-
-/**
 **  \cfeevscfg Default Event Log Filename
 **
 **  \par Description:
--- a/fsw/cfe-core/src/evs/cfe_evs_task.c
+++ b/fsw/cfe-core/src/evs/cfe_evs_task.c
@@ -74,14 +74,10 @@ bool CFE_EVS_VerifyCmdLength(CFE_MSG_Message_t *MsgPtr, size_t ExpectedLength);
 int32 CFE_EVS_EarlyInit ( void )
 {
 
-#ifdef CFE_PLATFORM_EVS_LOG_ON
-
    int32                Status;
    uint32               resetAreaSize = 0;
    cpuaddr              resetAreaAddr;
    CFE_ES_ResetData_t  *CFE_EVS_ResetDataPtr = (CFE_ES_ResetData_t *) NULL;
-
-#endif
 
    memset(&CFE_EVS_GlobalData, 0, sizeof(CFE_EVS_GlobalData_t));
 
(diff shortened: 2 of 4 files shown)


astrogeco:
[CCB:2021-01-27](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-January-27-2021) APPROVED
- Option to reduce memory footprint is to set max = 1 
- Keep it simple. Having cFE do this is extra complexity that might not be necessary. Let PSP work with this.

skliper:
Need to remove control based on HK parameter before merge (will submit as an additional HOTFIX merge, or fasttrack?)

astrogeco:
@skliper this is in draft form, can I change this to ready?

skliper:
@astrogeco Good to go, opened #1153 to address the CCB actions.

astrogeco:
> @astrogeco Good to go, opened #1153 to address the CCB actions.

Should we address the requirements issue #1131 before merging this though?

astrogeco:
> > @astrogeco Good to go, opened #1153 to address the CCB actions.
> 
> Should we address the requirements issue #1131 before merging this though?

Nvm, clicked merge too soon :P

PR #1142, opened 2021-01-29
Fix #1141, pipe name unset

### LINKED ISSUE #1141: "PipeName" is unset for debug event
**Describe the bug**
Mistake in previous PR #1092 where the `PipeName` buffer is used to send a debug event on success but it is only set to a value on failure.

**To Reproduce**
Enabled debug events, and "Subscription Rcvd" events do not have a valid name.

**Expected behavior**
Need to get name in all cases, not just failure.

**System observed on:**
Ubuntu 20.04

**Additional context**
My preference would be to only print the ID in a debug event, as its faster and it saves the work of copying the name since most of the time debug events are turned off anyway.  But a "quick fix" that doesn't change the format is to just move the line.

**Reporter Info**

jphickey:
**Describe the contribution**
Fix #1141 Only print the Pipe ID in the debug event

**Testing performed**
Enable debug events and run CFE, observe events

**Expected behavior changes**
Debug subscription events have pipe ID, no name.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -1142,9 +1142,9 @@ int32  CFE_SB_SubscribeFull(CFE_SB_MsgId_t   MsgId,
     {
         /* If no other event pending, send a debug event indicating success */
         CFE_EVS_SendEventWithAppID(CFE_SB_SUBSCRIPTION_RCVD_EID,CFE_EVS_EventType_DEBUG,CFE_SB.AppId,
-            "Subscription Rcvd:MsgId 0x%x on %s(%lu),app %s",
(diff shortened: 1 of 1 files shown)


jphickey:
Suggesting fast track here, as this fixes an issue in "main" when one enables debug events.

jphickey:
Correction - the original idea of just getting the pipe name broke many unit tests because there is (strangely) a debug event for getting a pipe name, and all unit test event counts became wrong again.  Simpler and better fix to just not print the name, only ID.

We should also remove the debug event on CFE_SB_GetPipeName success.  This generates a large number of clutter events, and makes UT fragile, and none of the other "get name" APIs have an event on success.

skliper (approved):

PR #1148, opened 2021-02-01
Fix #139, Do file writes in background

### LINKED ISSUE #139: TBL - Dump Table Registry Data Command Can Hog CPU (GSFC DCR 23031)

jphickey:
**Describe the contribution**
Implement a generic FS facility to perform file writes as a background job.  

Applications wanting to use this facility need to instantiate a state object (metadata) in global memory, and two callback APIs- one to get a data record, another to send events. 

The following file requests are changed to use this facility:
- ES ER Log dump
- SB Pipe Info
- SB Message Map
- SB Route Info
- TBL Registry Dump

Fixes #139

**Testing performed**
First built and ran "main" branch (unchanged) and issued all file write commands before change to get a baseline/reference copy.
Then re-built with this change applied and re-issued all file write commands.
Compared old files to new files - confirmed that the new files are correct (but see note below!).

**Expected behavior changes**
Files are written in the context of the ES background task.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
While examining the diffs between the old files and new files, I noticed that the queue depth in the Pipe Info was actually wrong in the original/reference data.  This was due to some mismatches between Pipe Info fields where names were getting crossed.

In order to fix this and avoid it from happening in the future - this changes the internal SB member names to be consistently named:
 - `MaxQueueDepth` for maximum depth at queue creation time (previously was `QueueDepth` or `Depth` depending on context)
 - `CurrentQueueDepth` for the running count (previously was `InUse` or `CurrentDepth` depending on context)
 - `PeakQueueDepth` for the highest watermark (previously was `PeakInUse` or `PeakDepth` depending on context)

In particular the `Depth` and `CurrentDepth` were not (previously) being propagated to the file correctly - as the names got crossed in the implementation.  This PR fixes it by making the names consistent.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_apps.h
+++ b/fsw/cfe-core/src/es/cfe_es_apps.h
@@ -249,9 +249,14 @@ bool CFE_ES_RunAppTableScan(uint32 ElapsedTime, void *Arg);
 bool CFE_ES_RunExceptionScan(uint32 ElapsedTime, void *Arg);
 
 /*
-** Check if ER log dump request is pending
-*/
-bool CFE_ES_RunERLogDump(uint32 ElapsedTime, void *Arg);
+ * Background file write data getter for ER log entry
+ */
+bool CFE_ES_BackgroundERLogFileDataGetter(void *Meta, uint32 RecordNum, void **Buffer, size_t *BufSize);
+
+/*
+ * Background file write event handler for ER log entry
+ */
+void CFE_ES_BackgroundERLogFileEventHandler(void *Meta, CFE_FS_FileWriteEvent_t Event, int32 Status, uint32 RecordNum, size_t BlockSize, size_t Position);
 
 /*
 ** Perform the requested control action for an application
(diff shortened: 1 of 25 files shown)


jphickey:
This needs a rebase after next baseline - change is currently in commit c6d9f27.  Pushing as draft to get review started.

astrogeco:
[CCB:2021-02-03](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-February-03-2021) APPROVED

astrogeco:
@jphickey is this ready?

jphickey:
Rebased to "main" - ready to merge now.

PR #1151, opened 2021-02-03
Fix #978, put task parameters into task record

### LINKED ISSUE #978: Move entry point address info from app table to task table
The ES global currently tracks task entry point in the app table, not task table.  As a result only the "main" task entry point of an app is tracked in this regard.  As child tasks are created, these all have different entry points, but they aren't stored, and therefore cannot be reported in the telemetry as reported via e.g. QUERY_ALL_TASKS.

Hold the actual runtime entry point address in the Task table rather than the App table.

**Additional context**
This would simplify the current task entry logic as it would only need to look at the task entry, not the app+task entry combination.
Still would need to hang onto the _configured_ (string version) of entry point for when an app is reloaded/restarted.

**Requester Info**

jphickey:
**Describe the contribution**
Encapsulate all parameters for apps and tasks into a structure object and clean up internal APIs to pass this object rather than individual parameters.  These parameters can then easily be stored with the relevant record in the internal table (tasks, apps, libs) and makes management easier - as code copies one object rather than many individual fields with (potentially) different names.

All parameters are stored in the record, which facilitates future telemetry generation/stats and/or for when an app might get restarted or reloaded in the future.

Notably this changes all task startup to go through the same basic routine (there is no longer a need for different accounting between main tasks and child tasks) and share the same common entry point.  

Fixes #978 

**Testing performed**
Build and sanity check CFE.  In particular ensure that all tasks + child tasks are starting and running normally.  Also checked that the app restart command is working as expected.  

Run all unit tests and confirm coverage.

**Expected behavior changes**
No public API change.  Internal APIs are simplified, internal data structures are more consistent, and more detail is stored with the relevant record (i.e. task record has all relevant task details, so when looking up task info one does not have to traverse to the app record which previously held some of this data).

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Using the common entry point for child tasks avoids a theoretical race condition where the child task might not have been fully accounted for in the global table at the time the user function started execution.  Previously this would directly invoke the user-supplied function immediately.  Now using the common entry point this delays until the task record is completely accounted for (linked back to app, etc) before calling the user function.  So functions such as `CFE_ES_GetAppID()` and others will work right from the beginning.

### DIFF
--- a/fsw/cfe-core/src/es/cfe_es_api.c
+++ b/fsw/cfe-core/src/es/cfe_es_api.c
@@ -1015,15 +1015,15 @@ int32 CFE_ES_GetAppInfo(CFE_ES_AppInfo_t *AppInfo, CFE_ES_AppId_t AppId)
        AppInfo->ResourceId = CFE_RESOURCEID_UNWRAP(AppId); /* make into a generic resource ID */
        AppInfo->Type = AppRecPtr->Type;
 
-       CFE_ES_CopyModuleBasicInfo(&AppRecPtr->StartParams.BasicInfo, AppInfo);
-       CFE_ES_CopyModuleStatusInfo(&AppRecPtr->ModuleInfo, AppInfo);
+       strncpy(AppInfo->Name, CFE_ES_AppRecordGetName(AppRecPtr), sizeof(AppInfo->Name)-1);
 
-       AppInfo->StackSize = AppRecPtr->StartParams.StackSize;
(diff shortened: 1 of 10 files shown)


jphickey:
This needs a rebase after next baseline.  Submitting as draft to get review started, actual change is in commit a0cda23.

astrogeco:
[CCB:2021-02-02](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-February-03-2021) APPROVED

jphickey:
Rebased to "main" - this is ready to go now.

PR #1154, opened 2021-02-04
Fix #1020, refactor SB buffer descriptor object

### LINKED ISSUE #1020: Questionable address adjustment in SB buffers may break alignment requirements
**Describe the bug**
Software bus message buffers should handle most restrictive alignment requirements for a message.  Address arithmetic used in manipulating buffers may break this alignment on some systems.  Even if not broken, a maintenance issue since an update to `CFE_SB_BufferD_t` could break things unexpectedly.

https://github.com/nasa/cFE/blob/32f3deeebf136ac427a6b779cb9e4854e73826f1/fsw/cfe-core/src/sb/cfe_sb_api.c#L1637-L1641
https://github.com/nasa/cFE/blob/32f3deeebf136ac427a6b779cb9e4854e73826f1/fsw/cfe-core/src/sb/cfe_sb_buf.c#L84-L88
https://github.com/nasa/cFE/blob/32f3deeebf136ac427a6b779cb9e4854e73826f1/fsw/cfe-core/src/sb/cfe_sb_buf.c#L119

**To Reproduce**
Not confirmed, but likely won't meet alignment requirement for a message with long double.

**Expected behavior**
Safer to use the real buffer type (instead of `void *`) in the descriptor along with offsetof to size the buffer correctly (CFE_SB_Msg_t for now, or maybe CFE_SB_Buffer_t from #1019)

**Code snips**
See above

**System observed on:**
NA - inspection

**Additional context**
#1019, #1009 

**Reporter Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
Refactor the SB buffer descriptor object (`CFE_SB_BufferD_t`) and simplify the zero-copy buffer paradigm.

Combine the zero-copy and the normal CFE buffer descriptor into a single unified CFE_SB_BufferD_t object.  This cleans up a bunch of extra logic related to zero-copy buffers, including the extra descriptor object.  The result is a simpler zero-copy design that is much less different from the standard (non-zero-copy) message path.

All message descriptor objects are now tracked in a list by SB, not just the zero-copy descriptors (for consistency - if any buffers need to be tracked, they should all be tracked).

This notably puts the buffer content as a _member_ within the descriptor, rather than a calculated pointer, so it will be aligned properly.

Fixes #1020 

**Testing performed**
Build and run all unit tests
Sanity check CFE

**Expected behavior changes**
All changes are internal to SB.  This does not affect API or behavior of any existing APIs (but see note)

It also ensures that zero-copy buffers (and the associated CFE_SB_TransmitBuffer API) behave as similarly as possible to the normal CFE_SB_TransmitMsg API.  Notably this corrects a minor issue where the MsgSendErrorCounter would get incremented if there were no subscribers, but only in the zero copy API.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This does not change public API in any way.  However now there is no extra descriptor for the "zero copy" buffers - they are all just buffer descriptors, regardless of whether they were allocated on the fly or pre-allocated by the application.

This means that the `CFE_SB_ZeroCopyHandle_t` value that is output from `CFE_SB_ZeroCopyGetPtr()` and others is now largely redundant.  It is equal to the buffer descriptor address, but it can be easily determined from the buffer content pointer too, so it is really not necessary to pass around two values for zero copy buffers.  This can be simplified in a future change - but that will change the public API.

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_buf.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_buf.c
@@ -39,6 +39,55 @@
 #include "cfe_es.h"
 #include "cfe_error.h"
 
+/*
+ * The actual message content of a SB Buffer Descriptor is the
+ * offset of the content member.  This will be auto-aligned by 
+ * the compiler according to the requirements of the machine.
+ */
+#define CFE_SB_BUFFERD_CONTENT_OFFSET   (offsetof(CFE_SB_BufferD_t,Content))
+
+
+/******************************************************************************
+ *
+ * Helper function to reset/clear the links on a list node (make empty)
+ */
+void CFE_SB_TrackingListReset(CFE_SB_BufferLink_t *Link)
+{
+    /* A singleton node/empty list points to itself */
+    Link->Prev = Link;
+    Link->Next = Link;
+}
+
+/******************************************************************************
+ *
+ * Helper function to remove a node from a tracking list
+ */
+void CFE_SB_TrackingListRemove(CFE_SB_BufferLink_t *Node)
+{
+    /* Remove from list */
+    Node->Prev->Next = Node->Next;
+    Node->Next->Prev = Node->Prev;
+
+    /* The node is now a singleton */
+    CFE_SB_TrackingListReset(Node);
+}
+
+/******************************************************************************
+ *
+ * Helper function to add a node to a tracking list
+ */
+void CFE_SB_TrackingListAdd(CFE_SB_BufferLink_t *List, CFE_SB_BufferLink_t *Node)
+{
+    /* Connect this node to the list at "prev" position (tail) */
+    Node->Prev = List->Prev;
+    Node->Next = List;
+
+    /* Connect list nodes to this node */
+    Node->Prev->Next = Node;
+    Node->Next->Prev = Node;
+}
+
+
 /******************************************************************************
 **  Function:   CFE_SB_GetBufferFromPool()
 **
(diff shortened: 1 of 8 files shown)


jphickey:
Like others this will need a rebase when the next mainline is done, actual change is in commit 365c095

fsw/cfe-core/src/sb/cfe_sb_buf.c
```diff
@@ -48,82 +97,60 @@
...
-
-    /* increment actual msg buffer ptr beyond the descriptor */
-    address += sizeof(CFE_SB_BufferD_t);
+    } /* end if */
 
     /* Initialize the buffer descriptor structure. */
-    bd->MsgId     = MsgId;
-    bd->UseCount  = 1;
-    bd->Size      = Size;
-    bd->Buffer    = (CFE_SB_Buffer_t *)address;
```
jbohren-hbr, line 94: Good to see this updated. This line in the current `main` branch increases alignment requirements on ARM32 when going from the `uint8*` to `CFE_SB_Buffer_t*`.

skliper:
@jphickey can you rebase this one? Blocking tests on vxworks...

astrogeco:
[CCB 2021-02-17](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-February-17-2021) APPROVED
- Adds to `CFE_SB_BUFFER_t`
- Removes `CFE_SB_ZeroCopyD_t`
- Consolidates information about buffer and descriptor without distinction between a "zero copy" and a "regular" one

jphickey:
Rebased to the new `main` branch ... should be OK to merge now.

PR #1156, opened 2021-02-05
Fix #981, implement better filename parser

### LINKED ISSUE #981: Update startup processing to optionally use module suffix
Startup scripts currently need to be modified across systems that have different module suffixes.

#941 makes the module suffix available in software.  If there is no suffix specified in the startup script, append it (backwards compatible).  This would make startup scripts more portable.

None

**Additional context**
#583 - table implementation would also benefit from not needing the suffix

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**

Implemements a new function in FS to parse an input file name from a user - `CFE_FS_ParseInputFileName()`
    
The new function produces fully-qualified output where the input may omit either a pathname or an extension.  These items will be  added from the specified defaults, if missing, and a complete fully-qualified filename will be output.
    
If the input is already a fully qualified filename, then the output is the same as the input (basically a copy).
    
This initially is used to provide better cross-platform startup  script processing, where module suffix may differ across platforms.  Only the "basename" of the filename needs to be specified in the startup script - everything else can come from defaults.

Fixes #981 

**Testing performed**
Build and sanity test CFE.
Check startup with all combinations of filename in script - fully qualified, basename only.
Check startup with new default startup script that has basename only - works on both Linux (.so) and RTEMS (.obj) with no change.

**Expected behavior changes**
Easier switching between platforms - startup script does not need to change to account for the different suffix that different platforms use.

**System(s) tested on**
Ubuntu 20.04 (native)
RTEMS 4.11 (i686, QEMU)


jphickey:
This also needs a rebase when next baseline is ready.  Commit for pre-review is in cc5c9689.

skliper (approved):

astrogeco:
[CCB 2021-02-17](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-February-17-2021)

jphickey:
This should be good to merge now.

PR #1168, opened 2021-02-16
Fix #735, add comment if null terminated or not.

zanzaben:
**Describe the contribution**
Fixes #735
added comments for if lengths included null terminals or not. 

**Testing performed**
Build and run unit test

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04

**Contributor Info - All information REQUIRED for consideration of pull request**
Alex Campbell GSFC

PR #1158, opened 2021-02-05
Fix #790, allow directories and symlinks in FILELIST.

jphickey:
**Describe the contribution**
Checks a name-based subdirectory under ${MISSION_DEFS} for files listed in the FILELIST for that target.  If file is a symlink, the link should be followed so the correct content is installed, not a symlink.

Fixes #790 

**Testing performed**
Created a symlink in a target name subdirectory (see additional context) and confirm that the correct file content is installed.

**Expected behavior changes**
Items in FILELIST can be in a target name directory, and may be symlinks.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This also adds another "NOTE" message to indicate exactly what file is being used, after all lookup/de-mapping logic is done, e.g. I made a symlink to test this out:
   
    $ cd sample_defs
    $ mv cpu1_cfe_es_startup.scr test_startup.scr
    $ mkdir cpu1
    $ cd cpu1
    $ ln -s ../test_startup.scr cfe_es_startup.scr
    $ cd ../..
    $ make prep
    <<snipped>>
    NOTE: Selected /home/joe/code/cfecfs/github/sample_defs/test_startup.scr as source for cfe_es_startup.scr on cpu1

This is similar to all the other informational messages for table sources and app sources indicating where they were found.

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #1171, opened 2021-02-17
cFE Integration candidate: 2021-02-23

### LINKED ISSUE #139: TBL - Dump Table Registry Data Command Can Hog CPU (GSFC DCR 23031)

### LINKED ISSUE #735: document configs to indicate that MAX string sizes do not include null-terminator
There is a laundry list of string size definitions in sample_mission_cfg.h, cpu1_platform_cfg.h, and default_osconfig.cmake that should be documented as to whether the setting includes a byte for a null-terminator (so say `CFE_MISSION_TBL_MAX_NAME_LENGTH` is defined as 16, does that mean I must limit my table name to 15 characters plus the null-terminator?)

This is related to #730 .

**Requester Info**
Christopher.D.Knight@nasa.gov

### LINKED ISSUE #790: Allow target name subdirectory for install files and handle symlinks
Just need to add to the paths of locations for the install list from TGT<N>_FILELIST the possibility of target_defs/<TGT_NAME>/<FILENAME>

Also, if the file is a symlink, it needs to be followed to its source but copied as the name in FILELIST suggests.

```
    foreach(INSTFILE ${TGT${TGTID}_FILELIST})
      if(EXISTS ${MISSION_DEFS}/${TGTNAME}/${INSTFILE})
        set(FILESRC ${MISSION_DEFS}/${TGTNAME}/${INSTFILE})
      elseif(EXISTS ${MISSION_DEFS}/${TGTNAME}_${INSTFILE})
        set(FILESRC ${MISSION_DEFS}/${TGTNAME}_${INSTFILE})
      elseif(EXISTS ${MISSION_DEFS}/${INSTFILE})
        set(FILESRC ${MISSION_DEFS}/${INSTFILE})
      else()
        set(FILESRC)
      endif()
      if (FILESRC)
        get_filename_component(_resolvedFile "${FILESRC}" REALPATH)
        install(FILES ${_resolvedFile} DESTINATION ${TGT_NAME}/${INSTALL_SUBDIR} RENAME ${INSTFILE})
      else(FILESRC)
        message("WARNING: Install file ${INSTFILE} for ${TGTNAME} not found")
      endif (FILESRC)
    endforeach(INSTFILE ${TGT${TGTID}_FILELIST})
```

Something like that maybe?

**Additional context**
Add any other context about the feature request here.

**Requester Info**
Full name and company/organization if applicable

### LINKED ISSUE #978: Move entry point address info from app table to task table

### LINKED ISSUE #1160: Default file name for Task Info is too long
**Describe the bug**
The default filename for the task info file is defined here:

https://github.com/nasa/cFE/blob/fcaa1d0c6b9e8244b071b8cafc765078680e24b4/cmake/sample_defs/cpu1_platform_cfg.h#L924

The filename portion of the string - "cfe_es_task_info.log" - is exactly 20 chars, and `OSAL_CONFIG_MAX_FILE_NAME` is also 20 chars, so it fails the max length test - because it needs to be less than the max for the NUL char.

**To Reproduce**
Build with default config, issue CFE_ES_QUERY_ALL_TASKS_CC command with no filename - which causes it to use default.
Observe error about failure to create file - error code `OS_FS_ERR_NAME_TOO_LONG`. 

**Expected behavior**
Defaults should work.

**System observed on:**
Ubuntu 20.04

**Additional context**
The prefix `cfe_es_` is 7 chars by itself.  A simple fix would be to trim this back to just `cfe_`.  Would recommend changing all the default filenames for consistency.  ER log is already just `cfe_erlog.log` (no es).

**Reporter Info**

### LINKED ISSUE #1172: Replace UT_Stub_CheckForceFail with UT_Stub_CheckDefaultReturnValue
See nasa/osal#737

Update to use new name

None

**Additional context**
None

**Requester Info**
Jacob Hageman - NASA/GSFC

astrogeco:
**Describe the contribution**
Fix #139, Do file writes in background
Fix #978, put task parameters into task record
Fix #790, allow directories and symlinks in FILELIST. 
Fix #735, add comment if null terminated or not.

Fix #1160, Shorten task info default filename 
Fix #1172, Update to use Ut_Stub_CheckDefaultReturnValue API  

**Testing performed**
See <https://github.com/nasa/cFS/pull/193/checks>

**Expected behavior changes**
PR #1148 - Implements a generic FS facility to perform file writes as a background job. Applications wanting to use this need to instantiate a state object (metadata) in global memory and two callback APIs, one to get a data record, another to send events. The following file requests are changed to use this facility:

- ES ER Log dump
- SB Pipe Info
- SB Message Map
- SB Route Info
- TBL Registry Dump

Changes the internal SB member names for consistency and thus fixes propagation of `Depth` and `CurrentDepth` into files:

- `MaxQueueDepth` for maximum depth at queue creation time (previously was QueueDepth or Depth depending on context)
- `CurrentQueueDepth` for the running count (previously was InUse or CurrentDepth depending on context)
- `PeakQueueDepth` for the highest "watermark" (previously was PeakInUse or PeakDepth depending on context)

PR #1151 - No public API change. Encapsulates all parameters for apps and tasks into a structure object. Cleans up internal APIs to pass this new object rather than individual parameters. Adds details to the relevant record (i.e. a task record has all relevant task details) which eliminates the need to traverse the app record to find some data.

PR #1158 - Enables items in `FILELIST` to be in a target name directory as well as symlinks. `arch_build.cmake` now checks a name-based subdirectory under `${MISSION_DEFS}` for files listed in the `FILELIST` for that target. If file is a symlink, the link should be followed so the correct content is installed, not a symlink.

PR #1168 - Mo impacts to behavior. Adds documentation on  inclusion presence of null terminators for length parameters.

PR #1175 - Shortened `CFE_PLATFORM_ES_DEFAULT_TASK_LOG_FILE` name so it is within the `OSAL_MAX_FILE_NAME` size limit. Will now output task info to default filename if no filename is provided in command.

PR #1173  - Replaces `UT_Stub_SetForceFail` with `UT_Stub_SetDefaultReturnValue`. No behavior change.

**System(s) tested on**
Ubuntu 18.04

**Additional context**
Part of <https://github.com/nasa/cFS/pull/193>

**Third party code**
None

@jphickey 
@zanzaben


astrogeco:
CodeQL fails due to dependency on osal, cFS bundle level PR checks are succesful

PR #1174, opened 2021-02-17
Fix #1170, refactor target config objects

### LINKED ISSUE #352: Add PSP_VERSION creation as part of version.cmake
PSP build is unique in that it doesn't follow the pattern of apps, cfe, and osal for producing version information (as part of version.cmake).  Consistent build framework is preferred.

### LINKED ISSUE #1170: Refactor "target_config.c" logic to avoid #include of data fragments
The content of the various CONFIGDATA global data structures is dynamically generated from a combination of build scripts, some which run at prep/generation time and some which run at build time.  

The pattern currently used here is that the scripts generate "fragment" data files that only contain the dynamic content, without any C declarations or any other syntax.  This is then paired/combined with the `target_config.c` file which is a regular version-controlled file that has the basic structure and declarations (i.e. all the non-dynamic bits) and uses `#include` to pull in the data fragments where needed (i.e. a "fill-in-the-blank" approach).  

This works well but there are some rules against using the `#include` directive with anything other than an normal/conventional fully formed C header file (i.e. it has `.h` extension, fully self contained, syntactically correct, standalone, include-able by any C source file at will).   So a "dynamic data fragment" file which is implemented here doesn't meet the criteria.

Have the build system generate fully-formed source files before passing to C compiler in some way.

Currently evaluating/considering several approaches or combination thereof.

1. Put more of the non-dynamic "C" content (structure) into CMake `configure_file` templates, generating a separate but fully-resolved .c source for each object that has dynamic content.
2. Pass the file through another explicit macro-processing step (such as C preprocessor or m4) before compiling (basically introduce another layer of indirection) to generate a final file.

Either way the result should be a fully formed .c/.h file without any `#include` in its body - as all data items should be "filled in" already - which is then built and linked as the current file is.

**Requester Info**

jphickey:
**Describe the contribution**
Rework the dynamic content so it is generated entirely via a CMake "configure_file()" command into a C source file (.c) that
can be built as normal.

This removes the need for inline `#include` statements to pull in data fragments to fill in the data.

Fixes #1170 
Fixes #352 

**Testing performed**
Build and run all unit tests, sanity test CFE, confirm versions reported as expected.

**Expected behavior changes**
This reports multiple event "91" (version info) - one event for each statically linked code module.  This now also includes supplemental modules such as "msg" and "sbr" (these were not reported at all before).

This also moves reporting of the "mission config" name from event 91 to event 92 (build info) as it is not really version info.  Ideally event 91 should be strictly for source control revision info.  Any other environment/config info should be in the other events.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This also attempts to better differentiate/clarify the two different versions that are reported.

The first version is administratively-assigned, traditional "semantic" version (major/minor/rev plus dev build number).  This is updated manually and it is what is reported in CFE event 2 during start up as follows:

```
EVS Port1 66/1/CFE_ES 2: cFS Versions: cfe v6.8.0-rc1+dev348, osal v5.1.0-rc1+dev262, psp v1.5.0-rc158. cFE chksm 65346
```

The second version is the automatically-obtained "source" version which is gained by introspection of the source code at compile time.  This is reported in CFE event 91, as follows:

```
EVS Port1 66/1/CFE_ES 91: Version Info: Component cfe-core, version git:v6.8.0-rc2-346-g3aa80ed0
EVS Port1 66/1/CFE_ES 91: Version Info: Component osal, version git:v6.0.0-rc1-1-g7576fed6
EVS Port1 66/1/CFE_ES 91: Version Info: Component psp, version git:v1.5.0-rc1-61-gf3dae6c-dirty
EVS Port1 66/1/CFE_ES 91: Version Info: Component msg, version git:v6.8.0-rc2-346-g3aa80ed0
EVS Port1 66/1/CFE_ES 91: Version Info: Component sbr, version git:v6.8.0-rc2-346-g3aa80ed0
EVS Port1 66/1/CFE_ES 91: Version Info: Component resourceid, version git:v6.8.0-rc2-346-g3aa80ed0
```

Important to note that instead of squeezing this all into a _single_ event 91 as had been done before, this sends a separate event for each component.  This means that _all_ components can be done (note this includes msg/sbr/resourceid now, these were not reported before) as well as making it possible to split "cfe-core" into its components as well.  It will scale more easily to any number of modular components.


jphickey:
Created as a draft because this still needs a few unit test updates to get full coverage on the modifications, but before I do that I wanted to get some other eyes on it to make sure the general concept is good to go.

jphickey:
Also worth a note - there was previous concern as to being potentially too tightly coupled with "git" for the version control info.  In this change I have also isolated the version control actions to a separate script (`generate_git_module_version.cmake`) and also added the capability for a user to provide their own script that does this - so they can call some other tool like SVN or adapt to however they like to do branching/tagging so it provides a useful result.  But the default one works fine for the git submodule-based bundle out of the box.

skliper (approved):

jphickey:
Rebased to "main" ... but will still hold this one as draft until consensus is reached in nasa/cFS#200 since this will also affected by that decision.

skliper (approved):

skliper:
@astrogeco I'm fine with fast-tracking this one if it helps... or review the WIP from Joe as long as we can get the WIP closed out and merged next cycle.

astrogeco:
[CCB 2021-03-03](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-03-2021) APPROVED
- Using a `.c` file to avoid developers including a header in different places since these defines are meant to be used in a very specific way.
- Splits version.cmake into two files `generate_build_env.cmake` and `generate_git_module_version.cmake` to accommodate implementations that DO NOT use git

Issue #1166, opened 2021-02-15
Question: where is the cFS Deployment Guide?


smalltalker55:
https://github.com/nasa/cFE/blob/main/cmake/README.md says
"For a more in-depth guide and specific examples, see Appendix A of the cFS Deployment Guide."   I have searched for the cFS Deployment Guide and cannot find it.   Has the name changed?   Where is it now?

PR #1193, opened 2021-03-01
Fix #1033, remove false child tasks statement.

### LINKED ISSUE #1033: Update cFE Application Developers Guide (2 comments)
**Describe the bug**

The cFE Application Developers Guide section 5.5 references the cFE Deployment Guide.  Please include the deployment guide or remove the reference.

The cFE Application Developers Guide section 4.1.3 states:
"Child Tasks can only execute at a priority equal to or less than the priority of the Application's Main Task." which several folks have said is not true.  It's not true in CFE 6.6.x.

**Reporter Info**
Dan Berry / NASA GSFC

zanzaben:
**Describe the contribution**
Fixes #1033
removes false child tasks statement. 

**Testing performed**
Build and run unit test

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04

**Additional context**
References to cFE Deployment Guide were fixed by #1166

Alex Campbell GSFC

### DIFF
--- a/docs/cFE Application Developers Guide.md
+++ b/docs/cFE Application Developers Guide.md
@@ -488,12 +488,12 @@ deciding on whether to create multiple Applications versus a single
 Application with multiple Tasks, the Application Developer should keep
 in mind these facts:
 
--   Child Tasks can only execute at a priority equal to or less than the
-    priority of the Application's Main Task.
+-   When the Application exits it is the responsiabilty of the 
+    Main Task to safely stop all of its Child Tasks.
 
 -   If the Main Task of an Application is stopped, either through
     detection of an exception or via command, all Child Tasks are also
-    stopped.
+    forcibly stopped in an unsafe manner. 
 
 Child Tasks can be useful in both "Software Only" and "Hardware Servicing"
 applications.


docs/cFE Application Developers Guide.md
```diff
@@ -429,9 +429,6 @@ deciding on whether to create multiple Applications versus a single
 Application with multiple Tasks, the Application Developer should keep
 in mind these facts:
 
--   Child Tasks can only execute at a priority equal to or less than the
-    priority of the Application's Main Task.
-
 -   If the Main Task of an Application is stopped, either through
     detection of an exception or via command, all Child Tasks are also
     stopped.
```
skliper, line 434: Is this comment valid?  Maybe there should be a more general comment related to the implementations?  @jphickey - recommendations?

jphickey: Its not totally _invalid_, it just glosses over the details....

If an exception occurs then only the specific task that triggered the exception is immediately stopped - Other tasks will keep running.

However, such an event will put the ES app record into an alternate exception state - and after a time delay (typically ~5 seconds) the ES will clean up the app in its background job.  As part of this clean up, all child tasks and the main task will be removed if they have not already exited.

So _eventually_ all tasks will be stopped, but it is not quite as that statement implies (the wording suggests/implies this happens at the same time, but it does not happen at the same time).

skliper: I think it's the generic language that bugs me the most since there's subtilties that don't seem to be sufficiently addressed by the original comment that could lead to incorrect assumptions.  Maybe it would help to start with "When ES stops or cleans up an application...(add your details here)"

zanzaben: If the main app is stopped via command, do the child tasks stop at the same time or do they still sit around until the ES clean up happens?

jphickey: @zanzaben - there really is "main task stopped by command" - the stop command being referred to here works on an App basis, not a task basis, and it's really a voluntary request, not a forced administrative action (at least at first).  

When this "stop" command is processed, it posts a request for the app to voluntarily stop itself.  The next time the main task calls CFE_ES_RunLoop() this function will return false (thereby causing its while loop to exit) and then the app _should_ do an orderly shutdown, closing any resources it was holding and removing its child tasks, and then exit.  This "cooperative" approach is the preferred way stop an app.  

However, for apps that are not cooperative, there is a backup plan: If it doesn't voluntarily shut itself down within the ES timeout then ES will forcibly kill all the tasks associated with the app, and also forcibly free all the other resources that CFE/OSAL is responsible for tracking that appear to be "owned" by that app.  This isn't ideal; it doesn't get everything (it can't) and there is no way to know if we are leaving something in an indeterminate state due to forcibly killing the tasks.

Also - in all cases child tasks continue running until this happens (either voluntary shutdown or delete by ES).  It is not synchronous/same time in any case.

jphickey: IMO - if ES ever has to forcibly clean up an app in the "unclean" (back-up) manner I described above - it implies an indeterminate state of the overall system.  There could be e.g. locks left in a wrong state, and there is no way to know for sure the way things were left.  The operator really should do a processor reset to bring it back to a fully known/stable state again.

zanzaben: @jphickey @skliper How about this new wording.

PR #1197, opened 2021-03-01
Fix #1189, correct compiler warnings

### LINKED ISSUE #1189: Possible uninitialized variables - compiler warnings (release build, CentOS 7)
Various warnings on CentOS 7 when BUILDTYPE=release:
```
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c: In function ‘CFE_SB_TransmitMsg’:
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1433:16: error: ‘RouteId’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
         Status = CFE_SB_TransmitBufferFull(BufDscPtr, RouteId, MsgId);
                ^
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c: In function ‘CFE_SB_SendMsg’:
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1433:16: error: ‘RouteId’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1395:25: note: ‘RouteId’ was declared here
     CFE_SBR_RouteId_t   RouteId;
                         ^
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c: In function ‘CFE_SB_PassMsg’:
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1433:16: error: ‘RouteId’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
         Status = CFE_SB_TransmitBufferFull(BufDscPtr, RouteId, MsgId);
                ^
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1395:25: note: ‘RouteId’ was declared here
     CFE_SBR_RouteId_t   RouteId;
                         ^
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c: In function ‘CFE_SB_RcvMsg’:
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:2045:43: error: ‘RcvStatus’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
                 CFE_EVS_SendEventWithAppID(CFE_SB_Q_RD_ERR_EID,CFE_EVS_EventType_ERROR,CFE_SB_Global.AppId,
                                           ^
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:1823:28: note: ‘RcvStatus’ was declared here
     int32                  RcvStatus;
                            ^
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c: In function ‘CFE_SB_ReceiveBuffer’:
/home/jhageman/cFS/cFS-GitHub/cfe/fsw/cfe-core/src/sb/cfe_sb_api.c:2045:43: error: ‘RcvStatus’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
                 CFE_EVS_SendEventWithAppID(CFE_SB_Q_RD_ERR_EID,CFE_EVS_EventType_ERROR,CFE_SB_Global.AppId,
                                           ^
```

Initialize variables where needed

None

**Additional context**
Compiler warnings, vs static analysis warnings reported in #1185

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
Corrects some (false) warnings about use before init.
Also corrects an alignment warning and removes unneeded union type (union type is not the best way to deal with this situation).

Fixes #1189 

**Testing performed**
Build and sanity check CFE for Raspberry Pi (older GCC w/strict alignment) and native (64-bit)
Run all unit tests

**Expected behavior changes**
None, just compiler warning fixes

**System(s) tested on**
Ubuntu 20.04 (native)
Raspbian (cross compiled; gcc 4.9.3)

### DIFF
--- a/fsw/cfe-core/src/sb/cfe_sb_api.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_api.c
@@ -1397,6 +1397,7 @@ int32  CFE_SB_TransmitMsg(CFE_MSG_Message_t *MsgPtr, bool IncrementSequenceCount
 
     PendingEventID = 0;
     BufDscPtr = NULL;
+    RouteId = CFE_SBR_INVALID_ROUTE_ID;
 
     Status = CFE_SB_TransmitMsgValidate(MsgPtr, &MsgId, &Size, &RouteId);
 
@@ -1886,6 +1887,7 @@ int32  CFE_SB_ReceiveBuffer(CFE_SB_Buffer_t **BufPtr,
     BufDscPtr = NULL;
     DestPtr = NULL;
     BufDscSize = 0;
+    RcvStatus = OS_SUCCESS;
     
     /*
      * Check input args and see if any are bad, which require
--- a/fsw/cfe-core/src/inc/private/cfe_sbr.h
+++ b/fsw/cfe-core/src/inc/private/cfe_sbr.h
@@ -38,6 +38,13 @@
 #include "cfe_msg_typedefs.h"
 #include "cfe_platform_cfg.h"
 
+/*
+ * Macro Definitions
+ */
+
+/** \brief Invalid route id */
+#define CFE_SBR_INVALID_ROUTE_ID ((CFE_SBR_RouteId_t) {.RouteId = 0})
+
 /******************************************************************************
  * Type Definitions
  */
--- a/fsw/cfe-core/src/sb/cfe_sb_priv.c
+++ b/fsw/cfe-core/src/sb/cfe_sb_priv.c
@@ -536,15 +536,7 @@ void CFE_SB_RemoveDestNode(CFE_SBR_RouteId_t RouteId, CFE_SB_DestinationD_t *Nod
 int32 CFE_SB_ZeroCopyReleaseAppId(CFE_ES_AppId_t AppId)
 {
     CFE_SB_BufferLink_t *NextLink;
-
-    /* Using a union type permits promotion of CFE_SB_BufferLink_t* to CFE_SB_BufferD_t*
-     * without violating aliasing rules or alignment rules.  Note that the first element
-     * of CFE_SB_BufferD_t is a CFE_SB_BufferLink_t, which makes this possible. */
-    union LocalBufDesc
-    {
-        CFE_SB_BufferLink_t As_Link;
-        CFE_SB_BufferD_t    As_Dsc;
-    } *BufPtr;
+    CFE_SB_BufferD_t    *DscPtr;
 
     /*
      * First go through the "ZeroCopy" tracking list and find all nodes
@@ -560,16 +552,18 @@ int32 CFE_SB_ZeroCopyReleaseAppId(CFE_ES_AppId_t AppId)
         while(!CFE_SB_TrackingListIsEnd(&CFE_SB_Global.ZeroCopyList, NextLink))
         {
             /* Get buffer descriptor pointer */
-            BufPtr = (union LocalBufDesc*)NextLink;
+            /* NOTE: casting via void* here rather than CFE_SB_BufferD_t* avoids a false
+             * alignment warning on platforms with strict alignment requirements */
+            DscPtr = (void *)NextLink;
 
             /* Read the next link now in case this node gets moved */
             NextLink = CFE_SB_TrackingListGetNext(NextLink);
 
             /* Check if it is a zero-copy buffer owned by this app */
-            if (CFE_RESOURCEID_TEST_EQUAL(BufPtr->As_Dsc.AppId, AppId))
+            if (CFE_RESOURCEID_TEST_EQUAL(DscPtr->AppId, AppId))
             {
                 /* If so, decrement the use count as the app has now gone away */
-                CFE_SB_DecrBufUseCnt(&BufPtr->As_Dsc);
+                CFE_SB_DecrBufUseCnt(DscPtr);
             }
         }
 
--- a/modules/sbr/private_inc/cfe_sbr_priv.h
+++ b/modules/sbr/private_inc/cfe_sbr_priv.h
@@ -31,12 +31,6 @@
  */
 #include "private/cfe_sbr.h"
 
-/*
- * Macro Definitions
- */
-
-/** \brief Invalid route id */
-#define CFE_SBR_INVALID_ROUTE_ID ((CFE_SBR_RouteId_t) {.RouteId = 0})
 
 /******************************************************************************
  * Function prototypes


jphickey:
Also confirmed that this builds cleanly on CentOS 7.9 with "gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)" and BUILDTYPE=release

fsw/cfe-core/src/sb/cfe_sb_api.c
```diff
@@ -1397,6 +1397,7 @@ int32  CFE_SB_TransmitMsg(CFE_MSG_Message_t *MsgPtr, bool IncrementSequenceCount
 
     PendingEventID = 0;
     BufDscPtr = NULL;
+    RouteId = CFE_SBR_INVALID_ROUTE_ID;
```
skliper, line 1400: What about having CFE_SB_TransmitMsgValidate actually set RouteId (actually done by the underlying GetRouteId call)?  Then it fixes everywhere the GetRouteId call is used.

jphickey: We can do that too, but I see it as a separate problem ...

1. A caller should never rely on/use output values from a function that returned non-success
2. A function should always set its output buffers to something, even when it returns non-success

I see (1) as the main issue, and (2) as the backup/failsafe.  I have no objection to doing both, I just see it as a secondary concern.  I can add it though.

skliper: In virtually every case I've seen that the static analyzer or compiler warning for possible uninitialized variable comes up 1 is met by the caller by either returning or checking status prior to using what is returned.  2 then is both good practices and squashes the warning.  Is initializing the variable in this case really required where 1 is already satisfied by logic?

This seems even more useless when the value being initialized is not checked explicitly before used, and would cascade into other issues down the line if there WAS a possible way to get to that code.  Basically, if you are checking for status ==  SUCCESS but depend on RouteId be valid why not check instead (or in addition if both cases need to be true) for RouteId being valid (that also squashes the warning).

skliper: Note for the case where the code IS checking for RouteId to be valid, I definitely agree explicit initialization is the way to go.  I just struggle with justifying initialization if it's never accessed based on failure logic.  Feels cleaner to me (and would be easier for me to maintain) when initialization is done locally when it matters (and squash errors via setting outputs in the underlying APIs when it doesn't matter).

skliper (approved):

fsw/cfe-core/src/sb/cfe_sb_api.c
jphickey: @skliper I added this in commit bf8f481 however I have my concerns/reservations - there is a chance that the original passed-in buffers themselves are NULL, where the status code becomes `CFE_SB_BAD_ARGUMENT` ... this will then try to write a value to the buffer.  Could add NULL checks on each, but then this is just getting overly complicated at that point.

skliper: Good point, I was thinking it would make sense to set them "inline" where it made sense (and really the getter calls could handle most of it, other then the elements not set yet).  Although I did just hit approve.

skliper: Alternatively just return for the NULL check failure, and initialize them to known values right after that?  Then you aren't overwritting anything that may have succeeded...

skliper: Or as an else case on the null check

skliper: > ...require substantial rework...

It already Is it really that big of an ask?  It checks for one pointer for NULL, seems like either all (and setting defaults only if good) or none would make more sense.  Either they can never be NULL or should be checked, no?  I'm fine with either approach as an internal function (and really don't have a real problem with your current PR), but just checking one pointer seems like a strange pattern to employ.  In the end I don't really care, but I really don't understand the pushback.  Could you provide background for your opposition?

skliper: Or if:
https://github.com/nasa/cFE/blob/bf8f481e2d87c13e59e2a32827f24dbd13c27394/fsw/cfe-core/src/sb/cfe_sb_api.c#L1536-L1541

Is really the sweet spot for checking, why not just add the else and set the outputs to defaults here instead of at the end?  Then it avoids setting them for the `CFE_SB_BAD_ARGUMENT` case.

jphickey: I mean, anything is possible, if you want to do that.  Each output _could_ be individually checked for NULL and assigned a default if not null.  If it helps meet a checkbox on some review procedure, that's well and good - but it doesn't make sense in the overall design of the software, it is just unnecessary extra complexity for no actual benefit (and cpu cycles burnt in code that actually does execute rather frequently in CFE)

I'm just taking a more holistic view of the software.  This is an internal helper function, called only from the two public "TransmitMsg" variants, in a specific scenario, to consolidate the common logic.

The only reason that `MsgPtr` is explicitly checked is because of the fact that (for some reason?) public APIs are all supposed to accept and detect a NULL pointer input.  This assumes the role of checking that input.

For the other 3 - those are outputs to stack variables.  They are stack variables in the parent - never NULL.  This is not a public API - they aren't going to be randomly called with arbitrary values in a general-purpose sense.  Again, the role of this function is to consolidate common code between CFE_SB_TransmitMsg and CFE_SB_TransmitBuffer, initializing local stack variables on behalf of the caller, such that the code is not repeated between the two.

By definition, any NULL pointer check here is impossible to get at runtime - thereby making it dead code.  The way I see it this is very similar to the recent case in OSAL where a check for an impossible condition was removed due to the code before it.  This is the exact same thing, just in a helper function.  The helper is not invoked with NULL args, so any check for NULL args is immediately dead/unreachable code.

jphickey:
Note - although I originally requested a fast track on this one - that was only for the initial commit (54adb9fb) that directly fixed the compiler warnings without really changing other code.

Perhaps we should fast track the original change so that main compiles cleanly, then create a new ticket to argue over the virtues (or not) of adding complexity to internal helper functions?

fsw/cfe-core/src/sb/cfe_sb_api.c
skliper: It was you're comment about "there is a chance that the original passed-in buffers themselves are NULL, where the status code becomes CFE_SB_BAD_ARGUMENT" that triggered my comments... so can they be NULL or not?  If not, the current implementation is fine (I already approved that one when I first saw it.  If they can be NULL, then check for it.  Alternatively, just set the passed in pointers as the else of the MsgPtr check.  Are all 3 of these possible approaches really not acceptable and the only option is to not initialize these values? I still don't follow why there is any opposition?  Maybe contact me for a chat?

jphickey: Sorry about that - that was only my initial knee-jerk before looking closer and seeing that the function really only does need to handle MsgPtr being null and not the others.  To allow the others to be null would be pointless.

jphickey:
I reverted this PR back to its original change - just the real warning fixes - no change to actual implementation of CFE_SB_TransmitMsgValidate().  

I recommend merging this as-is - it fixes the warning in #1189 without breaking anything and has minimal impact.

I created a new issue #1198 to further the discussion about whether internal helper functions need to separately validate their inputs (e.g. handle NULLs) or force outputs.  This is a more complex topic and probably not something to implement in a fast bug fix.

skliper (approved):

fsw/cfe-core/src/sb/cfe_sb_api.c
skliper: Copy.  And I'm fine w/ how it's currently just checking MsgPtr (probably would benefit from a comment to that affect, but implementation is fine) based on the background you provided.

astrogeco:
[CCB 2021-03-03](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-03-2021) APPROVED

jphickey:
For reference here is the link to stack overflow discussion on the topic: https://stackoverflow.com/questions/66427774/best-practices-for-object-oriented-patterns-with-strict-aliasing-and-strict-alig

The key point is that the union-of-pointers or pointer-to-pointer approach violates strict aliasing not on the data itself but the pointers.  A derived struct may alias a base struct and vice versa, but _pointers_ to those structs are a distinct type and the aliasing rules apply to them too.  When the compiler complained of strict aliasing rule violations, it wasn't about the data object, it was about the pointer to the data object.

PR #1210, opened 2021-03-04
Fix #1207, Add wrapper targets to simplify app builds

### LINKED ISSUE #1207: Update/resurrect the add_unit_test_lib/add_unit_test_exe functions
Currently the `arch_build.cmake` functions provide a routine to easily define a CFE app: `add_cfe_app()` .  This makes it easy for app developers to create an app target that uses all the right compiler definitions, include dirs and link/interfaces libraries.  Furthermore, having this wrapper in CFE helped future-proof so that apps didn't need to change this call when #972 was implemented.

The problem is that we didn't employ the same tactic for unit tests.  For these, all the special logic for adding flags/includes is basically put into the CMakeLists.txt in each app where the UT targets are defined.

The `arch_build.cmake` does have a `add_unit_test_lib` and `add_unit_test_exe` function that was originally intended for this purpose, but these were never really used.  We should implement the basic UT logic here and change sample app/lib to use them.

**Additional context**
Currently, the sample_lib/sample_app unit tests break when merging with #1203 because of the different paths and the transition toward interface libraries.  Rather than just updating these UT builds to work again, using these functions would help avoid this issue in the future.

**Requester Info**

jphickey:
**Describe the contribution**
Adds the following functions to simplify tracking of dependencies and use of interface libraries provided by CFE:

add_cfe_app_dependency - dependency that applies to both FSW and coverage test
add_cfe_coverage_dependency - dependency that applies only to coverage test (i.e. for overrides)
add_cfe_coverage_test - creates object + executable targets for coverage test
add_cfe_coverage_unit_include - add extra include path to unit build (i.e. for overrides)
add_cfe_coverage_stubs - add a stub library for coverage testing other modules

The goal is to reduce/eliminate the need to directly specify any include paths or compile definitions for dependencies, relying
on target properties to do this instead.

Fixes #1207

**Testing performed**
Build and run all CFE unit tests (this only affects UT, not FSW)

**Expected behavior changes**
None, just adding the routines for now, sample app/lib will be transitioned to use them in subsequent PR

**System(s) tested on**
Ubuntu 20.04

**Additional context**
This is needed as part of preparation for directory reorganization in #1203.  These wrapper functions will make it much easier to fixup apps that depended on the old include paths.

With this they shouldn't need to ever refer directly to an include path (or any path) from another target, so it should avoid any future issues if files get moved around again.

See also nasa/sample_lib#46 and nasa/sample_app#126


jphickey:
If possible I'd like to fast track this one since it's part of the path to merging #1203 and getting the directory reorg settled.  It's bigger than what I'd normally request to fast track but due to time line and the fact that it doesn't change FSW at all maybe we can?

skliper (approved):

PR #1229, opened 2021-03-16
Fix #1164, use FS file name parser for commands

### LINKED ISSUE #1164: Commands using files should utilize new filename parser
Issue #981 + PR #1156 introduces a new FS utility API to parse a filename which is aware of the expected path and extension, and will default to these if unspecified in the input.  The initial PR only applies it to ES startup script.

But this creates an inconsistency; now you can put just a basename e.g. `ci_lab` in the startup script, but passing the same string as the filename to `CFE_ES_START_APP_CC` command will __not__ work, which doesn't seem right.

For consistency in operation CFS should also employ the same filename parser in all commands that accept a filename.  

**Additional context**
It is (intentionally, by design) fairly trivial to update a current call to `CFE_SB_MessageStringGet()` to use the filename-aware alternative instead.  

The only minor issue/concern is that the new function has more input validation than `CFE_SB_MessageStringGet()` does, so it is more important to check the return status.  So doing this generally introduces another check + path to generate the error event, so a UT test case needs to be added.  But that's about it.

**Requester Info**

jphickey:
**Describe the contribution**
For commands containing file names, replace the call to CFE_SB_MessageStringGet() - which is just a basic copy - to
the new filename-aware function CFE_FS_ParseInputFileName().

This means that the default pathname/extension logic is applied here too and only a "basename" is strictly necessary, although
if a full/absolute path is given, it will be used as is.

Fixes #1164

**Testing performed**
Build and sanity check CFE
Issue commands which accept file names (e.g. RESTART app) and confirm they accept the same file names as the startup script accepts.

**Expected behavior changes**
Commands accept "bare" file names using same logic as startup script, assuming a default path and extension if omitted.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/modules/evs/fsw/src/cfe_evs_task.c
+++ b/modules/evs/fsw/src/cfe_evs_task.c
@@ -1693,20 +1693,33 @@ int32 CFE_EVS_WriteAppDataFileCmd(const CFE_EVS_WriteAppDataFileCmd_t *data)
     const CFE_EVS_AppDataCmd_Payload_t *CmdPtr = &data->Payload;
     char                                LocalName[OS_MAX_PATH_LEN];
 
-    /* Copy the commanded filename into local buffer to ensure size limitation and to allow for modification */
-    CFE_SB_MessageStringGet(LocalName, CmdPtr->AppDataFilename, CFE_PLATFORM_EVS_DEFAULT_APP_DATA_FILE,
-                            sizeof(LocalName), sizeof(CmdPtr->AppDataFilename));
+    /*
+    ** Copy the filename into local buffer with default name/path/extension if not specified
+    */
+    Result = CFE_FS_ParseInputFileNameEx(LocalName, CmdPtr->AppDataFilename, sizeof(LocalName),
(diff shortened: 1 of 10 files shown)


modules/evs/fsw/src/cfe_evs_task.c
```diff
@@ -1725,52 +1738,56 @@ int32 CFE_EVS_WriteAppDataFileCmd(const CFE_EVS_WriteAppDataFileCmd_t *data)
...
-        OS_close(FileHandle);
-    }
+            OS_close(FileHandle);
+        }
 
-    return (Result);
+        return (Result);
 
-} /* End CFE_EVS_WriteAppDataFileCmd */
+    } /* End CFE_EVS_WriteAppDataFileCmd */
```
skliper, line 1791: Doesn't look right...

jphickey: Yeah, something went amiss when I rebased across the format change.  I am trying again using another method....

astrogeco:
[CCB:2021-03-24](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-24-2021) APPROVED

PR #1250, opened 2021-03-23
Fix #1215, remove task registration calls

### LINKED ISSUE #1215: Clean up ES task/app registration
In nasa/osal#853 it is proposed that `OS_TaskRegister()` be finally deprecated/removed.

But this function is invoked by two places in CFE ES that follow a similar pattern: `CFE_ES_RegisterApp()` and `CFE_ES_RegisterChildTask()`.

At a minimum, calls to `OS_TaskRegister()` must be removed to allow the function to be deprecated.

Furthermore there is already a task startup wrapper in ES that can be used to call/handle setting up environment (`CFE_PSP_SetDefaultExceptionEnvironment()`) meaning that these two functions themselves can also be deprecated - basically using the same design pattern as OSAL uses such that we don't need to burden apps with calling this extra function.  This design is simpler and less error prone.

**Additional context**
See nasa/osal#853

**Requester Info**

jphickey:
**Describe the contribution**
Explicit task registration is no longer necessary, since all required actions can be done before invoking the entry point.

This moves the invocation of `CFE_PSP_AttachExceptions()` from the registration function to the pre-entry function, this was
the only remaining action in task registration.

All references to task registration in code, docs, and tests are removed.

Fixes #1215

**Testing performed**

**Expected behavior changes**
No impact to behavior, but this is an API change that affects apps.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
CFE_ES_RegisterApp and CFE_ES_RegisterChildTask functions no longer exist.  Requires updates to apps to remove these calls if they are referenced.


astrogeco:
[CCB:2021-03-24](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-24-2021) APPROVED

PR #1251, opened 2021-03-23
Fix #808, ES API functional tests

### LINKED ISSUE #808: Add cFE ES Information API functional tests
Need open source functional tests for certifiability

Add cFE Information API functional tests
– CFE_ES_GetResetType - Return the most recent Reset Type.
– CFE_ES_GetAppID - Get an Application ID for the calling Application.
– CFE_ES_GetAppIDByName - Get an Application ID associated with a specified Application name.
– CFE_ES_GetAppName - Get an Application name for a specified Application ID.
– CFE_ES_GetAppInfo - Get Application Information given a specified App ID.
– CFE_ES_GetTaskInfo - Get Task Information given a specified Task ID.

N/A

**Additional context**
N/A

**Requester Info**
Jacob Hageman - NASA/GSFC

zanzaben:
**Describe the contribution**
Fixes #808 
New tests for the ES Info API's 

**Testing performed**
This is the test

**Expected behavior changes**
none

**Additional Context**
Dependant on #1268

Addresses are only valid on VxWorks system so the tests for when that is true have not be run and verified. 

**System(s) tested on**
Ubuntu 20.04

Alex Campbell GSFC


modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,197 @@
...
+    UtAssert_StrCmp(TestAppInfo.FileName, "/cf/cfe_testrunner.so", "Test App Info -> FileName = %s",
+                    TestAppInfo.FileName);
+    UtAssert_True(strlen(ESAppInfo.FileName) == 0, "ES App Info -> FileName  = %s", ESAppInfo.FileName);
+
+    UtAssert_True(TestAppInfo.StackSize > 0, "Test App Info -> StackSz  = %d", TestAppInfo.StackSize);
+    UtAssert_True(ESAppInfo.StackSize > 0, "ES App Info -> StackSz  = %d", ESAppInfo.StackSize);
+
+    if (TestAppInfo.AddressesAreValid)
+    {
+        /* ToDo: check these are non 0 */
```
skliper, line 78: Resolve TODO

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+#include "cfe_test.h"
+
+void TestAppInfo(void)
+{
+    uint32           TestAppId;
+    uint32           ESAppId;
+    uint32           AppIdByName;
+    uint32           InvalidID   = 123;
```
skliper, line 41: Recommend using OS_OBJECT_ID_UNDEFINED instead of using an integer as an ID directly, don't really need a variable.

Same for the other tests

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,197 @@
...
+    UtAssert_StrCmp(TaskInfo.TaskName, AppInfo.MainTaskName, "TaskInfo.TaskName (%s) = AppInfo.MainTaskName (%s)",
+                    TaskInfo.TaskName, AppInfo.MainTaskName);
+
+    UtAssert_INT32_EQ(TaskInfo.TaskId, AppInfo.MainTaskId);
+    UtAssert_INT32_EQ(TaskInfo.AppId, AppId);
+    UtAssert_INT32_EQ(TaskInfo.ExecutionCounter, AppInfo.ExecutionCounter);
+
+    UtAssert_INT32_EQ(CFE_ES_GetTaskInfo(&TaskInfo, InvalidID), CFE_ES_ERR_RESOURCEID_NOT_VALID);
+    UtAssert_INT32_EQ(CFE_ES_GetTaskInfo(NULL, InvalidID), CFE_ES_BAD_ARGUMENT);
+    // UtAssert_INT32_EQ(CFE_ES_GetTaskID(NULL), CFE_ES_BAD_ARGUMENT);
```
skliper, line 144: Commented out code

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,197 @@
...
+    UtAssert_INT32_EQ(CFE_ES_GetLibInfo(&LibInfo, LibIdByName), CFE_SUCCESS);
+    UtAssert_True(LibInfo.Type == 3, "Lib Info -> Type = %d", LibInfo.Type);
+    UtAssert_StrCmp(LibInfo.Name, LibName, "Lib Info -> Name = %s", LibInfo.Name);
+    UtAssert_StrCmp(LibInfo.EntryPoint, "CFE_Assert_LibInit", "Lib Info -> EntryPt  = %s", LibInfo.EntryPoint);
+    UtAssert_StrCmp(LibInfo.FileName, "/cf/cfe_assert.so", "Lib Info -> FileName = %s", LibInfo.FileName);
+    UtAssert_True(LibInfo.StackSize == 0, "Lib Info -> StackSz  = %d", LibInfo.StackSize);
+
+    if (LibInfo.AddressesAreValid)
+    {
+        /* ToDo: check these are non 0 */
```
skliper, line 164: Resolve TODO

modules/cfe_testcase/src/es_info_test.c
zanzaben: The commented out bits are the dependency on #1268, it currently seg faults.

astrogeco:
[CCB:2021-03-31](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-31-2021) APPROVED
- Is there a `#define` to use in `es_info_test.c:50`?
- Assert on addresses, not sure if "zero" is invalid
- need to test on vxworks
- Note on file names: use the same name but append "coverage_test" as a prefix. See osal example
- Add a comment in cmake to specify that there is a convention
- Add an info line that specifies which API is being tested. There's a difference between *using* an API and *testing it*

modules/cfe_assert/src/cfe_assert_io.c
```diff
@@ -87,6 +87,9 @@ void UT_BSP_DoText(uint8 MessageType, const char *OutputMessage)
                 Prefix = "N/A";
                 break;
             case UTASSERT_CASETYPE_BEGIN:
+                CFE_ES_WriteToSysLog("\n");
+                CFE_ES_WriteToSysLog("\n");
+                CFE_ES_WriteToSysLog("\n"); /* add a bit of extra whitespace between tests */
```
jphickey, line 92: This should be left out of this PR ideally - we can do something better in #1266

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,261 @@
...
+ * Includes
+ */
+
+#include "cfe_test.h"
+
+void TestAppInfo(void)
+{
+    uint32           TestAppId;
+    uint32           ESAppId;
+    uint32           AppIdByName;
```
jphickey, line 40: These AppIDs must be `CFE_ES_AppID_t` type, not `uint32`

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,261 @@
...
+    const char *     InvalidName = "INVALID_NAME";
+    char             AppNameBuf[OS_MAX_API_NAME + 4];
+    CFE_ES_AppInfo_t TestAppInfo;
+    CFE_ES_AppInfo_t ESAppInfo;
+
+    UtPrintf("Testing: CFE_ES_GetAppID, CFE_ES_GetAppIDByName, CFE_ES_GetAppName, CFE_ES_GetAppInfo");
+
+    UtAssert_INT32_EQ(CFE_ES_GetAppIDByName(&AppIdByName, TestAppName), CFE_SUCCESS);
+    UtAssert_INT32_EQ(CFE_ES_GetAppID(&TestAppId), CFE_SUCCESS);
+    UtAssert_INT32_EQ(TestAppId, AppIdByName);
```
jphickey, line 52: Should not be tested as [u]int32, use CFE_RESOURCEID_TEST_EQUAL macro.

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,261 @@
...
+    UtAssert_INT32_EQ(ESAppInfo.ExceptionAction, 1);
+
+    UtAssert_True(TestAppInfo.Priority > 0, "Test App Info -> Priority  = %d", TestAppInfo.Priority);
+    UtAssert_True(ESAppInfo.Priority > 0, "ES App Info -> Priority  = %d", ESAppInfo.Priority);
+
+    UtAssert_True(TestAppInfo.NumOfChildTasks == 0, "Test App Info -> Child Tasks  = %d", TestAppInfo.NumOfChildTasks);
+    UtAssert_True(ESAppInfo.NumOfChildTasks > 0, "ES App Info -> Child Tasks  = %d", ESAppInfo.NumOfChildTasks);
+
+    UtAssert_INT32_EQ(CFE_ES_GetAppIDByName(&AppIdByName, InvalidName), CFE_ES_ERR_NAME_NOT_FOUND);
+    UtAssert_INT32_EQ(AppIdByName, CFE_ES_APPID_UNDEFINED);
```
jphickey, line 113: Should not test as [u]int32 - use `!CFE_RESOURCEID_TEST_DEFINED()` macro

jphickey (changes requested):
I pulled this to test it in conjunction with what I will propose for #1266, but found it does not build when using strict ID types.

CFE/CFS developers should ideally all set this option:

    set(MISSION_RESOURCEID_MODE STRICT)

in your local `global_build_options.cmake` file.  It isn't set that way by default in order to be compatible for external users that may have existing code, but all framework code should build with this strict option set.

modules/cfe_testcase/src/es_info_test.c
astrogeco: @Alex is this fixed yet?

zanzaben: doing it now, will be fixed soon.

zanzaben: @jphickey Do we really need to use that because it changes a simple UtAssert_INT32_EQ(TestAppId, AppIdByName); to UtAssert_True(CFE_RESOURCEID_TEST_EQUAL(TestAppId, AppIdByName), "TestAppId (%d) == AppIdByName (%d)", TestAppId, AppIdByName);

modules/cfe_testcase/src/es_info_test.c
zanzaben: @jphickey same as above, do we really need to change 
UtAssert_INT32_EQ(AppIdByName, CFE_ES_APPID_UNDEFINED);
to
UtAssert_True(!CFE_RESOURCEID_TEST_DEFINED(AppIdByName), "AppIdByName (%d) = CFE_ES_APPID_UNDEFINED (%d)", AppIdByName, CFE_ES_APPID_UNDEFINED);

modules/cfe_testcase/src/es_info_test.c
jphickey: Yes, absolutely.   Code should never assume a value is a simple integer.  Importantly - IDs should not be interchanged/intermixed with other numbers or other ID types.

For instance it would be very easy in your tests (in fact you can try this!) to mistakenly call `CFE_ES_GetTaskID()` instead of `CFE_ES_GetAppID()` and/or assign an appID value to a taskID variable or something of that nature.  We need to make sure mistakes like this are caught - we've had bugs in the past because something like this was done.  If all values are just `uint32` then they cannot be caught, anything goes.

When you use `set(MISSION_RESOURCEID_MODE STRICT)` in your global_build_options.cmake file as I mentioned this type of mistake becomes a compile error.  The only reason its not set that way by default is for backward compatibility, but all framework devs should have this enabled.

The correct test should be something like: 
`UtAssert_True(CFE_RESOURCEID_TEST_EQUAL(TestAppId, AppIdByName), "TestAppId (%lu) == AppIdByName (%lu)", CFE_RESOURCEID_TO_ULONG(TestAppId), CFE_RESOURCEID_TO_ULONG(AppIdByName));`

Yep, its more verbose. but it is type-safe.

modules/cfe_testcase/src/es_info_test.c
jphickey: Note that you'll need to use `CFE_RESOURCEID_TO_ULONG` macro and the `%lu` conversion specifier as I mentioned in my previous comment.  Please enable the `set(MISSION_RESOURCEID_MODE STRICT)` in your build as I mentioned, so the compiler will better tell you if you've made type assumptions or mixed together types like this.

We also shouldn't assume that `CFE_RESOURCEID_TEST_DEFINED` is the same as equality with the `CFE_ES_APPID_UNDEFINED` value.  There could be many values which are "not defined", of which `CFE_ES_APPID_UNDEFINED` is guaranteed to be one of them, but maybe not the other way around.   It's like a square/rectangle type of thing.

zanzaben: It was your idea to use CFE_RESOURCEID_TEST_DEFINED. Since I am comparing ID's I rather use CFE_RESOURCEID_TEST_EQUAL if we can't just UtAssert_INT32_EQ

jphickey: No, this test should not be checking for equality to a specific value.  The only thing that is ensured is that if `CFE_ES_GetAppIDByName()` fails and returns an error, that the output ID also will also __not__ pass the `CFE_RESOURCEID_TEST_DEFINED` check.

Yes, in the current implementation this outputs specifically the `CFE_ES_APPID_UNDEFINED` value, but that might not always be the case, and might not be the only possible value.  

So the only thing that should be checked/asserted here is that `CFE_RESOURCEID_TEST_DEFINED()` is false.  The actual value doesn't matter - we don't need to compare it to a specific value.  Just confirm that the test is false, and that's enough.

jphickey: To clarify I would recommend the assert as follows:

    UtAssert_True(!CFE_RESOURCEID_TEST_DEFINED(AppIdByName), "AppIdByName (%lu) not defined", CFE_RESOURCEID_TO_ULONG(AppIdByName));

Also note that I have no objections to making a local macro or inline function to do this check as this is likely to need to be repeated frequently, and it would make it more readable.

zanzaben: I disagree, since the code now sets it to CFE_ES_APPID_UNDEFINED that is what it should be testing for. If right now it was set to something different that was also invalid, I would view that as a bug since the code is not doing what you expect it to and so should be treated as a bug. If in the future someone changes the functionality of CFE_ES_GetAppIDByName then the test would be changed to test the new functionality.

astrogeco: Are we good with this?

modules/cfe_testcase/src/cfe_test.h
```diff
@@ -43,7 +43,16 @@
...
 #include "utassert.h"
 
-void  ES_Test_AppId(void);
+/* Compare two Resource IDs */
+#define UtAssert_ResourceID_EQ(actual, expect)                                                \
+    UtAssert_True(CFE_RESOURCEID_TEST_EQUAL(actual, expect), "%s (%lu) == %s (%lu)", #actual, \
+                  CFE_RESOURCEID_TO_ULONG(actual), #expect, CFE_RESOURCEID_TO_ULONG(expect))
+
+/* Check if a Resource ID is Undeifined */
+#define UtAssert_ResourceID_Undifeined(id) \
```
jphickey, line 52: Typo here

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,263 @@
...
+    if (TestAppInfo.AddressesAreValid)
+    {
+        UtAssert_True(TestAppInfo.AddressesAreValid > 0, "Test App Info -> AddrsValid?  = %d",
+                      (int)TestAppInfo.AddressesAreValid);
+        UtAssert_True(TestAppInfo.CodeAddress > 0, "Test App Info -> CodeAddress  = %d", (int)TestAppInfo.CodeAddress);
+        UtAssert_True(TestAppInfo.CodeSize > 0, "Test App Info -> CodeSize  = %d", (int)TestAppInfo.CodeSize);
+        UtAssert_True(TestAppInfo.DataAddress > 0, "Test App Info -> DataAddress  = %d", (int)TestAppInfo.DataAddress);
+        UtAssert_True(TestAppInfo.DataSize > 0, "Test App Info -> DataSize  = %d", (int)TestAppInfo.DataSize);
+        UtAssert_True(TestAppInfo.BSSAddress > 0, "Test App Info -> BSSAddress  = %d", (int)TestAppInfo.BSSAddress);
+        UtAssert_True(TestAppInfo.BSSSize > 0, "Test App Info -> BSSSize  = %d", (int)TestAppInfo.BSSSize);
```
jphickey, line 85: addresses should not be cast as `int`, they should be `unsigned long`

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -0,0 +1,263 @@
...
+    UtAssert_True(LibInfo.StackSize == 0, "Lib Info -> StackSz  = %d", (int)LibInfo.StackSize);
+
+    if (LibInfo.AddressesAreValid)
+    {
+        UtAssert_True(LibInfo.AddressesAreValid > 0, "Lib Info -> AddrsValid?  = %d", (int)LibInfo.AddressesAreValid);
+        UtAssert_True(LibInfo.CodeAddress > 0, "Lib Info -> CodeAddress  = %d", (int)LibInfo.CodeAddress);
+        UtAssert_True(LibInfo.CodeSize > 0, "Lib Info -> CodeSize  = %d", (int)LibInfo.CodeSize);
+        UtAssert_True(LibInfo.DataAddress > 0, "Lib Info -> DataAddress  = %d", (int)LibInfo.DataAddress);
+        UtAssert_True(LibInfo.DataSize > 0, "Lib Info -> DataSize  = %d", (int)LibInfo.DataSize);
+        UtAssert_True(LibInfo.BSSAddress > 0, "Lib Info -> BSSAddress  = %d", (int)LibInfo.BSSAddress);
```
jphickey, line 183: Same here - addresses should be unsigned long.

jphickey (changes requested):
I looked again, just a couple more minor updates  ... it's close!

modules/cfe_testcase/src/cfe_test.h
zanzaben: fixed

modules/cfe_testcase/src/es_info_test.c
zanzaben: fixed

modules/cfe_testcase/src/es_info_test.c
zanzaben: fixed

PR #1271, opened 2021-03-29
Fix #1270, limit check in pool validation

### LINKED ISSUE #1270: Incorrect limit check in CFE_ES_GenPoolValidateState
**Describe the bug**
A check for the "NumBuckets" member should be `<=` but it is using `<`.

**To Reproduce**
Attempt to validate PoolID 1 which uses the max number of buckets.  Function returns false, but record is valid.

**Expected behavior**
CFE_ES_GenPoolValidateState() should return true if using the max number of buckets (17 by default).

**Code snips**
One-liner here:
https://github.com/nasa/cFE/blob/98bc15851ebbca83b01947520232ca0cf67a30cf/modules/es/fsw/src/cfe_es_generic_pool.c#L604

**System observed on:**
Ubuntu 20.04

**Reporter Info**

jphickey:
**Describe the contribution**
It is OK/valid if NumBuckets is the maximum value.

Fixes #1270

**Testing performed**
Build and run all unit tests

**Expected behavior changes**
Pool structure using max value will correctly validate

**System(s) tested on**
Ubuntu 20.04


astrogeco:
[CCB:2021-03-31](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-March-31-2021) APPROVED
- Need a unit test

PR #1275, opened 2021-04-01
Fix #1157, increment sequence if indicated

### LINKED ISSUE #1157: Auto increment sequence on CMD packets
The `CFE_SB_TransmitMsg()` function has a boolean parameter to indicate if the sequence number should be incremented automatically based on the route.  This route-based sequence number overrides whatever was in the buffer.

However, this only works on telemetry packets.  There is a test that checks specifically if the packet is TLM, and only updates the sequence number if so.

This means that all locally-generated messages of the "CMD" variety - which includes the various TIME messages e.g. 1Hz, and all "SEND_HK" packets generated by SCH - always have a sequence number of 0 in their header.

The sequence number should probably be valid on all these - otherwise there is no way to tell if a message was missed.

**Additional context**
It is not clear why only locally-generated TLM packets would have a valid sequence, but not locally-generated CMD packets.

Note SCH_LAB currently calls `CFE_SB_TransmitMsg()` with the value set `false` -- so this would have to change too in order to get valid sequence numbers in SEND_HK packets.

**Requester Info**

jphickey:
**Describe the contribution**
Do not restrict sequence increment to TLM only, since CMD packets can be locally generated too (e.g. SCH, TIME).

Fixes #1157

**Testing performed**
Build and sanity check CFE, run all unit tests.

**Expected behavior changes**
All transmitted messages will get their sequence count incremented if transmitted with parameter `IncrementSequenceCount` set to true.  Previously, for some reason, this flag was only honored on TLM packets.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Thus far I've not found a real reason/explanation as to why this flag would be ignored for CMD packets.  If there is a good reason, it should be documented.  Otherwise for the sake of consistency it should be done for all messages.

### DIFF
--- a/modules/core_api/fsw/inc/cfe_sb.h
+++ b/modules/core_api/fsw/inc/cfe_sb.h
@@ -412,8 +412,7 @@ CFE_Status_t CFE_SB_UnsubscribeLocal(CFE_SB_MsgId_t MsgId, CFE_SB_PipeId_t PipeI
 ** \param[in]  MsgPtr       A pointer to the message to be sent.  This must point
 **                          to the first byte of the message header.
 ** \param[in] IncrementSequenceCount Boolean to increment the internally tracked
-**                                   sequence count and update the message if the
-**                                   buffer contains a telemetry message
+**                                   sequence count and update the message
 **
 ** \return Execution status, see \ref CFEReturnCodes
 ** \retval #CFE_SUCCESS         \copybrief CFE_SUCCESS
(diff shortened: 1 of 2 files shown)


jphickey:
Tagging @acudmore  and @jwilmot  in case they can offer some history on this subject.

As it stands we handle CMD vs. TLM packets inconsistently - basically we ignore the "IncrementSequenceCount" flag if it is a command - without really any explanation/justification as to why we would ignore it.  I'm suggesting removing this to make it consistent, but there may be some history I'm not aware of.

My recommendation would be - if there is a use case where the counter shouldn't be incremented, why can't the caller just set this parameter `false` ...?

astrogeco:
[CCB:2021-04-07](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-April-07-2021) 
- This has a history and we need to be careful changing this

Issue #921, opened 2020-09-30
update cFE source and unit tests to use CFE_Status_t


CDKnightNASA:
**Is your feature request related to a problem? Please describe.**
As a followon to #888, we should update all of the cFE source to use the CFE_Status_t return status type instead of int32.

**Describe alternatives you've considered**
May want to do this in phases, per component; or incorporate into a clang-format code cleanup change set.

**Requester Info**
Christopher.D.Knight@nasa.gov

jphickey:
I ended up basically implementing this as part of doing other reviews, so I will submit the PR, but it should go into the Draco set, not for Caelum.

dzbaker:
Closing as addressed in #2110.

Issue #508, opened 2020-02-06
Doxygen event documentation needs scrub


skliper:
**Is your feature request related to a problem? Please describe.**
Doxygen event output doesn't match code in at least 1 case, see CFE_ES_INITSTATS_INF_EID

**Describe the solution you'd like**
Scrub event doxygen and confirm it matches actual output (and type)

**Describe alternatives you've considered**
Maybe #define the string as a single definition and link in documentation

**Additional context**
N/A

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
Consider adding to the event macro to include the brief, possibly link to code (but not good if it links to a line number...)... adding a reference in code might work, but then you'd need to include all the code in the user's guide instead of just APIs (vs what's in the general doc).  Also related to suggestion in #655.

skliper:
Consider not linking the format string into the user's guide, and do it in the developers doc?  Could keep the user's guide generic with just the description, and add the necessary doxygen tags in the code to show the format string.

skliper:
Current plan - replace format string with short description (more useful to user).  Can always automate generation of EID/format string mapping from the code if really needed.  Confirm type and description is still relevant and remove unused.

Issue #275, opened 2019-09-30
ES comments need checked for consistency


skliper:
Lots of copy/paste type errors in ES. For example, after CFE_ES_IncrementTaskCounter() the comment says  /* End of CFE_ES_ExitChildTask() */

skliper:
_Imported from trac issue 244.  Created by sseeger on 2018-08-14T19:37:54, last modified: 2019-07-03T12:48:08_

skliper:
_Trac comment by jhageman on 2019-07-03 12:48:08:_

Moved unfinished 6.6.1 issues to next minor release

skliper:
Just piling on, `/* end of file */` in the middle of the file and a handful of others.

skliper:
Comments and format cleanup planned once this round of PRs are closed

thnkslprpt:
Most or all of the typos/mistakes that were originally noticed when this issue was opened have probably been cleared up by now.

I would suggest that this can be closed as completed by recent commits.

dzbaker:
Closing as these issues have been addressed.

Issue #10, opened 2018-01-10
submodule updates fail


nickcounts:
While attempting to build this project, the submodule update process fails on `GFS_CI.git`

```
Cloning into 'apps/ci'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:nasa/CFS_CI.git' into submodule path 'apps/ci' failed
```

It looks like the recent [ commit in .gitmodules  ](https://github.com/nasa/cFE/commit/edc14ceb428989b1f183d19027b9d28a29cf96cc)  added new apps and used ssh instead of HTTPS. I presume changing them to the form of

```
[submodule "apps/to"]
	path = apps/to
	url = https://github.com/nasa/CFS_TO.git
[submodule "apps/ci"]
	path = apps/ci
	url = https://github.com/nasa/CFS_CI.git
[submodule "apps/io_lib"]
	path = apps/io_lib
	url = https://github.com/nasa/CFS_IO_LIB.git
``` 

will prevent this issue for others.

the-other-james:
Thanks for the catch!

PR #1283, opened 2021-04-07
cFS-Caelum Review: CFS-40

astrogeco:
## Description

This is a "bookkeeping" Pull Request meant for the cFS-Caelum Review code inspection (full-scale code review). 

This PR is solely focused on "CFS-40". For more info see [this readme](https://github.com/astrogeco/cFE/blob/caelum-code-review-cfs40/README.md)

The Included files are

https://github.com/astrogeco/cFE/blob/c79d66969673a37900d468203b161785bd19874f/.gitignore#L11-L49 

```.gitignore
# Build Files
!cmake/**/target_config.*
!cmake/target/CMAKELists.txt
!cmake/arch_build.cmake
!cmake/cfe_generated_file.h.in
!cmake/check_header.c.in
!cmake/generate_build_env.cmake
!cmake/generate_git_module_version.cmake
!cmake/global_functions.cmake
!cmake/mission_build.cmake
!cmake/mission_defaults.cmake
!/CMakeLists.txt

# cFE-Core
!modules/core_api/fsw/inc/cfe.h
!modules/core_api/fsw/inc/cfe_endian.h
!modules/core_api/fsw/inc/cfe_error.h
!modules/core_api/fsw/inc/cfe_version.h

# ES
!modules/core_api/fsw/inc/cfe_es*
!modules/core_private/fsw/inc/cfe_es*

!modules/es/fsw/**
!modules/es/CMakeLists.txt

# ResourceID
!modules/core_api/fsw/inc/cfe_resourceid*
!modules/core_private/fsw/inc/cfe_core_resourceid_basevalues.h

!modules/resourceid/fsw/src/*
!modules/resourceid/option_inc/*
!modules/resourceid/CMakeLists.txt
!modules/resourceid/mission_build.cmake
```

## Objectives

1. This review starts on 2021-04-05 and ends on 2021-04-09
2. Dispositions of findings is on 2021-04-12

3. Reviewers only need to review source files, header files & build files.
4. Use .ppt, .pdf, .txt & .xlsx files for background information about the code.
5. See the Attachments section for Peer Review Data Package.

6. See also "The Power of 10" rules for safety-critical code. https://en.wikipedia.org/wiki/The_Power_of_10:_Rules_for_Developing_Safety-Critical_Code#:~:text=The%20Power%20of%2010%20Rules,to%20review%20or%20statically%20analyze

7. NOTE: Don't spend too much time over coding standard violations. The Static Code Analysis tool will enforce the coding standards. This code is developed by GSFC, so GSFC coding standards will be enforced for this code base.

### Notes

Note a few already existing issues (no need to individually comment on occurrences):

Doxygen event documentation doesn't match code: https://github.com/nasa/cFE/issues/508
End of function comments out of date (generalized/paraphrased version of https://github.com/nasa/cFE/issues/275)
Update code/unit tests to use CFE_Status_t: https://github.com/nasa/cFE/issues/921

If there's anything else that is observed as a repeated pattern, feel free to document as a general commen

There are several places that would trigger warnings with some common compilers/warning options. It would be nice to follow #10 rule in "The Power of 10".

Quick summary/references for currently enforced settings on the FSW
Compiler options (note -Wall and -Werror) - https://github.com/nasa/cFE/blob/0850b19a09417dd7730e9c1b44e0f5687ef0f721/cmake/sample_defs/arch_build_custom.cmake#L28-L37

cppcheck - https://github.com/nasa/cFE/blob/0850b19a09417dd7730e9c1b44e0f5687ef0f721/.github/workflows/run_fsw_cppcheck.sh#L3

CodeQL - https://github.com/nasa/cFE/blob/0850b19a09417dd7730e9c1b44e0f5687ef0f721/.github/workflows/codeql-build.yml#L35-L39

CodeSonar - currently using default set for cFE, extending to JPL and MISRA is future work

For CodeQL and CodeSonar we don't eliminate all warnings, but we do analyze and disposition them all (plan to report dispositions as part of certification package)

This approach is compliant with the latest GSFC 582 standard (that is still going through review). Happy to discuss any additional settings that you have concerns about.


.gitignore
```diff
@@ -0,0 +1,49 @@
+# Ignore everything
+*
+
```
astrogeco, line 3: My comment

.gitignore
```diff
@@ -0,0 +1,49 @@
...
+# But not these files...
+!/.gitignore
+
+
+# Build Files
+!cmake/**/target_config.*
+!cmake/target/CMAKELists.txt
+!cmake/arch_build.cmake
+!cmake/cfe_generated_file.h.in
```
astrogeco, line 15: Another comment

.gitignore
```diff
@@ -0,0 +1,49 @@
+# Ignore everything
+*
+
+# ...even if they are in subdirectories
+!*/
+
+# But not these files...
+!/.gitignore
+
```
astrogeco, line 9: Re-start Review

cmake/global_functions.cmake
```diff
@@ -0,0 +1,266 @@
...
+# we can create a build-specific include file that just #include's the real file
+#
+# This also supports "stacking" multiple component files together by specifying more than one
+# source file for the wrapper.
+#
+# This function now accepts named parameters:
+#   FILE_NAME - the name of the file to write
+#   FALLBACK_FILE - if no files are found in "defs" using the name match, this file will be used instead.
+#   MATCH_SUFFIX - the suffix to match in the "defs" directory (optional)
+#   PREFIXES - a list of prefixes to match in the "defs" directory (optional)
```
ezpollack, line 71: Should documentation  specify optional argument OUTPUT_DIRECTORY and its default location?

cmake/global_functions.cmake
```diff
@@ -0,0 +1,266 @@
...
+#       is forced to be last.
+#   MISSION_APPS: full list of applications specified in the whole mission
+#   SYSID_<arch>: set for each entry of TGTSYS_LIST, and indicates the
+#       toolchain specified in the target file for that CPU arch.
+#   TGTSYS_<arch>: set to a list of CPU numbers that utilize the same arch 
+#   TGTSYS_<arch>_APPS: set for each entry of TGTSYS_LIST, and indicates the
+#       full set of applications that need to built for that target architecture
+#   TGTSYS_<arch>_DRIVERS: set for each entry of TGTSYS_LIST, and indicates the
+#       full set of device drivers that need to built for that target architecture
+#
```
ezpollack, line 145: Documentation states that read_targetconfig creates variables TGTSYS_<arch>_DRIVERS and SYSID_<arch> which are not created in this function. Also might need to specify variables TGTSYS_<arch>_PSPMODULES, TGTSYS_<arch>_STATICAPPS

cmake/mission_build.cmake
```diff
@@ -0,0 +1,487 @@
...
+    "${CMAKE_BINARY_DIR}/doc/osconfig-example.h")
+
+  # The user guide should include the doxygen from the _public_ API files from CFE + OSAL
+  # NOTE: the userguide is built against the headers of the default core apps. Even if
+  # an alternate version of the module is in use, it should adhere to the same interface.
+  set(SUBMODULE_HEADER_PATHS
+    "${osal_MISSION_DIR}/src/os/inc/*.h"
+    "${psp_MISSION_DIR}/psp/fsw/inc/*.h"
+  )
+  foreach(MODULE core_api es evs fs msg sb tbl time)
```
ezpollack, line 291: Missing SBR and resourceID modules?

cmake/mission_build.cmake
```diff
@@ -0,0 +1,487 @@
...
+  # Generate the tools for the native (host) arch
+  # Add all public include dirs for core components to include path for tools
+  include_directories(
+    ${core_api_MISSION_DIR}/fsw/inc
+    #${es_MISSION_DIR}/fsw/inc
+    #${evs_MISSION_DIR}/fsw/inc
+    #${fs_MISSION_DIR}/fsw/inc
+    #${sb_MISSION_DIR}/fsw/inc
+    #${tbl_MISSION_DIR}/fsw/inc
+    #${time_MISSION_DIR}/fsw/inc
```
ezpollack, line 379: Should these commented out lines still be here?

modules/core_api/fsw/inc/cfe_version.h
```diff
@@ -0,0 +1,61 @@
...
+ * See @ref cfsversions for format differences between development and release versions.
+ */
+#define CFE_SRC_VERSION CFE_BUILD_BASELINE "+dev" CFE_STR(CFE_BUILD_NUMBER)
+
+/*! @brief Development Build Version String.
+ * @details Reports the current development build's baseline, number, and name. Also includes a note about the latest
+ * official version. @n See @ref cfsversions for format differences between development and release versions.
+ */
+#define CFE_VERSION_STRING                                                                                 \
+    " cFE DEVELOPMENT BUILD " CFE_SRC_VERSION " (Codename: Bootes)" /* Codename for current development */ \
```
ezpollack, line 58: Should this be updated to Caelum?

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
```diff
@@ -0,0 +1,193 @@
...
+typedef void (*CFE_ES_TaskEntryFuncPtr_t)(void); /**< \brief Required Prototype of Task Main Functions */
+typedef int32 (*CFE_ES_LibraryEntryFuncPtr_t)(
+    CFE_ES_LibId_t LibId); /**< \brief Required Prototype of Library Initialization Functions */
+
+/**
+ * \brief Compatible typedef for ES child task entry point.
+ *
+ * All ES task functions (main + child) use the same entry point type.
+ */
+typedef CFE_ES_TaskEntryFuncPtr_t CFE_ES_ChildTaskMainFuncPtr_t;
```
ezpollack, line 79: If all ES task functions (main + child according to the comments) use the same entry point type, couldn't we just define CFE_ES_TaskEntryFuncPtr_t here as "void (*CFE_ES_TaskEntryFuncPtr_t)(void)" and just replace instances of CFE_ES_ChildTaskMainFuncPtr_t with this type?

modules/core_private/fsw/inc/cfe_core_resourceid_basevalues.h
```diff
@@ -0,0 +1,95 @@
...
+
+    /* Other ES managed resources */
+    CFE_RESOURCEID_ES_APPID_BASE_OFFSET      = OS_OBJECT_TYPE_USER + 1,
+    CFE_RESOURCEID_ES_LIBID_BASE_OFFSET      = OS_OBJECT_TYPE_USER + 2,
+    CFE_RESOURCEID_ES_COUNTID_BASE_OFFSET    = OS_OBJECT_TYPE_USER + 3,
+    CFE_RESOURCEID_ES_POOLID_BASE_OFFSET     = OS_OBJECT_TYPE_USER + 4,
+    CFE_RESOURCEID_ES_CDSBLOCKID_BASE_OFFSET = OS_OBJECT_TYPE_USER + 5,
+
+    /* SB managed resources */
+    CFE_RESOURCEID_SB_PIPEID_RESOURCE_BASE_OFFSET = OS_OBJECT_TYPE_USER + 6
```
ezpollack, line 71: There seems to be a design decision to abstract cFE resource ID base values and limits by defining cfe_resourceid_basevalue.h using a local mission_build.cmake that would construct cfe_resourceid_basevalue.h using some overridable file. The fallback file for cfe_resourceid_basevalue.h is cfe_resourceid_osal_compatible.h. This file defines CFE_RESOURCE_ID_SHIFT, CFE_RESOURCEID_MAX, and CFE_RESOURCEID_MAKE_BASE to values that involve preprocessor definitions defined in OSAL. The file cfe_core_resourceid_basevalues.h has direct references to OS_OBJECT_TYPE_USER and OS_OBJECT_TYPE_OS_TASK when constructing ofsets for various ES and SB managed resource IDs. Should these offsets also be abstracted in a similar manner? Could we not just directly rely on user-defined CFE_RESOURCEID_MARK and CFE_RESOURCEID_SHIFT to separate the values by calling CFE_RESOURCEID_MAKE_BASE(n) where n is >=0 incrementing?

modules/core_api/fsw/inc/cfe_resourceid.h
```diff
@@ -0,0 +1,219 @@
...
+ *      CFE_ResourceId_IsDefined(CFE_RESOURCEID_UNDEFINED)
+ *
+ * Always returns false.
+ *
+ * @param[in]   id    Resource ID to check
+ * @returns True if the ID may refer to a defined entity, false if invalid/undefined.
+ */
+static inline bool CFE_ResourceId_IsDefined(CFE_ResourceId_t id)
+{
+    return (CFE_RESOURCEID_UNWRAP(id) != 0);
```
ezpollack, line 155: The value to denote whether a resource ID is undefined should be used here instead of just a hard 0. This should be cfe_resourceid_api_typedefs.h:CFE_RESOURCEID_UNDEFINED.

modules/core_api/fsw/inc/cfe.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "cfe_error.h" /* Define common cFE error codes */
+
+#include "cfe_es.h"   /* Define Executive Service API */
+#include "cfe_evs.h"  /* Define Event Service API */
+#include "cfe_fs.h"   /* Define File Service API */
+#include "cfe_sb.h"   /* Define Software Bus API */
+#include "cfe_time.h" /* Define Time Service API */
+#include "cfe_tbl.h"  /* Define Table Service API */
+
```
ezpollack, line 51: Missing include for cfe_sbr.h? This is included under MISSION_CORE_MODULES.

modules/resourceid/fsw/src/cfe_resourceid_api.c
```diff
@@ -0,0 +1,133 @@
...
+/*********************************************************************/
+/*
+ * CFE_ResourceId_GetSerial
+ *
+ * For complete API information, see prototype in header
+ */
+uint32 CFE_ResourceId_GetSerial(CFE_ResourceId_t ResourceId)
+{
+    return (CFE_ResourceId_ToInteger(ResourceId) & ((uint32)CFE_RESOURCEID_MAX));
+}
```
ezpollack, line 64: The CFE_RESOURCEID_MAKE_BASE macro is defined in each option_inc resource_id.h file. These files are intended to be swappable pending user configuration, so the way bases are created might be different pending user definitions. Should base and serial conversions be handled there as well?

modules/resourceid/fsw/src/cfe_resourceid_api.c
```diff
@@ -0,0 +1,133 @@
...
+int32 CFE_ResourceId_ToIndex(CFE_ResourceId_t Id, uint32 BaseValue, uint32 TableSize, uint32 *Idx)
+{
+    uint32 Serial;
+
+    if (Idx == NULL)
+    {
+        return CFE_ES_BAD_ARGUMENT;
+    }
+
+    Serial = CFE_ResourceId_ToInteger(Id) - BaseValue;
```
ezpollack, line 81: Should probably use  CFE_ResourceId_GetSerial(Id) to get the Serial here.

modules/resourceid/fsw/src/cfe_resourceid_api.c
```diff
@@ -0,0 +1,133 @@
...
+        {
+            CheckId = CFE_RESOURCEID_UNDEFINED;
+            break;
+        }
+
+        --Count;
+        ++Serial;
+        if (Serial >= CFE_RESOURCEID_MAX)
+        {
+            Serial %= TableSize;
```
ezpollack, line 125: I'm unsure that we need TableSize in this function, though I'm not clear on its use. I'm assuming TableSize should be the size of the resource ID table for each resource ID type, but I thought each table section should contain up to CFE_RESOURCE_ID_MAX IDs since their bases are all allocated on those boundaries by CFE_RESOURCEID_MAKE_BASE. Line 123 checks to see if we've gone over the table boundary by checking if the serial for the current ID is larger than CFE_RESOURCEID_MAX.  It then causes Serial to wrap around using the modulo operator with the supplied TableSize value. If TableSize is CFE_RESOURCEID_MAX, then Serial will be 0 which should be the start of the table. If we specified TableSize as a value smaller than CFE_RESOURCE_ID_MAX, then modulo will reset the serial number to some other place in the table entirely between 0 and TableSize after it rolls over. For instance, if we specified TableSize as 200 and we're trying to find the next ID and it rolls over the end of the table at 65535, then it will start checking again starting at serial number 35. 

We could maybe instead get away with just setting Serial to 0 on line 125 and setting Count to CFE_RESOURCEID_MAX.

modules/core_api/fsw/inc/cfe.h
skliper: SBR is only used internally by SB (no exposed API)

modules/core_api/fsw/inc/cfe_version.h
skliper: https://github.com/nasa/cFE/issues/1296 - happens upon official release

cmake/mission_build.cmake
skliper: SBR and resourceID are internal to cFE, no exposed API.

jphickey: Also note that this is changed in PR #1290, to use `${MISSION_CORE_MODULES}` instead.  See https://github.com/nasa/cFE/blob/1a72fad33f9329ca66b67688d19f2c351d56405a/cmake/mission_build.cmake#L291

cmake/mission_build.cmake
jphickey: Also fixed in PR #1290

modules/core_private/fsw/inc/cfe_core_resourceid_basevalues.h
jphickey: OTOH I was pondering if we should remove the `cfe_resourceid_osal_compatible.h` option and just force use of this config for now.  It is a proof of concept / future work to allow this (if desired) but right now due to other reasons we need to ensure CFE and OSAL IDs do not overlap, so its really not possible to use anything else.

modules/resourceid/fsw/src/cfe_resourceid_api.c
jphickey: Valid point, but the fundamental structure of the resource ID isn't really intended to be changeable by the user, it is a bitwise combination of a base and serial number, and other operations assume/require this general structure.  

What can change are the numeric values for the bases and the number of bits allocated to each.  (but currently the latter really should match OSAL at 16+16, for other reasons).

modules/resourceid/fsw/src/cfe_resourceid_api.c
jphickey: The computation is intentionally different here - where ` CFE_ResourceId_GetSerial()` works with any type by simply masking out the bits, ignoring base - the intent here is to take the difference from the _expected_ base.

Even if we changed this to use `CFE_ResourceId_GetSerial()` the code would still need to check for a match to the expected base value.  Doing it this way just gets both checks done in one operation.

modules/resourceid/fsw/src/cfe_resourceid_api.c
jphickey: Similar questions came up during OSAL review.  

Resource ID serial numbers and table positions are not the same thing.  Serial numbers change/increase each time a new object is allocated, up to `CFE_RESOURCEID_MAX`, whereas the `TableSize` is the actual number of objects that can coexist at the same time.  

The intent is that when an object is deleted and then a new object is created, it may re-use the same memory slot/table position that the old object had, but the new object does __not__ get the same ID as the old object had.

Since the number space is not infinite serial numbers will eventually be recycled but it takes thousands of allocations and deletions before that happens.

jphickey: Note that this is structured such that table position is always mathematically calculated/fixed based on serial number.   
Given a serial number (i.e. extracted from a passed in ID), taking modulus with the size of the table results in the expected/correct table position for that object.  

The reverse option cannot be mathematically computed, one has to go to the entry in the object table and get the resource ID out of it.

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
jphickey: Yes, but this was an existing type in the CFE API so it was kept the same name for backward compatibility.

modules/es/fsw/inc/cfe_es_events.h
```diff
@@ -0,0 +1,1451 @@
...
+
+/* **************************
+** ****** Maximum EID. ******
+** **************************
+** The EID's below may not necessarily be in order, so it can be difficult to
+** determine what the next EID is to use. When you add EID's, start with MAX_EID + 1
+** and when you're done adding, set this to the highest EID you used. It may
+** be worthwhile to, on occasion, re-number the EID's to put them back in order.
+*/
+#define CFE_ES_MAX_EID 92
```
stashakk, line 46: Why Not use an enum here instead of a # define for all of these, then you do not have to worry about assignment of numbers just add it to end, before MAX

modules/core_api/fsw/inc/cfe_es_extern_typedefs.h
```diff
@@ -0,0 +1,573 @@
...
+/* Use the local definitions of these types */
+
+#include "common_types.h"
+#include "cfe_resourceid_typedef.h"
+#include "cfe_mission_cfg.h"
+
+/**
+ * @brief Label definitions associated with CFE_ES_LogMode_Enum_t
+ */
+enum CFE_ES_LogMode
```
stashakk, line 47: For this and the other enums, why specify all of the values?  utilize the compiler/language which auto increments... use prescribed values if you are purposely skipping or starting someplace.

modules/core_api/fsw/inc/cfe_es_extern_typedefs.h
```diff
@@ -0,0 +1,573 @@
...
+    char EntryPoint[CFE_MISSION_MAX_API_LEN]; /**< \cfetlmmnemonic \ES_APPENTRYPT
+                                                   \brief The Entry Point label for the Application */
+    char FileName[CFE_MISSION_MAX_PATH_LEN];  /**< \cfetlmmnemonic \ES_APPFILENAME
+                                                   \brief The Filename of the file containing the Application */
+
+    CFE_ES_MemOffset_t StackSize;                  /**< \cfetlmmnemonic \ES_STACKSIZE
+                                                        \brief The Stack Size of the Application */
+    uint32 AddressesAreValid;                      /**< \cfetlmmnemonic \ES_ADDRVALID
+                                                        \brief Indicates that the Code, Data, and BSS addresses/sizes are valid */
+    CFE_ES_MemAddress_t CodeAddress;               /**< \cfetlmmnemonic \ES_CODEADDR
```
stashakk, line 462: Superficial comment, but there is a repetition of CFE_ES_MemAddress_t with CFE_ES_MemOffset_t, could make this into a new structure to associate these... So instead of code reading 

CFE_ES_MemAddress_t CodeAddress;
CFE_ES_MemOffset_t CodeSize;

could be:
CFE_ES_MemAddOff_t Code;

referenced by Code.Address and Code.Size
THus:

CFE_ES_MemAddOff_t Data;

CFE_ES_MemAddOff_t BSS;

possibly others

modules/core_api/fsw/inc/cfe_resourceid.h
skliper: Can use the macro

modules/resourceid/fsw/src/cfe_resourceid_api.c
skliper: Would be clarified if the configuration relative to OSAL is cleaned up

modules/core_api/fsw/inc/cfe_es_extern_typedefs.h
skliper: Even though they are enums, there are ground dependencies on these specific values.

modules/es/fsw/inc/cfe_es_events.h
skliper: There's a ground dependency on the actual values so prefer to avoid automatic adjustment.  Eventually these will probably go into a data dictionary (something shared between ground and target).

skliper:
This round of code review has been completed, for any further issues/comments please open a new issue.

modules/core_api/fsw/inc/cfe_es_extern_typedefs.h
skliper: See https://github.com/nasa/cFE/issues/1300

modules/core_private/fsw/inc/cfe_core_resourceid_basevalues.h
skliper: See https://github.com/nasa/cFE/issues/1301

modules/resourceid/fsw/src/cfe_resourceid_api.c
skliper: Related to https://github.com/nasa/cFE/issues/1301

modules/core_api/fsw/inc/cfe_resourceid.h
skliper: Dee https://github.com/nasa/cFE/issues/1302

modules/es/fsw/inc/cfe_es_msg.h
```diff
@@ -0,0 +1,1570 @@
...
+typedef struct CFE_ES_HousekeepingTlm_Payload
+{
+    uint8 CommandCounter;      /**< \cfetlmmnemonic \ES_CMDPC
+                                \brief The ES Application Command Counter */
+    uint8 CommandErrorCounter; /**< \cfetlmmnemonic \ES_CMDEC
+                           \brief The ES Application Command Error Counter */
+
+    uint16 CFECoreChecksum;    /**< \cfetlmmnemonic \ES_CKSUM
+                                    \brief Checksum of cFE Core Code */
+    uint8 CFEMajorVersion;     /**< \cfetlmmnemonic \ES_CFEMAJORVER
```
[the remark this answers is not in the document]
skliper: See https://github.com/nasa/cFE/issues/1303

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
skliper: See https://github.com/nasa/cFE/issues/1304

cmake/global_functions.cmake
skliper: See https://github.com/nasa/cFE/issues/1306

cmake/global_functions.cmake
skliper: See https://github.com/nasa/cFE/issues/1306

pavll:
Sorry for accidentally mentioning this pull request directly.

PR #1276, opened 2021-04-01
Fix #1266, use events for CFE test asserts

jphickey:
**Describe the contribution**
Reports test failures as CFE events, which allows for more processing capability, and allows failures to be recieved externally (e.g. ground system).

Fixes #1266

**Testing performed**
Update startup script to load the CFE functional test libs + apps and execute CFE to run the tests.  

**Expected behavior changes**
Test status messages sent as Events rather than Syslog.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Also tested in conjunction with #1251.

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #1292, opened 2021-04-09
Fix #1281, remove testrunner and convert testcase to app

### LINKED ISSUE #1281: Make functional tests runnable via CFE_ES_START_APP_CC command
In the functional test proof-of-concept, the test cases are libraries that are loaded in the startup script, and the "testrunner" app actually runs them.  But because libraries are not loadable/unloadable via command, it requires a custom startup script, and restarting CFE in order to run tests.

The functional tests should be integrated with test runner such that they are each their own app, therefore runnable by simply issuing the ES command to start the app if they were not part of the initial startup script.

**Additional context**
This will simplify the process of running tests on targets, since it will alleviate the need to customize the startup script.

The `cfe_assert` library itself will likely need to remain as a library, because it is a common dependency between all test apps.  But it shouldn't hurt to simply load this by default in the sample startup script, it won't interfere with FSW.

**Requester Info**

jphickey:
**Describe the contribution**
- Removes the separate CFE "testrunner" module and moves the logic associated with running a test into cfe_assert library.
- Converts the "testcase" module from a library into an app, by calling into the runner logic that is now inside cfe_assert.

This permits tests to be invoked at any time by issuing the ES "start app" command and specifying the testcase app.

Fixes #1281

**Testing performed**
Build and sanity check CFE.
Run all currently-implemented functional tests

**Expected behavior changes**
Each functional test is a separate app, not a library, so it can be started and stopped via ES command like any other app.

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Requires #1276 as prerequisite


jphickey:
Rebased to latest mainline.

zanzaben (changes requested):
Need to also update cFS/cfe/docs/README_functionaltest.md file with the new instructions on how to run the tests.

jphickey:
> Need to also update cFS/cfe/docs/README_functionaltest.md file with the new instructions on how to run the tests.

Updated in ec2d2f679

skliper:
CCB:2021-04-14 - APPROVED

zanzaben (approved):

zanzaben (changes requested):
2 of the test are failing and need to be updated. 

Test App Info -> EntryPt (line 66)
Test App Info -> FileName (line 69)

jphickey:
Update to correct name checks.  See commit e4afd35

jphickey:
Corrected string checks in commit 23c835b

zanzaben (approved):

PR #1373, opened 2021-04-16
Fix #1338, check status of call to CFE_ES_CDS_CachePreload

### LINKED ISSUE #1338: CFE_ES_ClearCDS: check status before while loop to zero
Status not checked from CFE_ES_CDS_CachePreload:
https://github.com/nasa/cFE/blob/e80aae94e0f56b868657daba965c590766a4dc57/modules/es/fsw/src/cfe_es_cds.c#L534

Analyze, check status if needed otherwise document why not for future reference

None

**Additional context**
Code review

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
Confirm that the call returned CFE_SUCCESS before continuing.

Fixes #1338 

**Testing performed**

**Expected behavior changes**
None

**System(s) tested on**
Ubuntu 20.04

**Additional context**
Note that the error/failure condition here is effectively dead code from the beginning.  Not even coverage test can exercise it, because its a void function, and the args are largely fixed/constant values.  So this now shows up as untested lines in the coverage report.

### DIFF
--- a/modules/es/fsw/src/cfe_es_cds.c
+++ b/modules/es/fsw/src/cfe_es_cds.c
@@ -529,9 +529,11 @@ int32 CFE_ES_ClearCDS(void)
     size_t                 RemainSize;
     int32                  Status;
 
+    Status = CFE_SUCCESS;
+
     /* Clear the CDS to ensure everything is gone */
     /* Create a block of zeros to write to the CDS */
-    Status = CFE_ES_CDS_CachePreload(&CDS->Cache, NULL, 0, sizeof(CDS->Cache.Data.Zero));
+    CFE_ES_CDS_CachePreload(&CDS->Cache, NULL, 0, sizeof(CDS->Cache.Data.Zero));
 
     /* While there is space to write another block of zeros, then do so */
     while (CDS->Cache.Offset < CDS->TotalSize)


skliper (commented):
If it can't fail, just comment as such and ignore the return.  It's ignored in the one other calling location: https://github.com/nasa/cFE/blob/e80aae94e0f56b868657daba965c590766a4dc57/modules/es/fsw/src/cfe_es_cds_mempool.c#L89

skliper:
> ... its a void function

I don't follow:
https://github.com/nasa/cFE/blob/e80aae94e0f56b868657daba965c590766a4dc57/modules/es/fsw/src/cfe_es_cds.h#L252

jphickey:
Updated to just ignore the return of CFE_ES_CDS_CachePreload here.  We still need to init the local status to CFE_SUCCESS, though (it had been doing that too).

Anyway, by "void function" I was referring to the parent, i.e. CFE_ES_ClearCDS() - there are no args to this function that coverage test can change, to get it to call CFE_ES_CDS_CachePreload any differently.  It always calls it with the same args, and they are always good/within range so there is no way to get the failure path from this function.

skliper:
Would it make sense to use sizeof the appropriate Cache->Data union member (instead of the underlying type)?  Then it really could never get out of sync?

astrogeco:
[CCB:2021-04-21](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-April-21-2021) APPROVED

PR #1437, opened 2021-04-28
Fix #1436, table object file name

### LINKED ISSUE #1436: Incorrect assumption about file name patterns in table build
**Describe the bug**
On most CMake platforms the generated object file names are simply the file name (including the .c) concatenated with CMAKE_C_OUTPUT_EXTENSION.  So, for example, on table build for sample_app, the file `sample_app_tbl.c` gets compiled into `sample_app_tbl.c.o` which can then be used to call `elf2cfetbl`.

Apparently this relationship is not guaranteed, there is at least one example of a platform where this gets compiled to `sample_app_tbl.o` instead of `sample_app_tbl.c.o`.

**To Reproduce**
Attempt to build on a platform that does not follow the expected pattern.  Tables will fail to build with a failure to open the object file due to the name mismatch.

**Expected behavior**
Table should build successfully.

**Code snips**
I traced this back to a workaround that was put in for older CMake, which assumes the name of the object file:
https://github.com/nasa/cFE/blob/73c338d544f85cb0c1a97b743d3ae9f7a35039b5/cmake/arch_build.cmake#L200-L201

The preferred/correct way to do this is to use the `$<TARGET_OBJECTS:...>` generator expression, but this only works in newer CMake versions.  The workaround was to assume a file name based on the expected pattern, which works most of the time, but there are exceptions.

**System observed on:**
Originally reported by @klystron78  when building for VxWorks 7.

**Reporter Info**

jphickey:
**Describe the contribution**
Use a shell glob instead of assuming a specific file name, this should pick up the object file no matter how it was named, as
long as it ends in ${CMAKE_C_OUTPUT_EXTENSION}.

It does rely on/assume there is only one such file, so this might break if one changes the name of a table without a clean rebuild.

A proper fix would still be to use TARGET_OBJECTS generator expression, but that can't happen until Ubuntu 18.04 (w/ ver 3.10) is EOL.

Fixes #1436

**Testing performed**
Build and sanity check CFE, confirm tables are built successfully

**Expected behavior changes**
Tables should build even if the file name pattern is not as expected.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/cmake/arch_build.cmake
+++ b/cmake/arch_build.cmake
@@ -197,9 +197,6 @@ function(add_cfe_tables APP_NAME TBL_SRC_FILES)
       add_library(${TGT}_${TBLWE}-obj STATIC ${TBL_SRC})
       target_link_libraries(${TGT}_${TBLWE}-obj PRIVATE core_api)
 
-      get_filename_component(TBLOBJ ${TBL} NAME)
-      string(APPEND TBLOBJ ${CMAKE_C_OUTPUT_EXTENSION})
-
       # IMPORTANT: This rule assumes that the output filename of elf2cfetbl matches
       # the input file name but with a different extension (.o -> .tbl)
       # The actual output filename is embedded in the source file (.c), however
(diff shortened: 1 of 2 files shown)


jphickey:
Marking this as draft for now, need coordination w/ @klystron78 to test this on the VxWorks7 build where the issue was observed

astrogeco:
[CCB:2021-05-05](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-May-05-2021)

astrogeco:
@jphickey is this ready?

PR #1451, opened 2021-04-29
Fix #1343, OSAL config file simplification

### LINKED ISSUE #1343: Update the cmake/sample_defs/default_osconfig.cmake to be just overrides/changes from osal/default_config.cmake
default_osconfig.cmake repeats many of the settings with the same selection as in default_config.cmake, confusing dependency

Remove duplication, document in default_osconfig.cmake that it's for overriding values in default_config.cmake.

None

**Additional context**
Came up when trying to figure out why OSAL coverage test results were different at the bundle level vs osal level, wasn't clear that default_config.cmake was being used and hard to sort out what's actually being changed in default_osconfig.cmake

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
The CFE framework build uses mostly default values for OSAL compile time configuration.  There are just a few exceptions where the CFE build is tuned for a more debug-friendly result.

This updates the configuration files in "sample_defs" to reflect only those values which are set to a non-default value, along with
information about why the configurable item is changed. Currently, these are only items related to debugging.

This change also includes more documentation/comments about how this configuration facility works and where to find information on the complete set of options available (which is OSAL version dependent).

Fixes #1343 

**Testing performed**

**Expected behavior changes**
None, this is just documentation/sample config cleanup, but does not change the effective OSAL configuration, since most of the values in this sample file were the same as the default value.

**System(s) tested on**
Ubuntu 20.04

### DIFF
--- a/cmake/sample_defs/default_osconfig.cmake
+++ b/cmake/sample_defs/default_osconfig.cmake
@@ -1,210 +1,54 @@
 ##########################################################################
 #
-# CFE configuration options for OSAL
+# CFE-specific configuration options for OSAL
 #
-# This file specifies the default values for various compile-time options
-# supported by OSAL.  These options can be further tuned by the specific
-# OSAL and BSP selection, as well as the user application.
+# This file specifies the CFE-specific values for various compile options
+# supported by OSAL.
+#
+# OSAL has many configuration options, which may vary depending on the
+# specific version of OSAL in use.  The complete list of OSAL options,
+# along with a description of each, can be found OSAL source in the file:
+#
+#    osal/default_config.cmake
+#
+# A CFE framework build utilizes mostly the OSAL default configuration.
+# This file only contains a few specific overrides that tune for a debug
+# environment, rather than a deployment environment.
+#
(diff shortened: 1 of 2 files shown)


skliper (approved):

astrogeco:
[CCB:2021-05-05](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-May-05-2021) APPROVED

jwilmot:
Talked with Joe. Concern is that an inexperienced user might leave OSAL_CONFIG_UTILITYTASK_PRIORITY set high and see  intermittent bad/non-deterministic real-time behavior. Maybe even losing a mission. You want to be very clear that setting a high priority is a bad idea for real-time flight systems.

jphickey:
Pushed update to resolve @jwilmot's concerns - this just moves the setting of the `OSAL_CONFIG_UTILITYTASK_PRIORITY` value to the `native_osconfig.cmake` file instead of the `default_osconfig.cmake` file.  This way it only applies when building with `SIMULATION=native` flag (i.e. debug builds for the local workstation).  For any other normal build, it will use the default priority (low).

Issue #1393, opened 2021-04-20
Move `CFE_FS_RunBackgroundFileDump` to internal API


skliper:
**Is your feature request related to a problem? Please describe.**
`CFE_FS_RunBackgroundFileDump` is only used internally, shouldn't be exposed as external API

**Describe the solution you'd like**
Move to *_core_internal.h

**Describe alternatives you've considered**
None

**Additional context**
Code review, once moved would also benefit from #1387

**Requester Info**
Jacob Hageman - NASA/GSFC

Issue #1387, opened 2021-04-20
Move internal API's and stubs to core_private


skliper:
**Is your feature request related to a problem? Please describe.**
Currently the *core_internal.h files that define cFE internal APIs are in core_api along with stubs.

**Describe the solution you'd like**
Move these to core_private since they don't need to be exposed outside of cFE

**Describe alternatives you've considered**
None

**Additional context**
Observations related to code review CFS-41, but not directly related to a comment so not marking w/ label.

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
Closed by #1463 (confirmed in main, not sure why github didn't close issue)

PR #1463, opened 2021-04-30
Fix #1417, generated coverage stubs for CFE core

### LINKED ISSUE #1417: Generate new set of CFE coverage test stubs
To achieve max flexibility the CFE coverage test stubs should have no built-in logic, with all handler logic in a separate function.

This is the pattern that was introduced with nasa/osal#966, along with the stub generator script.

Run the stub generator script on CFE core API headers to generate a new set of "pure" stub implementation files with no extra logic in them.
Migrate any existing post-hook logic in the old stub to be a default handler function, as was done for OSAL stubs.

**Additional context**
Extension of original request in nasa/osal#832, but for CFE.  Depends on nasa/osal#966 to be accepted/merged before this can be done.

**Requester Info**

jphickey:
**Describe the contribution**
Update CFE core stub libraries to use generated stubs, using the generator script now part of UT Assert.  All existing stub logic is converted to a default handler routine.

This also entailed moving all of the internal API prototypes to be in the "core_private" interface lib rather than in "core_api".  The
only reason this was not done earlier was due to issues with the stubs, but that is alleviated when using generated stubs.

Fixes #1417

**Testing performed**

**Expected behavior changes**
No impact to FSW.
For unit testing, this allows the stub handler to be completely replaced by a test case, bypassing any default handler logic, if the default handler function is not desired.

**System(s) tested on**
Ubuntu 20.04

**Additional Context**
Originally when the directory reorganization was done (#972) the "core internal" function prototypes were kept in `core_api` because that's where the stubs were, and splitting them would require splitting the stub library.

When using generated stubs, this is no longer an issue, and actually flips the problem around - such that to maintain the patterns these header files should be moved so that the stubs are generated properly with the right scope.

As a result - this moves the "internal" header files from core_api to core_private, but does not change them.  This is where they really should have been all along.


skliper:
Does this also fix #1393 and #1387?  If so, link.

astrogeco:
[CCB:2021-05-05](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-May-05-2021) APPROVED
- Extends "handler function" concept from osal to cfe

astrogeco (changes requested):
Rename `*hooks.c` as `*handler.c` where appropriate

jphickey:
Rebased to main to resolve conflict, and updated to rename hooks to handlers (see commit 2d2a7f91)

skliper:
Note this also fixed #1055

PR #1520, opened 2021-05-13
Fix #1505, accept "NULL" as entry point

### LINKED ISSUE #1505: "NULL" entry no longer supported for libraries in startup script
**Describe the bug**
Set sample_lib entry point to NULL:
```
OS_GenericSymbolLookup_Impl():118:Error: NULL: ./cf/sample_lib.so: undefined symbol: NULL
1980-012-14:04:03.48915 ES Startup: Could not find symbol:NULL. EC = 0xFFFFFFFF
```

**To Reproduce**
See above.

**Expected behavior**
NULL should skip the symbol lookup and load.  Note it does skip if the field is just left empty, as in:
`CFE_LIB, /cf/fs_lib.so, , FS_LIB, 0, 0, 0x0, 0;`

**Code snips**
Only skips if empty string:
https://github.com/nasa/cFE/blob/82c1bd4d1c8937deff801cfebf2698866228ea6a/modules/es/fsw/src/cfe_es_apps.c#L453
Probably could just convert "NULL" to empty in the CFS_LIB processing section.

**System observed on:**
 - Hardware: Docker
 - OS: Ubuntu 18.04
 - Versions: Bundle main

**Additional context**
Found by @excaliburtb

**Reporter Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
Recognize the special string "NULL" to indicate no entry point should be called for the library.  Equivalent to leaving the
field empty. 

Fixes #1505 

**Testing performed**
Build and sanity check CFE, and use a startup script that loads a CFE_LIB with a NULL entry point
Run all unit tests.

**Expected behavior changes**
Specifying the special string `NULL` as the entry point in a startup script should work as it did in previous versions - skips the entry point.

**System(s) tested on**
Ubuntu

### DIFF
--- a/modules/es/fsw/src/cfe_es_apps.c
+++ b/modules/es/fsw/src/cfe_es_apps.c
@@ -450,7 +450,8 @@ int32 CFE_ES_LoadModule(CFE_ResourceId_t ParentResourceId, const char *ModuleNam
     /*
      * If the Load was OK, then lookup the address of the entry point
      */
-    if (ReturnCode == CFE_SUCCESS && LoadParams->InitSymbolName[0] != 0)
+    if (ReturnCode == CFE_SUCCESS && LoadParams->InitSymbolName[0] != 0 &&
+        strcmp(LoadParams->InitSymbolName, "NULL") != 0)
     {
         StatusCode = OS_ModuleSymbolLookup(ModuleId, &InitSymbolAddress, LoadParams->InitSymbolName);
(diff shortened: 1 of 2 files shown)


skliper (approved):

astrogeco:
[CCB:2021-05-19](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-May-19-2021) APPROVED

PR #1549, opened 2021-05-18
Fix #1538, add capability to generate multiple tables

### LINKED ISSUE #1538: Support multiple/alternative table definitions
Some users have requested the ability to generate multiple different table files/configurations as part of the build, to represent different hardware configurations.  This way, they can choose the active configuration at install time (by choosing which file gets copied to the target) or at run time, by loading a different table file.

Currently the "add_cfe_tables" does not directly support this, but with a little work it should be able to.  This requires adding an "install" hook so the user has a proper place to call this function to generate customized table files.

While this is possible to some degree with the current build system, all decisions are made at build time.  For instance the C preprocessor can be used (via e.g. add_definitions) to select different table values based on build-time configurations.  Alternatively, the user can write their own script to compile a C file and call `elf2cfetbl` on it, but it can be difficult to determine the proper include paths and compile definitions.

**Requester Info**

jphickey:
Adds a "install_custom.cmake" hook that can be put into a CPU-specific subdirectory under the "defs" directory, that can perform extra installation steps as required for the project/mission.

Tweaks the "add_cfe_tables" function such that it can also be called from the install_custom.cmake script to generate additional/alternative table binary files for that CPU.

The main update is that it uses the "APP_NAME" parameter to associate the table files with the app library, so the same set of include files can be used.  This relies on the target-scope properties being used.  Historically that string wasn't verified, it could have been any unique string, but now it should match the app if this is to work as expected.

**Describe the contribution**
Fixes #1538 

**Testing performed**
Add multiple/alternative table definitions for sample_app in a local 2-cpu configuration.  Confirm that all table files are built and installed to the staging area correctly, and that sample_app can be tweaked to load one of the alternatives, just by changing the filename it loads.

**Expected behavior changes**
Because the "APP_NAME" (1st parameter) of `add_cfe_tables` was never validated historically, it just had to be unique, this still allows any string for backward compatibility.  It will generate a new "Note" message to the user if it does _not_ match an application name, but it is still accepted.

If the multiple table feature is used, it actually needs to match the application name, or else the include paths may be incomplete.

**System(s) tested on**
Ubuntu

**Additional context**
See also nasa/sample_app#148

### DIFF
--- a/cmake/arch_build.cmake
+++ b/cmake/arch_build.cmake
@@ -148,80 +148,100 @@ endfunction(add_cfe_app_dependency)
 #
 function(add_cfe_tables APP_NAME TBL_SRC_FILES)
 
-  # The table source must be compiled using the same "include_directories"
-  # as any other target, but it uses the "add_custom_command" so there is
-  # no automatic way to do this (at least in the older cmakes)
+    if (TGTNAME)
+        set (TABLE_TGTLIST ${TGTNAME})
+    else()
+        set (TABLE_TGTLIST ${TGTLIST_${APP_NAME}})
+    endif()
 
-  # Create the intermediate table objects using the target compiler,
-  # then use "elf2cfetbl" to convert to a .tbl file
-  set(TBL_LIST)
-  foreach(TBL ${TBL_SRC_FILES} ${ARGN})
+    # The first parameter should match the name of an app that was
+    # previously defined using "add_cfe_app".  If target-scope properties
+    # are used for include directories and compile definitions, this is needed
+    # to compile tables with the same include path/definitions as the app has.
+    # However historically this could have been any string, which still works
+    # if directory-scope properties are used for includes, so this is not
+    # an error.
+    if (NOT TARGET ${APP_NAME})
+        message("NOTE: \"${APP_NAME}\" passed to add_cfe_tables is not a previously-defined application target")
+    endif()
 
-    # Get name without extension (NAME_WE) and append to list of tables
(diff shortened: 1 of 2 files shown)


skliper:
Ping @excaliburtb, does this meet your need for building multiple versions of tables?

excaliburtb:
This looks like what I would want. Is there a way to add include paths to a specific table?

jphickey:
> This looks like what I would want. Is there a way to add include paths to a specific table?

You can always call the `target_compile_definitions` and/or `target_include_directories` cmake commands to configure how these table sources get compiled.  This does require you to know the actual target name, which in this case is a combination of the target name, app name, and table basename (i.e. without extension).  That is, you can do:

`target_include_directories(${TGT}_${APP_NAME}_${TBLWE} ${extra_includes})`

astrogeco:
[CCB:2021-05-19](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-May-19-2021) APPROVED
- Lets users install arbitrary files by calling "add cfe tbl"
- Renamed library name to include the TGT_NAME and APP_NAME. Previously it could be an arbitrary string, it is still acceptable but it might result in a build failure.
  - There's an open issue in sample_app <https://github.c<om/nasa/sample_app/issues/148> to make it compatible with this pattern
- Do we want to add a sample implementation? 
  - [ ] open issue 
- Opened a ticket on elf2cfetbl to enforce naming <https://github.com/nasa/elf2cfetbl/issues/79>

PR #1555, opened 2021-05-20
Fix #1554, add doxygen aliases for OSAL parameter/retvals

### LINKED ISSUE #1554: Documentation tag to indicate return vals that are not tested in functional test
The pattern employed in OSAL (and CFE) is that functions should list all directly-generated return values from an implementation as `@retval` tags in doxygen markup.  These are then cross referenced between the test logs to ensure that all documented return value are tested, and all tested return values are documented.

Coverage test should be able to exercise all values all the time, but functional test may not be able to exercise all of them.  In particular the ones which depend on a failure of the underlying system call are not likely to be trigger-able from a functional test environment.

These return codes should be marked in the doxygen as being "coverage test only".  This will capture the fact that this has been checked/examined and the functional test is not expected to reproduce this value.

The retval statements could be entirely removed if they are not trigger-able in a functional test environment.  (Many are pass-through values from the low level impl, but probably not all of them)

**Additional context**
This info is valuable not only for this test/documentation round but also any future audits of these test cases by projects and/or future releases.

**Requester Info**

jphickey:
**Describe the contribution**

Adds "nonnull", "nonzero", and "covtest" tags to mark parameters and return values in documentation.  This info is helpful when
auditing the test cases.

Fixes #1554

**Testing performed**
Build osalguide documentation, confirm output in the generated HTML

**Expected behavior changes**
Documentation only, no FSW.
Doxygen parameters and retvals can now be marked accordingly

**System(s) tested on**
Ubuntu

**Additional context**
This markup will be required for some of the OSAL return value verifications

### DIFF
--- a/cmake/osal-common.doxyfile.in
+++ b/cmake/osal-common.doxyfile.in
@@ -16,6 +16,10 @@ ABBREVIATE_BRIEF       = "The $name class " \
                          the
 TAB_SIZE               = 8
 OPTIMIZE_OUTPUT_FOR_C  = YES
+ALIASES               +=  nonnull="(must not be null)"
+ALIASES               +=  nonzero="(must not be zero)"
+ALIASES               +=  covtest="(return value only verified in coverage test)"
+
 #---------------------------------------------------------------------------
 # Build related configuration options
 #---------------------------------------------------------------------------


jphickey:
If anyone has suggestions for a better tag name than "covtest" please suggest.

skliper (approved):

zanzaben:
> If anyone has suggestions for a better tag name than "covtest" please suggest.

Something that relates to how those errors are produced by underlying system would be better. So for example: "underlying" or "system" maybe.

jphickey:
actually no need to rush this one, see #1563 and nasa/osal#1033 instead which moves these templates/configs into OSAL, which is a preferable solution.

jphickey:
This may be needed after all, since the CFE users guide also references the OSAL headers.  So the extra nonnull/nonzero/covtest tags need to be also included in the CFE run to avoid warnings.

jphickey:
Merged into integration-candidate in 51ecd53

PR #1568, opened 2021-05-24
cFE Integration candidate: 2021-05-25

astrogeco:
## Description

### DIFF
--- a/modules/cfe_testcase/src/es_task_test.c
+++ b/modules/cfe_testcase/src/es_task_test.c
@@ -86,7 +86,8 @@ void TestCreateChild(void)
 
     OS_TaskDelay(500);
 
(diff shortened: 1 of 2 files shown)


astrogeco:
@jphickey looks like 1549 broke SCH_LAB, see https://github.com/nasa/cFS/runs/2658714501?check_suite_focus=true

jphickey:
I will check into it

jphickey:
@astrogeco  - Corrected as hotfix in commit d7073fa2.

I had originally patched my build to use the correct app names in the sch_lab and to_lab table builds.  When I set the name back it caused no table file to be built.  Fixed in the above commit by referencing the original list variable in that case.

jphickey:
Also worth noting, I submitted nasa/sch_lab#80 to correct for the infinite loop observed here.  Missing a table file is a condition that should cause SCH_LAB to simply exit with an error - not a continuous spew of error events.

PR #1593, opened 2021-06-01
Fix #1544, add time get reference error bit

### LINKED ISSUE #1544: Improve `CFE_TIME_GetReference` error handling
Hard coding Retry count to 4 is a bit "magic", not clear how much margin this has, not clear statistically how may conflicts are occurring, no error reporting, no way for the calling routine to take action or track, not clear even if these values are always required to be in sync or what happens if they aren't, not obvious what is intended here.

Clarify design, track/monitor/report performance/errors.  If there's uses where it's critical, may need to deconflict (protect query from update).  Maybe provide API that's slower but always correct, vs faster but possibly invalid if that's really a need.

None

**Additional context**
Code review

**Requester Info**
Jacob Hageman - NASA/GSFC

jphickey:
**Describe the contribution**
Use one of the unused time state bits to indicate if an error has occurred where CFE_TIME_GetReference was not able to get a consistent copy of the reference state data.

In a functional system this should never occur - there should be at most one retry, which only happens in the event there was a burst of updates (more than 4) concurrently with reading the structure.

The previous implementation did not report or handle the condition at all, this at least sets a TLM status bit and returns a reference struct filled with all zeros.

Fixes #1544 

**Testing performed**
Check coverage report to ensure the new tests are being executed and confirm that the read error flag gets set, and the output state is all 0 in that case.
Force the error to occur during a syslog call and confirm the timestamp gets set, e.g.

    1980-001-00:00:00.00000 ES Startup: Opened ES App Startup file: /cf/cfe_es_startup.scr

**Expected behavior changes**
If CFE_TIME_GetReference was ever not able to get a valid reference structure, it outputs all zeros and the "REFERR" bit (0x0010) will be set in the output TLM structure until the system is rebooted.

**System(s) tested on**
Ubuntu

**Additional context**
This solution was chosen over an event/syslog because both of those will also get the time, thereby introducing a recursive loop if the error happens again.  

By zero-filling the structure the output is also identifiable; i.e. if the date "1980-001-00:00:00.00000" ever appears in a log file, failure of CFE_TIME_GetReference is really the only way a timestamp like that is possible.  Without this change the time data would be variable/unpredicatable.

### DIFF
--- a/modules/time/fsw/inc/cfe_time_msg.h
+++ b/modules/time/fsw/inc/cfe_time_msg.h
@@ -717,7 +717,9 @@
 #define CFE_TIME_FLAG_ADDTCL 0x0080 /**< \brief Time Client Latency is applied in a positive direction */
 #define CFE_TIME_FLAG_SERVER 0x0040 /**< \brief This instance of Time Services is a Time Server */
 #define CFE_TIME_FLAG_GDTONE 0x0020 /**< \brief The tone received is good compared to the last tone received */
-#define CFE_TIME_FLAG_UNUSED 0x001F /**< \brief Reserved flags - should be zero */
+#define CFE_TIME_FLAG_REFERR \
+    0x0010 /**< \brief GetReference read error, will be set if unable to get a consistent ref value */
+#define CFE_TIME_FLAG_UNUSED 0x000F /**< \brief Reserved flags - should be zero */
 /** \} */
 
 /*************************************************************************/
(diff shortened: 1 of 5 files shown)


skliper (approved):

astrogeco:
[CCB:2021-06-02](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-June-02-2021) APPROVED

PR #1598, opened 2021-06-03
Fix #1299, Updated FS Read/WriteHeader API return documentation

### LINKED ISSUE #1299: CFE_FS_ReadHeader & WriteHeader api headers state wrong returns
**Describe the bug**
In the header api's for ReadHeader and WriteHeader they both say that they return the execution status, however the implementation has them return the size of the file they read/wrote if successful and an error code if not. 

**Reporter Info**
Alex Campbell GSFC

skliper:
**Describe the contribution**
Fix #1299 - updated return documentation for CFE_FS_ReadHeader and CFE_FS_WriteHeader to state they return bytes read/written or error code.  Left as CFE_Status_t since they do return error codes, although overloaded.  It is confusing since 0 actually indicates the data wasn't read or written, but that would require an API change.  Related issue is #483.

**Testing performed**
CI - documentation only 

**Expected behavior changes**
None, documentation only

**System(s) tested on**
CI

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/modules/core_api/fsw/inc/cfe_fs.h
+++ b/modules/core_api/fsw/inc/cfe_fs.h
@@ -65,7 +65,7 @@
 **                         filled with the contents of the Standard cFE File Header. *Hdr is the contents of the
 **                         Standard cFE File Header for the specified file.
 **
-** \return Execution status, see \ref CFEReturnCodes
+** \return Bytes read or error status, see \ref CFEReturnCodes
 **
 ** \sa #CFE_FS_WriteHeader
 **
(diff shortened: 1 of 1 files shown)


zanzaben (dismissed):
Both can return OSReturnCodes.

skliper:
> Both can return OSReturnCodes.

See #1599, will address generally as part of that resolution.

zanzaben (approved):

jphickey (approved):
Only minor nitpick is that the documentation doesn't say what "success" is.  There is still only one "success" return value to expect here, except it is `sizeof(CFE_FS_Header_t)` instead of `CFE_SUCCESS`.

I'm not sure its worth changing though, because in the next version I'd rather fix this to return CFE_SUCCESS like everything else does, rather than continuing to return a size.

astrogeco:
[CCB:2021-06-09](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-June-09-2021) APPROVED

PR #1601, opened 2021-06-03
Fix #567, Document CFE_ES_RunLoop increment task counter behavior

### LINKED ISSUE #567: CFE_ES_RunLoop increments the main task ExecutionCounter rather than the task that called it

Inside the `CFE_ES_RunLoop()` function itself, it looks up the MainTaskId of the caller and only increments _that_ execution counter, not the execution counter of its own task record.  This means no matter what child task actually calls this function, it implements the execution counter of the main task only.

Note that if the real main task is doing something else and also increments its own task counter, this is a race condition.

I suggest one of the following:

1. Just have `CFE_ES_RunLoop()` invoke `CFE_ES_IncrementTaskCounter()` to increment the counter for the task from which it was called. So if it gets called from a child task, then that child task gets incremented, not the main task.  This is at least straightforward/consistent and avoids the race condition.
2. Maintain a separate "app" exec counter which is incremented by CFE_ES_RunLoop (only), and use the `CFE_ES_IncrementTaskCounter` to account for other regular task activity.

**Additional context**
Noticed this when fixing #480 and it seemed rather odd/incorrect to be storing the ExecutionCounter where it is.  This causes the code to jump to other entries in the table for the sole purpose of reading/updating this value, when it already had the correct app record to start.

**Requester Info**

skliper:
**Describe the contribution**
Fix #567 - updated documentation

**Testing performed**
CI - documentation only

**Expected behavior changes**
None

**System(s) tested on**
CI

**Additional context**
None

**Third party code**
None

Jacob Hageman - NASA/GSFC

### DIFF
--- a/modules/core_api/fsw/inc/cfe_es.h
+++ b/modules/core_api/fsw/inc/cfe_es.h
@@ -355,7 +355,9 @@ void CFE_ES_ExitApp(uint32 ExitStatus);
 **         the system.
 **
 ** \par Assumptions, External Events, and Notes:
-**          None
+**        This API updates the internal task counter tracked by ES for the calling task.
+**        For ES to report application counters correctly this API should be called
+**        from the main app task as part of it's main processing loop.
 **
 ** \param[in]  RunStatus    A pointer to a variable containing the Application's
 **                          desired run status.  Acceptable values are:


jphickey (approved):

zanzaben:
The App Developer's Guide should also be updated.

zanzaben (changes requested):
The App Developer's Guide should also be updated.

skliper:
> The App Developer's Guide should also be updated.

Where specifically?

skliper:
> The App Developer's Guide should also be updated.

The developer's guide has CFE_ES_RunLoop in the sample code and the API documents use.  No further updates identified.  Should be sufficient for closing current issue.  Open a new issue (or provide explicit requested changes here) for additional changes.

zanzaben (approved):

astrogeco:
[CCB:2021-06-09](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-June-09-2021) APPROVED

PR #1624, opened 2021-06-16
Fix #596, scrub all UT_Report calls

jphickey:
**Describe the contribution**
Scrub through all UT reporting calls and replace with preferred macro where possible.

- All calls to check status/return code are replaced with macro that logs all arguments and return value
- All calls that involved multiple conditions AND'ed together are replaced with individual asserts on each condition.

Fixes #596
Fixes #470 

**Testing performed**
Build and run unit tests, confirm logging is as intended, and coverage level is not reduced

**Expected behavior changes**
Better, more complete context information is logged into the file
All individual assert conditions are logged separately (no more checking of multiple AND'ed values/conditions in a single report)

**System(s) tested on**
Ubuntu

**Additional context**
This also removes all the "free-form" info that was in the logs, since it wasn't really providing value (a developer will go to the source file/line number and see full context, nor was it complete or consistent enough to provide any value for cross referencing or documentation).

**Contributor Info - All information REQUIRED for consideration of pull request**
Joseph Hickey, Vantage Systems, Inc.

PR #1627, opened 2021-06-16
Fix #1626, rename/clean CFE coverage assert macros

### LINKED ISSUE #1626: Consistency in CFE coverage test helper macros
The following helper macros are defined in `ut_support.h` and used in sb_UT.c, with the intent of applying to other modules:

- SETUP
- ASSERT
- ASSERT_EQ
- ASSERT_TRUE
- EVTCNT
- EVTSENT

These macros are all specific to/tuned toward CFE coverage testing (i.e. they are not generic) because they embed certain CFE patterns/assumptions, i.e. SETUP and TEARDOWN check for CFE_SUCCESS, the "EVENT" macros deal with the UT stub event capture logic, etc.  

The genuinely CFE-specific macros should be updated with a `CFE_UtAssert_` prefix to indicate that they are assertion statements that are specific to CFE.

For macros which are actually generic, such as "ASSERT_EQ", consider replacing this with the existing generic `UtAssert_INT32_EQ` macro.

Also for wider coverage test applicability and consistency of logging output, should also add the following:

- `VOIDCALL` macro to consistently log functions that do not have a return value, but still want to include the full text of what was called
- `RESOURCEID_EQ` macro to check for equality between resource ID values (INT32_EQ is not sufficient here because one should not assume that IDs are integers)

**Additional context**
Discussed in CCB 2021-06-16, as part of issue #596 and improving the test log output

**Requester Info**

jphickey:
**Describe the contribution**
Rename CFE coverage test assert macros in ut_support.h to have consistent name prefix.

Also adds a "VOIDCALL" and "RESOURCEID_EQ" macro for logging void functions and ID checks, respectively.

Fixes #1626 

**Testing performed**
Build and run all unit tests

**Expected behavior changes**
None - just renames macros for consistency, and simplifies the implementation of some (reusing existing macro where feasible).

**System(s) tested on**
Ubuntu

**Additional context**
Discussed initially in CCB on 2021-06-16, as part of discussion on #1624, then split off into a separate issue/PR.


jphickey:
This should be good to go, the "Build cFE documentation" workflow issue is not related, and should be corrected by #1630

skliper (approved):

pepepr08 (approved):

jphickey:
After getting into some of the table services use cases, I'm finding it would be better/more useful if these macros had a pass/fail return value like the OSAL ones do.  The problem, particularly with SETUP, is that there is no way to tell if your setup failed and if therefore e.g. a loop should stop.

jphickey:
Re-requesting review, this version seems to work well and I confirmed this set of macros should be sufficient for all CFE core test cases.

Biggest difference between this and the previous set is that the "impl" functions (underneath the macro) return the pass/fail status of the test case as a `bool`.  There are some CFE test cases, particularly in TBL, that need to break out of a loop if setup fails, and this pattern allows that to happen.

Note to reviewers, the actual changes are in `ut_support.h` and `ut_support.c` ... the changes to all other modules in this PR should be a straight search and replace just to use the new macro names, nothing else.  Actual changes/improvement to test cases comes in the next PR.

jphickey:
Pushed again to resolve formatting check issues ....

pepepr08 (approved):

astrogeco:
DEVS:2021-06-21, We need a "gold standard" way of doing things. 
- the benefit of macros is that they prevents copy-
-paste errors

skliper (approved):

astrogeco:
[CCB:2021-06-23](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-June-23-2021) APPROVED

PR #1674, opened 2021-07-19
Fix #1667, CFE_SB_MsgHdrSize returns size_t but still attempts to ret…

### LINKED ISSUE #1667: CFE_SB_MsgHdrSize returns size_t but still attempts to return CFE status code
**Describe the bug**
The `CFE_SB_MsgHdrSize()` function returns a `size_t` value.  Therefore, there is no provision/possibility for an error to be returned.

However, if passed a NULL argument, this still attempts to return `CFE_SB_BAD_ARGUMENT` which will be implicitly converted to a `size_t` value - which is platform-defined.  Notably, since `size_t` is unsigned, the result is always positive (success range) and will not be correctly interpreted by the caller.

**To Reproduce**
Call `CFE_SB_MsgHdrSize()` with a NULL `MsgPtr` argument.

**Expected behavior**
This function should probably return 0 instead of `CFE_SB_BAD_ARGUMENT` in this case.
Cannot return CFE status codes from a function that does not return `CFE_Status_t` (or `int32`).

**Code snips**
https://github.com/nasa/cFE/blob/2afdbc16e360530a9b4bbe4e0dcc814f675c1acb/modules/sb/fsw/src/cfe_sb_util.c#L54-L57

**System observed on:**
Ubuntu

**Additional context**
This line was also not being covered by the coverage test (#471) and the return type mismatch prevents this from being covered properly.

**Reporter Info**

oliverhamburger:
**Describe the contribution**
Fixes Issue #1667 by changing the return type in CFE_SB_MsgHdrSize and CFE_SB_GetUserDataLength

**Testing performed**
Build and ran unit tests

**Expected behavior changes**
CFE_SB_MsgHdrSize and CFE_SB_GetUserDataLength now return 0 when *MsgPtr is NULL

**System(s) tested on**
 Ubuntu 20.04

**Additional context**

**Contributor Info**
Oliver Hamburger GSFC

### DIFF
--- a/modules/core_api/fsw/inc/cfe_sb.h
+++ b/modules/core_api/fsw/inc/cfe_sb.h
@@ -685,7 +685,7 @@ void *CFE_SB_GetUserData(CFE_MSG_Message_t *MsgPtr);
 ** \param[in]  MsgPtr  A pointer to the buffer that contains the software bus message.
 **                     This must point to the first byte of the message header.
 **
-** \return The size (in bytes) of the user data in the software bus message.
+** \return The size (in bytes) of the user data in the software bus message, returns 0 if *MsgPtr is Null
 **/
 size_t CFE_SB_GetUserDataLength(const CFE_MSG_Message_t *MsgPtr);
 
--- a/modules/core_api/fsw/inc/cfe_sb.h
+++ b/modules/core_api/fsw/inc/cfe_sb.h
@@ -685,7 +685,8 @@ void *CFE_SB_GetUserData(CFE_MSG_Message_t *MsgPtr);
 ** \param[in]  MsgPtr  A pointer to the buffer that contains the software bus message.
 **                     This must point to the first byte of the message header.
 **
-** \return The size (in bytes) of the user data in the software bus message, returns 0 if *MsgPtr is Null
+** \return The size (in bytes) of the user data in the software bus message.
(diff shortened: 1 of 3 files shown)


zanzaben (approved):

jphickey (changes requested):
Just a minor recommendation - when documenting return values, the `\returns` tag should be just a generic description, and the `\retval` tag should be used for documenting specific values.

So, I'd recommend something like:

     \returns Estimated number of bytes in the message header for the given message
     \retval 0 if an error occurs, such as if the MsgPtr argument is not valid or header type cannot be identified

(Also to be pedantic, it is `MsgPtr` that is checked for NULL, not `*MsgPtr`)

jphickey (approved):
Looks good.  I would only recommend a "squash commit" now to consolidate the 3 commits into one on this branch.  But otherwise looks good.

dmknutsen:
@oliverhamburger can you fill out a CLA?  We are trying to submit an NTR for the Draco development cycle.
https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md

oliverhamburger:
> @oliverhamburger can you fill out a CLA? We are trying to submit an NTR for the Draco development cycle. https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md

CLA sent to GSFC-SoftwareRelease@mail.nasa.gov

PR #1707, opened 2021-07-25
Fix #1690, Tests for time conversion api

### LINKED ISSUE #1690: Add functional tests for cFE Time Conversion APIs
Need open source functional tests for certifiability

Add functional tests for cFE Time Conversion APIs
    CFE_TIME_MET2SCTime - Convert specified MET into Spacecraft Time.
    CFE_TIME_Sub2MicroSecs - Converts a sub-seconds count to an equivalent number of microseconds.
    CFE_TIME_Micro2SubSecs - Converts a number of microseconds to an equivalent sub-seconds count.

N/A

**Additional context**
N/A

**Requester Info**
Alex Campbell GSFC

pavll:
**Describe the contribution**
- Fixes #1690 (New unit-tests for the Time Conversion API like described in this issue)

**Expected behavior changes**
 - No impact on behavior 

**Additional context**
Hi, I'm new to this repository and tried my hand at a "good first issue". I would be very grateful if someone who knows about this repo can have a quick look and tell me if my first commit for this issue works like this. That's why the unit test is commented so much.
Thanks already :)

**Contributor Info**
Paul Oberosler, Individual

### DIFF
--- a/modules/cfe_testcase/src/cfe_test.c
+++ b/modules/cfe_testcase/src/cfe_test.c
@@ -59,6 +59,7 @@ void CFE_TestMain(void)
     FSHeaderTestSetup();
     SBPipeMangSetup();
     TimeCurrentTestSetup();
+    TimeConversionTestSetup();
 
     /*
      * Execute the tests
--- a/modules/cfe_testcase/src/time_conversion_test.c
+++ b/modules/cfe_testcase/src/time_conversion_test.c
@@ -1,0 +1,135 @@
+/*************************************************************************
+**
+**      GSC-18128-1, "Core Flight Executive Version 6.7"
+**
+**      Copyright (c) 2006-2021 United States Government as represented by
+**      the Administrator of the National Aeronautics and Space Administration.
+**      All Rights Reserved.
+**
+**      Licensed under the Apache License, Version 2.0 (the "License");
+**      you may not use this file except in compliance with the License.
+**      You may obtain a copy of the License at
+**
+**        http://www.apache.org/licenses/LICENSE-2.0
+**
+**      Unless required by applicable law or agreed to in writing, software
+**      distributed under the License is distributed on an "AS IS" BASIS,
+**      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+**      See the License for the specific language governing permissions and
+**      limitations under the License.
+**
+** File: time_conversion_test.c
+**
+** Purpose:
+**   Functional test of basic Time Conversion APIs
+**
+*************************************************************************/
+
+/*
+ * Includes
+ */
+#include "cfe_test.h"
+
+void TestConvertMET2SCTime(void)
+{
+    UtPrintf("Testing: CFE_TIME_MET2SCTime");
+
+    /* Mission Elapsed Time */
+    CFE_TIME_SysTime_t MET;
+    /* MET + SCTF */
+    CFE_TIME_SysTime_t TAI;
+    /* MET + SCTF - Leap Seconds */
+    CFE_TIME_SysTime_t UTC;
+    /* Spacecraft Time */
+    CFE_TIME_SysTime_t SCTime;
+
+    OS_time_t start;
+    OS_time_t end;
+    OS_time_t difference;
+
+    /* Print buffers */
+    char timeBuf1[sizeof("yyyy-ddd-hh:mm:ss.xxxxx_")];
+    char timeBuf2[sizeof("yyyy-ddd-hh:mm:ss.xxxxx_")];
+
+    OS_GetLocalTime(&start);
+
+    /* Get Times */
+    MET = CFE_TIME_GetMET();
+    TAI = CFE_TIME_GetTAI();
+    UTC = CFE_TIME_GetUTC();
+
+    OS_GetLocalTime(&end);
+
+    /* Convert - should produce a TAI or UTC at the moment of GetMET() */
+    SCTime = CFE_TIME_MET2SCTime(MET);
+
+    /* write Spacecraft Time into second buffer */
+    CFE_TIME_Print(timeBuf2, SCTime);
+
+    difference = OS_TimeSubtract(end, start);
+
+    /* Check conversion */
+#if (CFE_MISSION_TIME_CFG_DEFAULT_TAI == true)
+    /* SCTime is TAI format */
+
+    /* avoid unused compiler warning */
+    (void)UTC;
+
+    /* write TAI into first buffer */
+    CFE_TIME_Print(timeBuf1, TAI);
+
+    UtAssert_True(TimeInRange(SCTime, TAI, difference), "TAI (%s) = MET2SCTime (%s)", timeBuf1, timeBuf2);
+#else
+    /* SCTime is UTC format */
+
+    /* avoid unused compiler warning */
+    (void)TAI;
+
+    /* write UTC into first buffer */
+    CFE_TIME_Print(timeBuf1, UTC);
+
+    UtAssert_True(TimeInRange(SCTime, UTC, difference), "UTC (%s) = MET2SCTime (%s)", timeBuf1, timeBuf2);
+#endif
+}
+
+void TestConvertSubSeconds2MicroSeconds(void)
+{
+    UtPrintf("Testing: CFE_TIME_Sub2MicroSecs");
+
+    /* predefined amount of sub-seconds */
+    uint32 SUB = 31000;
+    /* correct micro-seconds equal to the predefined sub-seconds */
+    uint32 ExpectedMS = 7;
+    uint32 Sub2Micro;
+
+    /* run Sub2MicroSecs with predefined amount of sub-seconds and save result */
+    Sub2Micro = CFE_TIME_Sub2MicroSecs(SUB);
+
+    UtAssert_UINT32_EQ(ExpectedMS, Sub2Micro);
+}
+
+void TestConvertMicroSeconds2SubSeconds(void)
+{
+    UtPrintf("Testing: CFE_TIME_Micro2SubSecs");
+
+    /* predefined micro-seconds */
+    uint32 MS = 64512;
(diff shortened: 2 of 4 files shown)


astrogeco (changes requested):
Hi, thanks for the contribution! 

Can you change the c++ style comments `//` to use `/*` blocks?

pavll:
> Can you change the c++ style comments `//` to use `/*` blocks?

Done...

astrogeco (approved):

astrogeco:
@pavll looks like you still have some checks failing, I recommend "enabling actions" in your fork.

modules/cfe_testcase/src/time_conversion_test.c
```diff
@@ -0,0 +1,172 @@
...
+*************************************************************************/
+
+/*
+ * Includes
+ */
+#include "cfe_test.h"
+
+/* Check if its nearly possible that Time == Target */
+/* DUPLICATE! from time_current_test.c */
+bool TimeInRange(CFE_TIME_SysTime_t Time, CFE_TIME_SysTime_t Target, OS_time_t difference)
```
zanzaben, line 35: Since this function is now used in two places, it should be moved to cfe_test.h to prevent duplicate code.

modules/cfe_testcase/src/time_conversion_test.c
```diff
@@ -0,0 +1,172 @@
...
+	OS_time_t TimeMS;
+
+	/* convert */
+	SUB = CFE_TIME_Micro2SubSecs(MS);
+
+	/* manually convert example micro-seconds into OS_time_t */
+	TimeMS = OS_TimeAssembleFromNanoseconds(0, MS * 1000);
+
+	/* get sub-seconds form TimeMS and compare with SUB */
+	UtAssert_True(OS_TimeGetSubsecondsPart(TimeMS) == SUB, "micro-seconds (%lu) = Micro2SubSecs (%lu)", MS, SUB);
```
zanzaben, line 164: Need to test what happens when the number of microseconds passed in is greater than one second

modules/cfe_testcase/src/cfe_test.c
```diff
@@ -59,6 +59,7 @@ void CFE_TestMain(void)
     FSHeaderTestSetup();
     SBPipeMangSetup();
     TimeCurrentTestSetup();
+    TimeConversionTestSetup();
```
zanzaben, line 62: Minor thing but would put conversion before current, trying to keep the run order alphabetical moving forward just to have some order.

modules/cfe_testcase/src/time_conversion_test.c
```diff
@@ -0,0 +1,172 @@
...
+	uint32 SUB;
+	uint32 MS;
+	OS_time_t TimeSUB;
+
+	SUB = CFE_TIME_GetMETsubsecs();
+	MS = CFE_TIME_Sub2MicroSecs(SUB);
+
+	TimeSUB = OS_TimeAssembleFromSubseconds(0, SUB);
+
+	UtAssert_True(OS_TimeGetMicrosecondsPart(TimeSUB) == MS, "sub-seconds (%lu) = Sub2MicroSecs (%lu)", SUB, MS);
```
zanzaben, line 144: This test is calling the same methods that the function it's testing is which is not ideal. Would be better if you started with some specific subsecond value, call the function, and then compare that with the known microsecond value.

modules/cfe_testcase/src/time_conversion_test.c
zanzaben, line 164: Same issue of using the same methods to test like with sub2micro.

zanzaben (changes requested):

modules/cfe_testcase/src/time_conversion_test.c
pavll: Ok, thanks.

modules/cfe_testcase/src/cfe_test.c
pavll: I've had this before, but I thought I should test if it works with the CFE_TIME_GetTAI(), for example, before testing if the times can be converted correctly.

astrogeco:
CCB:2021-06-27, looking good, revisit after edits are done

modules/cfe_testcase/src/time_conversion_test.c
pavll: Ok, but what is the factor for calculating sub into microseconds? For example, if I have SUB = 400.

modules/cfe_testcase/src/time_conversion_test.c
pavll: Ok, i no implemented these predefined values: `MS=64512;` and the corresponding sub-seconds value: `SUB=2770770;`

pavll: Ok, shouldn't be a problem now with the predefined values.

modules/cfe_testcase/src/time_conversion_test.c
```diff
@@ -0,0 +1,150 @@
...
+    /* predefined amount of sub-seconds */
+    uint32 SUB = 400;
+    /* correct micro-seconds equal to the predefined sub-seconds */
+    uint32 MS = 93;
+    uint32 Sub2Micro;
+
+    /* run Sub2MicroSecs with predefined amount of sub-seconds and save result */
+    Sub2Micro = CFE_TIME_Sub2MicroSecs(SUB);
+
+    UtAssert_True(MS == Sub2Micro, "sub-seconds (%u) = Sub2MicroSecs (%u)", SUB, Sub2Micro);
```
zanzaben, line 125: You should use the macro "UtAssert_UINT32_EQ(MS, Sub2Micro)" instead of UtAssert_True, it will handle the print out for you. Should also change the variable name of MS to something like ExpectedMS.

modules/cfe_testcase/src/time_conversion_test.c
```diff
@@ -0,0 +1,150 @@
...
+    uint32 MS = 64512;
+    /* sub-seconds */
+    uint32 SUB = 2770770;
+    uint32 Micro2Sub;
+
+    /* convert */
+    Micro2Sub = CFE_TIME_Micro2SubSecs(MS);
+
+    /* get sub-seconds form TimeMS and compare with SUB */
+    UtAssert_True(Micro2Sub == SUB, "micro-seconds (%u) = Micro2SubSecs (%u)", MS, Micro2Sub);
```
zanzaben, line 142: Still need a second test here for when Microseconds is great then 1,000,000

zanzaben (changes requested):

modules/cfe_testcase/src/time_conversion_test.c
pavll: Oh, right.

modules/cfe_testcase/src/time_conversion_test.c
pavll: Ok, I made it now with >= 1000000 so > 999999, because this would correspond to the implementation of CFE_TIME_Micro2SubSecs.

pavll: Short question: could i use the <limits.h> header in this project?

zanzaben (changes requested):
Code looks all good. Last thing you need to do is squash your commits down to 1 with the message "Fix #1690, Add Time Conversion Functional Test"

pavll:
Ok, I just removed the TimeInRange again from the cfe_test.h header because I replaced it with CFE_Time_Compare in my ut.

pavll:
Ok, done. I squashed the commits.

zanzaben:
You can't use Time Compare there, that only returns -1, 0, or 1 so it passes as true only when the two times are different. You need to use time in range.

pavll:
Ok, I revert this.

pavll:
> You can't use Time Compare there, that only returns -1, 0, or 1 so it passes as true only when the two times are different. You need to use time in range.

Ok, done. I now use TimeInRange again.

modules/cfe_testcase/src/time_conversion_test.c
```diff
@@ -0,0 +1,135 @@
...
+#if (CFE_MISSION_TIME_CFG_DEFAULT_TAI == true)
+    /* SCTime is TAI format */
+
+    /* avoid unused compiler warning */
+    (void)UTC;
+
+    /* write TAI into first buffer */
+    CFE_TIME_Print(timeBuf1, TAI);
+
+    UtAssert_True(TimeInRange(SCTime, TAI, difference), "TAI (%s) = MET2SCTime (%s)", timeBuf1, timeBuf2);
```
zanzaben, line 81: The print out is backwards here, should be  

UtAssert_True(TimeInRange(SCTime, TAI, difference), "MET2SCTime (%s) = TAI (%s)", timeBuf2, timeBuf1);

modules/cfe_testcase/src/time_conversion_test.c
```diff
@@ -0,0 +1,135 @@
...
+#else
+    /* SCTime is UTC format */
+
+    /* avoid unused compiler warning */
+    (void)TAI;
+
+    /* write UTC into first buffer */
+    CFE_TIME_Print(timeBuf1, UTC);
+
+    UtAssert_True(TimeInRange(SCTime, UTC, difference), "TAI (%s) = MET2SCTime (%s)", timeBuf1, timeBuf2);
```
zanzaben, line 91: Also backwards printout. 

UtAssert_True(TimeInRange(SCTime, UTC, difference), "MET2SCTime (%s) = UTC (%s)", timeBuf2, timeBuf1);

zanzaben (changes requested):

modules/cfe_testcase/src/time_conversion_test.c
pavll: Ok, fixed. What a stupid mistake.

zanzaben (approved):

astrogeco:

PR #1731, opened 2021-07-29
Fix #1730, Explain FS Header offset & add offset functional tests.

### LINKED ISSUE #1730: Explicitly document/test file offset behavior in relation to FS file read/write APIs
It's not clear from the API where the file offset is left after using any of the FS APIs that access the file:
CFE_FS_ReadHeader
CFE_FS_WriteHeader
CFE_FS_SetTimestamp

Right now, at least some apps in the wild depend on current behavior so preferred fix for Caelum is just to document current behavior (success and error).  Error can be undefined, but success should be consistent/explicit.

Also worth adding a functional check to confirm this behavior (if it's not already checked).

Could explicitly be undefined, or change behavior to reset to the start of the file but that would ripple through all the apps.

**Additional context**
From an app code review, currently undocumented behavior means apps should all lseek but rather than change every app it's likely better to document the current behavior explicitly in the API. 

**Requester Info**
Jacob Hageman - NASA/GSFC, ping @ejtimmon

zanzaben:
**Describe the contribution**
Fixes #1730
Add explanation of offset to the header file. Add functions test for the offset. 

**Testing performed**
Build and run unit test

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04

Alex Campbell GSFC

### DIFF
--- a/modules/cfe_testcase/src/fs_header_test.c
+++ b/modules/cfe_testcase/src/fs_header_test.c
@@ -107,7 +107,7 @@ void TestTimeStamp(void)
     cFE_FTAssert_VOIDCALL(CFE_FS_InitHeader(&Header, TestDescription, CFE_FS_SubType_ES_ERLOG));
     UtAssert_INT32_EQ(CFE_FS_WriteHeader(fd, &Header), sizeof(CFE_FS_Header_t));
     UtAssert_INT32_EQ(CFE_FS_SetTimestamp(fd, NewTimestamp), CFE_SUCCESS);
-    UtAssert_INT32_EQ(OS_lseek(fd, sizeof(Header.Description), OS_SEEK_CUR), sizeof(CFE_FS_Header_t));
+    UtAssert_INT32_EQ(OS_lseek(fd, 0, OS_SEEK_CUR), (offsetof(CFE_FS_Header_t, TimeSeconds) + sizeof(NewTimestamp)));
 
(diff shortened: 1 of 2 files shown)


modules/cfe_testcase/src/fs_header_test.c
```diff
@@ -105,6 +107,8 @@ void TestTimeStamp(void)
     cFE_FTAssert_VOIDCALL(CFE_FS_InitHeader(&Header, TestDescription, CFE_FS_SubType_ES_ERLOG));
     UtAssert_INT32_EQ(CFE_FS_WriteHeader(fd, &Header), sizeof(CFE_FS_Header_t));
     UtAssert_INT32_EQ(CFE_FS_SetTimestamp(fd, NewTimestamp), CFE_SUCCESS);
+    UtAssert_INT32_EQ(OS_lseek(fd, sizeof(Header.Description), OS_SEEK_CUR), sizeof(CFE_FS_Header_t));
```
skliper, line 110: This could break if the header get's updated... how about comparing the current offset to the location after the time field by using ofsetof the time field in the header + the sizeof the time field?

skliper (changes requested):
Good work!  See suggestion to make the time offset check agnostic to header changes.

skliper (approved):

astrogeco:

PR #1737, opened 2021-07-30
Fix #1733, Move global count into test global struct.

### LINKED ISSUE #1730: Explicitly document/test file offset behavior in relation to FS file read/write APIs

### LINKED ISSUE #1733: Move "count" global to test global struct
**Describe the bug**
Should not have random global variables around, particularly variables with simple/common names, as it may silently overwrite an existing variable (i.e. if any other app/lib makes a global called "count" some OS's will just happily bind to the existing variable, not make a separate instance, when dynamic refs are resolved after loading).

https://github.com/nasa/cFE/blob/cc8c9a1d79dfdf6c098f18431d852a9e8e28592e/modules/cfe_testcase/src/es_task_test.c#L36

**Expected behavior**
Now that we have a proper global struct, `CFE_FT_Global_t` - This should be moved into it.

**Reporter Info**

zanzaben:
**Describe the contribution**
Fixes #1733 
Move global count into the test global struct

**Testing performed**
Build and run unit test

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04

Alex Campbell GSFC

### DIFF
--- a/modules/cfe_testcase/src/cfe_test.h
+++ b/modules/cfe_testcase/src/cfe_test.h
@@ -46,6 +46,7 @@
 typedef struct
 {
     CFE_FS_FileWriteMetaData_t FuncTestState;
+    int                        count;
 } CFE_FT_Global_t;
 
 /**
--- a/modules/cfe_testcase/src/cfe_test.h
+++ b/modules/cfe_testcase/src/cfe_test.h
@@ -46,8 +46,11 @@
 typedef struct
 {
     CFE_FS_FileWriteMetaData_t FuncTestState;
-    int                        count;
+    /* Generic utility counter */
+    int Count;
 } CFE_FT_Global_t;
+
+extern CFE_FT_Global_t CFE_FT_Global;
 
 /**
  * Name of log file to write
--- a/modules/cfe_testcase/src/es_task_test.c
+++ b/modules/cfe_testcase/src/es_task_test.c
@@ -33,13 +33,11 @@
 
 #include "cfe_test.h"
 
-CFE_FT_Global_t CFE_FT_Global;
-
 void TaskFunction(void)
 {
-    while (CFE_FT_Global.count < 200)
+    while (CFE_FT_Global.Count < 200)
     {
-        CFE_FT_Global.count += 1;
+        CFE_FT_Global.Count += 1;
         OS_TaskDelay(100);
     }
     return;
@@ -47,9 +45,9 @@ void TaskFunction(void)
 
 void TaskExitFunction(void)
 {
-    while (CFE_FT_Global.count < 200)
+    while (CFE_FT_Global.Count < 200)
     {
-        CFE_FT_Global.count += 1;
+        CFE_FT_Global.Count += 1;
         CFE_ES_ExitChildTask();
     }
     return;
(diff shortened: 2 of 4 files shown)


modules/cfe_testcase/src/cfe_test.h
```diff
@@ -46,6 +46,7 @@
 typedef struct
 {
     CFE_FS_FileWriteMetaData_t FuncTestState;
+    int                        count;
```
skliper, line 49: Recommend either a more useful name or add a comment as to what the intent for this variable is (is it intended as a generic utility global or specifically for use in the es task test?) another improvement would be consistent case (camel).   Helpful also to have `extern CFE_FT_Global_t CFE_FT_Global` in the header, and instantiate it once.

modules/cfe_testcase/src/es_task_test.c
```diff
@@ -33,23 +33,23 @@
 
 #include "cfe_test.h"
 
-uint32 count = 0;
+CFE_FT_Global_t CFE_FT_Global;
```
skliper, line 36: See comment above (define an extern in the header).

modules/cfe_testcase/src/es_task_test.c
```diff
@@ -135,24 +136,24 @@ void TestChildTaskDelete(void)
     size_t                     StackSize    = CFE_PLATFORM_ES_PERF_CHILD_STACK_SIZE;
     CFE_ES_TaskPriority_Atom_t Priority     = CFE_PLATFORM_ES_PERF_CHILD_PRIORITY;
     uint32                     Flags        = 0;
-    count                                   = 0;
+    CFE_FT_Global.count                     = 0;
     int ExpectedCount                       = 5;
```
skliper, line 140: declarations first (coding standard)

modules/cfe_testcase/src/es_task_test.c
```diff
@@ -67,13 +67,14 @@ void TestCreateChild(void)
     CFE_ES_TaskPriority_Atom_t Priority      = CFE_PLATFORM_ES_PERF_CHILD_PRIORITY;
     uint32                     Flags         = 0;
     int                        ExpectedCount = 5;
+    CFE_FT_Global.count                      = 0;
```
skliper, line 70: Add a gap between declarations and other code for readability.

skliper (changes requested):

skliper (approved):

astrogeco:

PR #1756, opened 2021-08-03
Fix #1755, separate variable for OSAL status

### LINKED ISSUE #1755: Improve consistency when working with OSAL status codes
Quite often in the CFE implementation, OSAL status codes are mixed with CFE status codes, often times sharing a single stack variable to hold either one, and even switching back and forth.

To improve clarity and consistency, OSAL status codes should be stored in a _separate_ stack variable from the CFE status codes.  Ideally, in the future, these should also migrate to a separate type (e.g. `osal_status_t`, as in nasa/osal#1108, or `CFE_Status_t`, as in nasa/cfe#921).  

- Anywhere that an OSAL status code is stored on the stack, make a _separate_ variable.
- Always use a consistent name for that variable, e.g. `OsStatus`  (as opposed to "Status" or "ReturnCode" etc, which are vague).
- Avoid intermixing CFE status codes and OSAL status codes whenever possible (there will still be some times its necessary, when calling generic functions).  When it is necessary, make conversions between these types explicit/obvious via comments and/or type casting.
- When printing OSAL status codes in a log or event message, print it as long decimal (%ld), not as hex as the CFE status codes are.  This is because OSAL defines its status codes as decimal numbers, so logs/events should match this.

**Additional context**
This can be considered a prerequisite to nasa/cfe#921 (cannot introduce a unique `CFE_Status_t` type until it is no longer being used to store OSAL codes).

This would be a fix for next release (not 7.0.0).

**Requester Info**

jphickey:
**Describe the contribution**
Ensure a separate stack variable, always named "OsStatus", is used to hold the result of an OSAL API call.  Do not use the same variable that is used to store a CFE status code.

This also fixes a few return code translations, where some functions were returning the OSAL status code (perhaps unintentionally) because it was stored in the same variable.

Fixes #1755

**Testing performed**

**Expected behavior changes**
Nothing significant, however some APIs which returned an OSAL status will now translate to CFE status.

**System(s) tested on**
Ubuntu


zanzaben:
several functional tests will also need to be updated for this.

skliper (approved):

astrogeco:
@jphickey check build fails

jphickey:
The build failure here was a bogus warning about a "possible uninitialized value" - in fact there is no way to get to that case statement without `OsStatus` getting initialized to something.  It appears gcc 9 optimizer (CI build) is not smart enough to see that detail, but gcc 10 optimizer (on my local machine) is - this is why I didn't see the warning myself.

At any rate, I added an explicit init of `OsStatus` and also rebased to main, should be good to go now.

astrogeco:
[CCB:2021-08-18](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-August-18-2021) 
- If it works let's merge it soon

PR #1790, opened 2021-08-10
Fix #1784, add CFE assert macros to functional test

### LINKED ISSUE #1784: Port "CFE_UtAssert_SuccessCheck" and related macros from coverage test to functional test
The CFE coverage test has a number of useful test macros:
- CFE_UtAssert_SETUP
- CFE_UtAssert_TEARDOWN
- CFE_UtAssert_SUCCESS
- CFE_UtAssert_MSGID_EQ
- CFE_UtAssert_RESOURCEID_EQ
- CFE_UtAssert_MEMOFFSET_EQ

However, these are only accessible/usable from the coverage test environment.  It would be helpful if equivalent macros/functions existed in the functional test environment.

Define the same basic set of macros in `cfe_assert.h`  so they can be used by any functional test.

**Additional context**
Due to the considerable differences in the application linkage, there currently is no (real) library that is common between the coverage test and functional test environments, aside from UtAssert itself.

Unfortunately this means there is no easy way to use the exact same definition of these macros, but as they are just wrappers around UtAssert functions so it isn't that bad to duplicate them.

**Requester Info**

jphickey:
**Describe the contribution**
Adds the following macros to CFE assert library in cfe_assert.h:

- CFE_UtAssert_STATUS_OK
- CFE_UtAssert_STATUS_ERROR
- CFE_UtAssert_RESOURCEID_EQ
- CFE_UtAssert_RESOURCEID_UNDEFINED
- CFE_UtAssert_MEMOFFSET_EQ
- CFE_UtAssert_MSGID_EQ

Fixes #1784

**Testing performed**
Build and run all tests (including with some new test cases that use the new macros) and confirm all is working as expected.

**Expected behavior changes**
None right now, these are new macros that test cases are not using yet.

**System(s) tested on**
Ubuntu

**Additional context**
Provides improved feature parity with coverage test environment, gives a common macro to use for common tests/asserts, and more consistent naming convention.

Joseph Hickey, Vantage Systems, Inc

### DIFF
--- a/modules/cfe_assert/inc/cfe_assert.h
+++ b/modules/cfe_assert/inc/cfe_assert.h
@@ -39,12 +39,117 @@
 *************************************************************************/
 #include "common_types.h"
 #include "cfe_es_api_typedefs.h"
+#include "utassert.h"
+#include "cfe_error.h"
 
 /************************************************************************
 ** Type Definitions
 *************************************************************************/
 
 typedef void (*CFE_Assert_StatusCallback_t)(uint8 MessageType, const char *Prefix, const char *OutputMessage);
+
+/*************************************************************************
+** CFE-specific assertion macros
+** (similar to macros in the CFE coverage test)
+*************************************************************************/
+
+/*****************************************************************************/
+/**
+** \brief Asserts the nominal execution of the function being tested.
+**
+** \par Description
+**        The core of each unit test is the execution of the function being tested.
+**        This function and macro should be used to test the nominal execution of the
+**        function; the expectation is that it will return CFE_SUCCESS or an
+**        unspecified positive value.
+**
+** \par Assumptions, External Events, and Notes:
+**        None
+**
+******************************************************************************/
+#define CFE_UtAssert_STATUS_OK(FN) \
+    CFE_UtAssert_StatusCheck(FN, true, UTASSERT_CASETYPE_FAILURE, __FILE__, __LINE__, #FN)
+
+/*****************************************************************************/
+/**
+** \brief Asserts the off-nominal execution of the function being tested.
+**
+** \par Description
+**        The core of each unit test is the execution of the function being tested.
+**        This function and macro should be used to test the generic off-nominal execution
+**        of the function; the expectation is that it will return an unspecified negative
+**        value.
+**
+** \par Assumptions, External Events, and Notes:
+**        This should be used in cases where a specific error for a particular condition
+**        is not known/documented.  Whenever a specific error is indicated by the documentation,
+**        tests should check for that error instead of using this.
+**
+******************************************************************************/
+#define CFE_UtAssert_STATUS_ERROR(FN) \
+    CFE_UtAssert_StatusCheck(FN, false, UTASSERT_CASETYPE_FAILURE, __FILE__, __LINE__, #FN)
+
+/*****************************************************************************/
+/**
+** \brief Macro to check CFE resource ID for equality
+**
+** \par Description
+**        A macro that checks two resource ID values for equality.
+**
+** \par Assumptions, External Events, and Notes:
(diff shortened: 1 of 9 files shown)


jphickey:
Updated per discussion in today's tag-up to remove SETUP and TEARDOWN macros.

Side note / food for thought -- my preference has always been to err toward including too much information in test logs, as you can always ignore the extra detail if it isn't relevant to your goals.  However, you can't go the other way, and use information you don't have in the log.

Therefore, I still would prefer to keep SETUP/TEARDOWN macros, as it only affects the details in the log file - which can be ignored/discarded when the goal is to show system-level functionality.  But I still do think there are cases where this detail can be useful in the log file, and you can't use it if its not there.

jphickey:
Probably also worth noting that the OSAL _functional_ tests also have UT_SETUP and UT_TEARDOWN macros for this same purpose, and this is the same type/level of functional test.  So I still think it should be there for consistency with all of our other tests.

skliper (approved):

zanzaben (dismissed):

modules/cfe_assert/inc/cfe_assert.h
```diff
@@ -39,13 +39,117 @@
...
+**        The core of each unit test is the execution of the function being tested.
+**        This function and macro should be used to test the nominal execution of the
+**        function; the expectation is that it will return CFE_SUCCESS or an
+**        unspecified positive value.
+**
+** \par Assumptions, External Events, and Notes:
+**        None
+**
+******************************************************************************/
+#define CFE_UtAssert_SUCCESS(FN) CFE_UtAssert_SuccessCheck(FN, true, UTASSERT_CASETYPE_FAILURE, __FILE__, __LINE__, #FN)
```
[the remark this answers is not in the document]
skliper: Names are hard, but macros for both == CFE_SUCCESS and not an error code seem appropriate.  Maybe CFE_UtAssert_CFE_SUCCESS does == CFE_SUCCESS, and SUCCESS is any non-error?

zanzaben (approved):

jphickey:
The latest update changes the name of the SUCCESS/NOT_SUCCESS asserts to CFE_UtAssert_STATUS_OK and CFE_UtAssert_STATUS_ERROR respectively.

modules/cfe_assert/src/cfe_assert_runner.c
```diff
@@ -69,6 +69,26 @@ static CFE_EVS_BinFilter_t CFE_TR_EventFilters[] = {
...
+        Disp = '=';
+    }
+    else
+    {
+        /* expecting non-success; result should be inverted */
+        Result = !Result;
+        Disp   = '!';
+    }
+
+    return UtAssertEx(Result, CaseType, File, Line, "%s (0x%lx) %c= CFE_SUCCESS", Text, (unsigned long)Status, Disp);
```
skliper, line 89: Log message should match the test actually performed.

skliper (changes requested):
Log message update requested

modules/cfe_assert/src/cfe_assert_runner.c
jphickey: OK, updated again to tweak the assert string output.  I also temporarily hacked a couple test cases to use this macro and get a sample output.  In my latest push (4f9d3df9) it looks like this in the final log:

```
[ PASS] 01.001 es_application_control_test.c:42 - CFE_ES_GetAppID(&TestAppId) (0x0) is OK
[ PASS] 01.002 es_application_control_test.c:44 - CFE_ES_RestartApp(CFE_ES_APPID_UNDEFINED) (0xffffffffc4000001) is ERROR
```

skliper (approved):

jphickey:
Last push was for whitespace/clang-format fixup ... 7th time is the charm on this hopefully

PR #1809, opened 2021-08-13
Fix #1807, increase SB pool max size bucket

### LINKED ISSUE #1807: Default configuration does not permit max-size SB message buffer to be allocated
**Describe the bug**
All messages need to be wrapped in a software bus message descriptor while in transit, and the software bus should be capable of accepting + passing messages up to `CFE_MISSION_SB_MAX_SB_MSG_SIZE` (configurable).

However, the default pool configuration does not allow this.  Calling `CFE_SB_AllocateMessageBuffer(CFE_MISSION_SB_MAX_SB_MSG_SIZE)` retuns NULL due to this.

**To Reproduce**
Build with default configuration, and call  `CFE_SB_AllocateMessageBuffer(CFE_MISSION_SB_MAX_SB_MSG_SIZE)` - returns NULL.

**Expected behavior**
Should return non-NULL, as `CFE_MISSION_SB_MAX_SB_MSG_SIZE` is the upper limit that should be acceptable.

**System observed on:**
Ubuntu

**Additional context**
This is because in the default SB pool configuration, there is not enough extra space in the maximum size pool block.  It is defined here:

https://github.com/nasa/cFE/blob/eb9c52319fb1f31f72159551735ad5f1365e5a79/cmake/sample_defs/cpu1_platform_cfg.h#L305

The hardcoded "+ 40" bytes is not quite enough for the descriptor on a 64-bit CPU (it might work on 32-bit, did not test)

**Reporter Info**

jphickey:
**Describe the contribution**
In the default config, the maximum bucket size for the SB pool was not quite big enough to allocate a message of `CFE_MISSION_SB_MAX_SB_MSG_SIZE`.

This increases the margin so this can be allocated again.

Fixes #1807 

**Testing performed**
Build and run all tests

**Expected behavior changes**
A call to `CFE_SB_AllocateMessageBuffer(CFE_MISSION_SB_MAX_SB_MSG_SIZE)` now succeeds and returns non-null.

**System(s) tested on**
Ubuntu

**Additional context**
On my system the `sizeof(CFE_SB_BufferD_t)` is 64, hence why 40 was not enough.  Using 128 just adds an extra margin for padding and platform differences, etc.  Users can still reduce this in their tuning if they are using every possible byte.

Note we cannot use `sizeof(CFE_SB_BufferD_t)` directly here because its a private structure.


skliper (approved):

zanzaben (approved):

astrogeco:
[CCB:2021-08-18](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-August-18-2021) APPROVED

PR #1785, opened 2021-08-10
Fixes #1776, #1770, Stop memory leak & add cds size test.

zanzaben:
**Describe the contribution**
Fixes #1776 
Delete memory pools to prevent memory leak. 

Fixes #1770
Add Test for registering CDS with same name and different size. 

**Testing performed**
Build and run unit test

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
Ubuntu 20.04

**Contributor Info - All information REQUIRED for consideration of pull request**
Alex Campbell GSFC

PR #1816, opened 2021-08-13
Fix #1815, add retroactive CFE status asserts

### LINKED ISSUE #1815: Add CFE_UtAssert macros to support tests where status is not predictable
The typical method of using `UtAssert_INT32_EQ` to validate the result of a CFE call will both invoke the function and check the result in a single macro.  However, this requires advance knowledge of what the result is _supposed to be_, before the call is made.

In some functional test circumstances, particularly where the test is not being run in a clean/isolated environment, it may not always be feasible to predict the "correct" return code from an API call before it is made.  For instance, there may be cases where a set of return values are acceptable, or the correct value depends on another aspect of system state that isn't known beforehand.

A test case may handle this currently by using a stack variable, for instance:

```
CFE_Status_t Status;

Status = CFE_SB_CreatePipe(&Pipe, ...);

<check other system state....determine that result should have been SUCCESS>

UtAssert_INT32_EQ(Status, CFE_SUCCESS);
```

While this works, the test log entry will only show the check for `Status == CFE_SUCCESS`, it will not show the full function call that was tested here (`CFE_SB_CreatePipe(&Pipe, ...)`).  To make the logs most useful, this should show the full function call that was made, and that it was CFE_SUCCESS (the fact that the assertion was retroactive doesn't really matter).

Of course this can be done with the free-form `UtAssert_True` macro, but that requires the programmer to repeat the text of the function call, and is subject to getting stale/incorrect or cut/paste errors as any "repeated" info always is.

Implement CFE UtAssert macros that decouple the function call from the expected return status value, so the test case can call the function and then retroactively/separately determine what the correct result should have been.

The CFE UtAssert library can temporarily hold the status in a temp variable.  This makes it simpler for the programmer to use, avoids issues with repeating the info, keeping things honest.  The use pattern would be:

```
/* call CFE_SB_CreatePipe() but do not actually assert specific status immediately */
CFE_UtAssert_STATUS_STORE(CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName));

<check other system state....determine that result should have been SUCCESS>

CFE_UtAssert_STATUS_WAS(CFE_SUCCESS);
```

With this pattern, the assert library holds the full text of the function between the store and assert, so it can be logged in the same format as UtAssert_INT32_EQ does, and not require the programmer to repeat it, or jump through other hoops to make sure that the call gets into the test log as it should.

Leave as is, but this imposes burdens on the programmer for corner cases (by repeating info unnecessarily, and having to be explicitly concerned with writing it to the test log to make the test count) and/or makes tests more fragile by repeating info (after cut/paste/move the text of the copy can get out of sync with the real call).  Having a false/incorrect log is often worse than not having a log at all.

**Requester Info**

jphickey:
**Describe the contribution**
Add a set of macros that decouple the function call from the expected status:

- `CFE_Assert_STATUS_STORE`
- `CFE_Assert_STATUS_MAY_BE`
- `CFE_Assert_STATUS_MUST_BE`

The first will make the function call and put the status into a temporary holding area, but will not assert on any particular
result.

The others will check the value of the holding area.

Fixes #1815

**Testing performed**
Build and run all tests
Also Add new test cases that need this paradigm and confirm macros working as expected (will be in other PRs).

**Expected behavior changes**
None, new test macros only, not used yet.

**System(s) tested on**
Ubuntu

**Additional context**
These macros should be used whenever a specific status is not predictable before the call, but other checks after the call can determine if the result was OK or not.

A typical use case for a function that might return any of 3 status values (e.g. status1, status2, status3) would be something like:

```
CFE_Assert_STATUS_STORE(...)
if (CFE_Assert_STATUS_MAY_BE(status1))
{
... do other checks after getting status1
}
else if (CFE_Assert_STATUS_MAY_BE(status1))
{
... do other checks after getting status2
}
else
{
   CFE_Assert_STATUS_MUST_BE(status3);
... do other checks after getting status3
}
```

### DIFF
--- a/modules/cfe_assert/src/cfe_assert_runner.c
+++ b/modules/cfe_assert/src/cfe_assert_runner.c
@@ -89,6 +89,68 @@ bool CFE_UtAssert_StatusCheck(CFE_Status_t Status, bool ExpectSuccess, UtAssert_
     return UtAssertEx(Result, CaseType, File, Line, "%s (0x%lx) is %s", Text, (unsigned long)Status, MatchText);
 }
 
+CFE_Status_t CFE_Assert_Status_Store(CFE_Status_t Status, const char *File, uint32 Line, const char *Text)
+{
+    const char *BaseName;
+
+    /* All this needs to do is save the code+text, will assert later */
+    CFE_Assert_Global.StoredStatus = Status;
+    strncpy(CFE_Assert_Global.StoredText, Text, sizeof(CFE_Assert_Global.StoredText) - 1);
+    CFE_Assert_Global.StoredText[sizeof(CFE_Assert_Global.StoredText) - 1] = 0;
+
+    BaseName = strrchr(File, '/');
+    if (BaseName == NULL)
+    {
+        BaseName = File;
+    }
+    else
+    {
+        ++BaseName;
+    }
+    strncpy(CFE_Assert_Global.StoredFile, BaseName, sizeof(CFE_Assert_Global.StoredFile) - 1);
+    CFE_Assert_Global.StoredFile[sizeof(CFE_Assert_Global.StoredFile) - 1] = 0;
+    CFE_Assert_Global.StoredLine                                           = Line;
+
+    /* Status code is just passed thru so the test case can check it however it needs to */
+    return Status;
+}
+
+bool CFE_Assert_Status_DeferredCheck(CFE_Status_t Status, UtAssert_CaseType_t CaseType, const char *File, uint32 Line,
+                                     const char *Text)
+{
+    bool        Result;
+    const char *ExtraTag;
+
+    if (CFE_Assert_Global.StoredText[0] == 0)
+    {
+        /* If no status was stored, then this is a bug in the test program (need to store a result first) */
+        UtAssertEx(false, UTASSERT_CASETYPE_FAILURE, File, Line, "TEST BUG: No stored status to assert (%s)", Text);
+        Result = false;
+    }
+    else
+    {
+        Result = (Status == CFE_Assert_Global.StoredStatus);
+        if (Result)
+        {
+            /* no extra tag added to "true" conditions */
+            ExtraTag = "";
+        }
+        else
+        {
+            /* if condition was false add an exta marker so user does not necessarily need to decode the string */
+            ExtraTag = " [false]";
+        }
+
+        /* This produces a log message similar to what UtAssert_INT32_EQ would produce.
+         * Note the file/line will reflect where the call was made, not where this assertion was done */
+        Result = UtAssertEx(Result, CaseType, CFE_Assert_Global.StoredFile, CFE_Assert_Global.StoredLine,
+                            "%s (%ld) == %s (%ld)%s", CFE_Assert_Global.StoredText,
+                            (long)CFE_Assert_Global.StoredStatus, Text, (long)Status, ExtraTag);
+    }
+
+    return Result;
+}
+
 void CFE_Assert_StatusReport(uint8 MessageType, const char *Prefix, const char *OutputMessage)
 {
     uint16 EventType;
--- a/modules/cfe_assert/inc/cfe_assert.h
+++ b/modules/cfe_assert/inc/cfe_assert.h
@@ -151,6 +151,101 @@ typedef void (*CFE_Assert_StatusCallback_t)(uint8 MessageType, const char *Prefi
     UtAssert_GenericUnsignedCompare(CFE_SB_MsgIdToValue(mid1), UtAssert_Compare_EQ, CFE_SB_MsgIdToValue(mid2), \
                                     UtAssert_Radix_HEX, __FILE__, __LINE__, "MsgId Check: ", #mid1, #mid2)
 
+/*****************************************************************************/
+/**
+** \brief Calls a function that returns CFE_Status_t with deferred validation
+**
+** \par Description
+**        The typical method of using #UtAssert_INT32_EQ to validate the result of a CFE call
+**        will both invoke the function and check the result in a single macro.  However, this
+**        requires advance knowledge of what the result is supposed to be, before the call is made.
+**
+**        This macro does invokes the function like the other macros do, but does _not_ check the
+**        return value.  Rather, it  stores the return the value in a local buffer for later checking,
+**        using the #CFE_Assert_STATUS_MAY_BE or #CFE_Assert_STATUS_MUST_BE macros.
+**
+** \par Assumptions, External Events, and Notes:
+**        In some functional test circumstances, particularly where the test is not being run
+**        in a clean/isolated environment, it may not always be feasible to predict the correct
+**        return code from an API call.
+**
+**        In these cases, the test program will need to check the result of the call itself,
+**        typically by storing the result on the stack and checking it for correctness by
+**        reading system state as necessary.
+**
+**        While the normal UtAssert_INT32_EQ assertion macro can still be used to retroactively check
+**        any status variable value (including a value on the stack from a previous API call), this
+**        will not include the "full text" of the API call, and so the test log will not reflect the
+**        full call details.
+**
+**        The pair of macros (CFE_Assert_STATUS_STORE and CFE_Assert_STATUS_WAS) can be used in these
+**        circumstances, to call an API function and record the full text of the call, when the expected
+**        status is not fully known, but still confirm the status was correct once it is known.
+**
+** \sa #CFE_Assert_STATUS_MAY_BE, #CFE_Assert_STATUS_MUST_BE
+**
+** \returns Actual CFE_Status_t value from the call
+*/
+#define CFE_Assert_STATUS_STORE(FN) CFE_Assert_Status_Store(FN, __FILE__, __LINE__, #FN)
+
+/*****************************************************************************/
+/**
+** \brief Retroactively check for an acceptable status value from CFE_Assert_STATUS_STORE
+**
+** \par Description
+**        The typical method of using #UtAssert_INT32_EQ to validate the result of a CFE call
+**        will both invoke the function and check the result in a single macro.  However, this
+**        requires advance knowledge of what the result is supposed to be, before the call is made.
+**
+**        This retroactive macro does _not_ invoke any function, but rather checks the stored status
+**        from a previous call to #CFE_Assert_STATUS_STORE.  It should be used for each status
+**        code that should be considered acceptable from the previous function call.
+**
+** \par Assumptions, External Events, and Notes:
+**
+**        While the normal UtAssert_INT32_EQ assertion macro can still be used to check any
+**        status variable value (including a value on the stack from a previous API call), this
+**        will not include the "full text" of the API call.  This macro is intended for those cases
+**        where it is desired to log the full text (function + arguments) of the API call, but
+**        when the call has already been made and the status value is stored in a local variable.
+**
+** \sa #CFE_Assert_STATUS_STORE, #CFE_Assert_STATUS_MUST_BE
+**
+** \returns Boolean pass/fail status
+**
+******************************************************************************/
+#define CFE_Assert_STATUS_MAY_BE(expected) \
+    CFE_Assert_Status_DeferredCheck(expected, UTASSERT_CASETYPE_FLOW, __FILE__, __LINE__, #expected)
+
+/*****************************************************************************/
+/**
+** \brief Retroactively check for a required status value from CFE_Assert_STATUS_STORE
+**
+** \par Description
+**        The typical method of using #UtAssert_INT32_EQ to validate the result of a CFE call
+**        will both invoke the function and check the result in a single macro.  However, this
+**        requires advance knowledge of what the result is supposed to be, before the call is made.
+**
+**        This retroactive macro does _not_ invoke any function, but rather checks the stored status
+**        from a previous call to #CFE_Assert_STATUS_STORE.  This should be used as the final
+**        assertion, after checking for other acceptable values via #CFE_Assert_STATUS_MAY_BE.
+**
+** \par Assumptions, External Events, and Notes:
+**
+**        While the normal UtAssert_INT32_EQ assertion macro can still be used to check any
+**        status variable value (including a value on the stack from a previous API call), this
+**        will not include the "full text" of the API call.  This macro is intended for those cases
+**        where it is desired to log the full text (function + arguments) of the API call, but
+**        when the call has already been made and the status value is stored in a local variable.
+**
+** \sa #CFE_Assert_STATUS_STORE, #CFE_Assert_STATUS_MAY_BE
+**
+** \returns Boolean pass/fail status
+**
+******************************************************************************/
+#define CFE_Assert_STATUS_MUST_BE(expected) \
+    CFE_Assert_Status_DeferredCheck(expected, UTASSERT_CASETYPE_FAILURE, __FILE__, __LINE__, #expected)
+
 /*************************************************************************
 ** Exported Functions
 *************************************************************************/
@@ -275,5 +370,45 @@ void CFE_Assert_CloseLogFile(void);
 ******************************************************************************/
 bool CFE_UtAssert_StatusCheck(CFE_Status_t Status, bool ExpectSuccess, UtAssert_CaseType_t CaseType, const char *File,
                               uint32 Line, const char *Text);
+
+/*****************************************************************************/
+/**
+** \brief Helper function for nominal CFE calls with deferred check
+**
+** \par Description
+**        This helper function will store the status into a temporary holding area,
+**        but _not_ assert on any specific value.
+**
+** \par Assumptions, External Events, and Notes:
+**        This facility should only be used by one task at a time.  Normally tests
+**        are single-threaded, and CFE assert will serialize test apps, so this is
+**        not a concern in the typical test environment.
+**
+**        However, if a particular test case uses child tasks, then the programmer must
+**        explicitly ensure that only one task uses this facility at a time.
+**
+** \returns Status value (pass through)
+*/
+CFE_Status_t CFE_Assert_Status_Store(CFE_Status_t Status, const char *File, uint32 Line, const char *Text);
+
+/**
+** \brief Helper function for nominal CFE calls with deferred check
+**
+** \par Description
+**        This helper function will assert on the status previously stored to a
+**        temporary holding area.
+**
+** \par Assumptions, External Events, and Notes:
+**        This facility should only be used by one task at a time.  Normally tests
+**        are single-threaded, and CFE assert will serialize test apps, so this is
+**        not a concern in the typical test environment.
+**
+**        However, if a particular test case uses child tasks, then the programmer must
+**        explicitly ensure that only one task uses this facility at a time.
+**
+** \returns Test pass status, returns true if status was successful, false if it failed.
+*/
+bool CFE_Assert_Status_DeferredCheck(CFE_Status_t Status, UtAssert_CaseType_t CaseType, const char *File, uint32 Line,
+                                     const char *Text);
 
 #endif /* CFE_ASSERT_H */
--- a/modules/cfe_assert/src/cfe_assert_io.c
+++ b/modules/cfe_assert/src/cfe_assert_io.c
@@ -86,10 +86,14 @@ void UT_BSP_SysLogStatusReport(uint8 MessageType, const char *Prefix, const char
 
 void UT_BSP_WriteLogFile(osal_id_t FileDesc, uint8 MessageType, const char *Prefix, const char *OutputMessage)
 {
-    char LogFileBuffer[CFE_ASSERT_MAX_LOG_LINE_LENGTH];
+    char   LogFileBuffer[CFE_ASSERT_MAX_LOG_LINE_LENGTH];
+    uint32 MsgEnabled = CFE_Assert_Global.CurrVerbosity >> MessageType;
 
-    snprintf(LogFileBuffer, sizeof(LogFileBuffer), "[%5s] %s\n", Prefix, OutputMessage);
-    OS_write(FileDesc, LogFileBuffer, strlen(LogFileBuffer));
+    if (MsgEnabled & 1)
+    {
+        snprintf(LogFileBuffer, sizeof(LogFileBuffer), "[%5s] %s\n", Prefix, OutputMessage);
+        OS_write(FileDesc, LogFileBuffer, strlen(LogFileBuffer));
+    }
 }
 
 void UT_BSP_DoText(uint8 MessageType, const char *OutputMessage)
@@ -97,45 +101,7 @@ void UT_BSP_DoText(uint8 MessageType, const char *OutputMessage)
     const char *                Prefix;
     CFE_Assert_StatusCallback_t StatusCallback;
 
-    switch (MessageType)
-    {
-        case UTASSERT_CASETYPE_ABORT:
-            Prefix = "ABORT";
-            break;
-        case UTASSERT_CASETYPE_FAILURE:
-            Prefix = "FAIL";
-            break;
-        case UTASSERT_CASETYPE_MIR:
-            Prefix = "MIR";
-            break;
-        case UTASSERT_CASETYPE_TSF:
-            Prefix = "TSF";
-            break;
-        case UTASSERT_CASETYPE_TTF:
-            Prefix = "TTF";
-            break;
-        case UTASSERT_CASETYPE_NA:
-            Prefix = "N/A";
-            break;
-        case UTASSERT_CASETYPE_BEGIN:
-            Prefix = "BEGIN";
-            break;
-        case UTASSERT_CASETYPE_END:
-            Prefix = "END";
-            break;
-        case UTASSERT_CASETYPE_PASS:
-            Prefix = "PASS";
-            break;
-        case UTASSERT_CASETYPE_INFO:
-            Prefix = "INFO";
-            break;
-        case UTASSERT_CASETYPE_DEBUG:
-            Prefix = "DEBUG";
-            break;
-        default:
-            Prefix = "OTHER";
-            break;
-    }
+    Prefix = UtAssert_GetCaseTypeAbbrev(MessageType);
 
     StatusCallback = CFE_Assert_Global.StatusCallback;
 
--- a/modules/cfe_assert/src/cfe_assert_priv.h
+++ b/modules/cfe_assert/src/cfe_assert_priv.h
@@ -136,6 +136,28 @@ typedef struct
      */
     char CurrentTestName[CFE_MISSION_MAX_API_LEN];
 
+    /* The following members support the "Deferred" assert feature */
+
+    /**
+     * Actual CFE status value from a previous function call
+     */
+    CFE_Status_t StoredStatus;
+
+    /**
+     * Full text of previous function call that produced "StoredStatus"
+     */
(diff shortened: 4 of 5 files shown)


skliper (commented):
Do the globals get cleared on a unit test reset?  I'd think they should be (or at least clear StoredText so it won't warn)... I'm thinking the case where something isn't implemented and we just want to skip the rest of the test, but the next store shouldn't error since it was intended to skip.

zanzaben:
I don't think we need to give a warning at all about overriding the store.

jphickey:
Just had a new idea here, which may make this cleaner.  The idea is to change "WAS" to be a soft fail - that is, it asserts with the NA case type, so you can do more than one test.  Then we can also have a "MUSTBE" for a hard fail.  This may make testcase code more readable.

jphickey:
Fairly trivial change to the macro/code makes the use case look like this:

```
        CFE_UtAssert_STATUS_STORE(CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName));
        /*
         * Normally, this will return CFE_SUCCESS, until the max number of pipes is reached.
         * Confirm that the last creation attempt returned CFE_SB_MAX_PIPES_MET
         *
         * NOTE: this also mimics the same format as UtAssert_INT32_EQ so that any post-procesing
         * test log analysis tools will see this call as well.
         */
        if (!CFE_UtAssert_STATUS_WAS(CFE_SUCCESS))
        {
            CFE_UtAssert_STATUS_MUSTBE(CFE_SB_MAX_PIPES_MET);
            break;
        }
```

And the output log looks like this:
```
[ PASS] 42.057 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (0) == CFE_SUCCESS (0)
[ PASS] 42.058 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (0) == CFE_SUCCESS (0)
[  N/A] 42.059 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (-905969660) == CFE_SUCCESS (0)
[ PASS] 42.060 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (-905969660) == CFE_SB_MAX_PIPES_MET (-905969660)
[ PASS] 42.061 sb_pipe_mang_test.c:100 - NumPipes (58) < CFE_PLATFORM_SB_MAX_PIPES (64)
```

zanzaben:
I think that makes this more complicated than it really needs to be. Also having an N/A without any kind of explanation why that test is no longer applicable would just be confusing.

zanzaben:
If you want to do more than one test on it why not just not clear out the store after a WAS is called. I don't think it's a problem if Store just overrides everytime.

skliper:
Not convinced breaking the assert pattern for STATUS_WAS is worth the extra complexity (introducing the concept of soft fail).  The test can store the status on the stack and do whatever logic you want.

jphickey:
I dunno, focusing on the test case implementation, IMO this makes it read like a book.  No extra local variables, no checking for things that are already checked for.

```
CFE_UtAssert_STATUS_STORE(CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName));
if (!CFE_UtAssert_STATUS_WAS(CFE_SUCCESS))
{
    CFE_UtAssert_STATUS_MUSTBE(CFE_SB_MAX_PIPES_MET);
    break;
}
```

The whole point of all this is we have plenty of cases where testing cannot boil down to not "one right value" in our test code.  That's just reality... In that sense, the UtAssert "pattern" is already broken (because they are written with the notion of one right value).  

So yes, this is a "new" pattern - sort of.  Because we've done one-off workarounds in a number of places already (and as such, not new).  It's just a matter of how it was solved in the past and how we want to solve it going forward.  Do we want lots of different ways of doing this type of thing, or do we want consistent patterns in the testing?  That's really the choice here.

Whether we go with this suggestion or something different, we do need to establish a some sort of recommended pattern or at least a modicum of consistency when dealing with cases where the output of a function is permitted to be one of several different values.  It isn't that rare in functional test.  (at least not rare enough, IMO, to say "lets not bother with that" - in coverage test, sure - there should be only one right value - but in functional test with uncontrolled preconditions - not so much).

jphickey:
My concern with the original form is that it's a little goofy to use in the test case code, typically asserting on something you just checked, in order to get the logs complete.  Take the CDS use case - with the original proposal, it would go something like this:

```
    status = CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name);

    if (status == CFE_ES_CDS_ALREADY_EXISTS)
    {
        CFE_UtAssert_STATUS_WAS(CFE_ES_CDS_ALREADY_EXISTS);
        UtAssert_NA("CDS already exists. CFE_ES_RegisterCDS new allocation could not be properly tested");
    }
    else
    {
        CFE_UtAssert_STATUS_WAS(CFE_SUCCESS);
    }
```

The fact that we checked for `CFE_ES_CDS_ALREADY_EXISTS` and then asserted for the same is a little ... odd?  It works, though (the goal being to record in the log file that we called CFE_ES_RegisterCDS with those parameters and got CFE_ES_CDS_ALREADY_EXISTS back).

Again, the whole impetus of this, is to get some improvement and consistency in what goes into the test log file.  With test cases handling each exception like this in a one-off manner, there is no consistency in either implementation or what gets logged (making tracing more difficult - we have to go back to the source code to see what "status == CFE_SUCCESS" meant here).  

The macros are intended to improve the accuracy of the log file - to more fully capture what was called and what was returned.  No earth-shattering new assert capabilities, just a better log file at the end of the day.

skliper:
I do like the STORE and WAS asserts as they relate to the logs (with WAS being a hard assert).  I don't see the added value of having a soft fail w/ an NA in the log, it seems more straight forward to me to keep it focused on actual asserts.  Or use a completely different prefix since it's not really an assert in the classical sense (at least not how I currently understand the proposed pattern) and I'd prefer it be very obvious and hard to get wrong... if you really want logic flow sort of messages in the log.  UtBranch/UtFlow/UtPath or similar, and could have it as another stage of reporting (or just make them debug) where it's optional if you want?

jphickey:
And in the form I'm proposing today, the same CDS use case would look something like: 

```
    CFE_UtAssert_STATUS_STORE(CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name));

    if (CFE_UtAssert_STATUS_WAS(CFE_ES_CDS_ALREADY_EXISTS))
    {
        UtAssert_NA("CDS already exists. CFE_ES_RegisterCDS new allocation could not be properly tested");
    }
    else
    {
        CFE_UtAssert_STATUS_MUSTBE(CFE_SUCCESS);
    }
```

No extra local status or local checks, no double-checking.  Reads clean, which I like - very clear what's going on here!

skliper:
CFE_UtFlow_STATUS_CHECK(blah) and keep CFE_UtAssert_STATUS_WAS as a hard assert?  STATUS_CHECK could just be debug info, or a step below the PASS/FAIL.

jphickey:
The value of "soft fail with NA" is that it's a casetype we already have, and requires absolutely no updates to the UtAssert library.  Yes, it is a stretch from the original intent of NA, and we could consider using a different name, but for the proof of concept, it fit what was needed to make a check that doesn't fail the test.

The extra "N/A" line in the log is a valid concern, but I don't think its wrong to have it in there:

`[  N/A] 42.059 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (-905969660) == CFE_SUCCESS (0)`

It's the truth, anyway - the test case DID check for it.  I concur its probably not much value to the end user, but to squelch it would require an update to UtAssert (I think) to differentiate from the cases you _do_ want to see - so that's where IMO it's better to leave it as is.  More info in logs is OK (ignore the extra), but too little is bad.

zanzaben:
I get it now, you are using CFE_UtAssert_STATUS_WAS as a getter that makes a record in the log. That idea is fine but I agree with skilper on the name change since this isn't a assert anymore. I also think that it should be an [info] event in the log instead of an [n/a].

skliper:
> My concern with the original form is that it's a little goofy to use in the test case code, typically asserting on something you just checked, in order to get the logs complete. Take the CDS use case - with the original proposal, it would go something like this:

Sorry, reading out of order.  So I think the example you had in this case is definitely goofy, but only because you are asserting something you weren't actually originally testing for.  Keep that specific test related to creating new, and the assert for already existing should be separate.  You have to do it if it is new, so why try to double up the purpose of the first test?

At minimum change the prefix... I don't care all that much about reporting, NA is fine but it needs to be clear it's not a PASS/FAIL assert.

skliper:
In other words, I see:
```
status = CFE_UtAssert_STATUS_STORE(CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name));

    if (status == CFE_ES_CDS_ALREADY_EXISTS)
    {
             UtAssert_NA("CDS already exists. CFE_ES_RegisterCDS new allocation could not be properly tested");
    }
    else
    {
        CFE_UtAssert_STATUS_WAS(CFE_SUCCESS);
    }
```
as being plenty good.  Sure it doesn't put the non-assert flow info in there, but it's really obvious in the log what it did...

jphickey:
> as being plenty good. Sure it doesn't put the non-assert flow info in there, but it's really obvious in the log what it did...

To a human reader, maybe ... They'll see the string "CDS already exists. CFE_ES_RegisterCDS new allocation could not be properly tested" ... one has to go back to the code and see that it was from a call to CFE_ES_RegisterCDS(), with three valid input arguments, returning CFE_ES_CDS_ALREADY_EXISTS.   My point is, that latter info should really be in the log too - like we log everything else that we call and returns a value, we should log the fact that we called `CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name)` and we got `CFE_ES_CDS_ALREADY_EXISTS` back, just like we log every other function call that way.  It's just as valid/important to know as the CFE_SUCCESS case was.

jphickey:
Also, I read the prefix a little differently.  Per coding standards/naming conventions, all macros/constants/functions defined by a module should be prefixed with the name of that module.  That's what the `CFE_UtAssert_` prefix is here - they are from the `cfe_assert` library (probably my mistake that "ut" was initally missed/added here, could always correct that too, we could change prefix across the board to "CFE_Assert" instead of "CFE_UtAssert" - that would better adhere to convention).  

The point is, that prefix is not to say that everything makes some sort of "assert" in the log - far from it, the library has plenty of helper functions that are also prefixed with _CFE_RegisterTest, CFE_Assert_LibInit, and on.  Just because it starts with "CFE_Assert" does not mean it triggers an assert in the log.

jphickey:
Latest update does the following:
- Use `CFE_Assert_` prefix on new macros instead of `CFE_UtAssert_`
- Use names `CFE_Assert_STATUS_MAY_BE` and `CFE_Assert_STATUS_MUST_BE` which hopefully makes the intent clearer.  The former checks for a passable/acceptable result but does not fail if it does not match.  The latter is more of a conventional assert, but would be the _last_ thing used - the test implementer can use MAY_BE first to skip over the MUST_BE check.

jphickey:
The base branch does not yet include a test for create pipe max number, but that test now looks like this:

```
        CFE_Assert_STATUS_STORE(CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName));
        if (CFE_Assert_STATUS_MAY_BE(CFE_SB_MAX_PIPES_MET))
        {
            break;
        }

        CFE_Assert_STATUS_MUST_BE(CFE_SUCCESS);
```

zanzaben:
Heads up, this will have conflicts with https://github.com/nasa/cFE/pull/1785

jphickey:
Note, see also nasa/osal#1131 - adding a couple new case types to OSAL allows us to better capture the nuanced differences between these N/A cases.  With that, we relegate these new "MAY_BE" messages to the debug level if they don't match, which cleans up the log nicely.  In my own test branch (including nasa/osal#1131) the logs now look like this.

For CDS normal startup (PR):
```
[ INFO] es_cds_test.c:45:Testing: CFE_ES_RegisterCDS
[ PASS] 02.002 es_cds_test.c:55 - CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name) (0) == CFE_SUCCESS (0)
[ PASS] 02.003 es_cds_test.c:75 - CFE_ES_RegisterCDS(&CDSHandlePtr2, BlockSize, Name) (1140850701) == CFE_ES_CDS_ALREADY_EXISTS (1140850701)
[ PASS] 02.004 es_cds_test.c:78 - CFE_ES_RegisterCDS(NULL, BlockSize, Name) (-1006632950) == CFE_ES_BAD_ARGUMENT (-1006632950)
```

For CDS already exists startup (PR):
```
[ INFO] es_cds_test.c:45:Testing: CFE_ES_RegisterCDS
[ PASS] 02.001 es_cds_test.c:55 - CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name) (1140850701) == CFE_ES_CDS_ALREADY_EXISTS (1140850701)
[ SKIP] 02.002 es_cds_test.c:64 - CDS already exists. CFE_ES_RegisterCDS new allocation could not be properly tested
[ PASS] 02.003 es_cds_test.c:78 - CFE_ES_RegisterCDS(NULL, BlockSize, Name) (-1006632950) == CFE_ES_BAD_ARGUMENT (-1006632950)
```

And finally, the CreatePipe max test (all other "max" tests would be similar):
```
[ PASS] 42.116 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (0) == CFE_SUCCESS (0)
[ PASS] 42.117 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (-905969660) == CFE_SB_MAX_PIPES_MET (-905969660)
[ PASS] 42.118 sb_pipe_mang_test.c:95 - NumPipes (58) < CFE_PLATFORM_SB_MAX_PIPES (64)
[ PASS] 42.119 sb_pipe_mang_test.c:101 - CFE_SB_DeletePipe(PipeId[NumPipes]) (0) == CFE_SUCCESS (0)
```

zanzaben (approved):

astrogeco:
[CCB:2021-08-18](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-August-18-2021) APPROVED
- Once nasa/osal#1131 is merged we can replace NA with the new tags

jphickey:
Latest proposal just uses a case type of "FLOW" (see latest updates in nasa/osal#1131)

Here is what a "Nominal" CDS test run looks like in the log (PO reset, CDS does not exist, default log level):

```
[ INFO] es_cds_test.c:46:Testing: CFE_ES_RegisterCDS
[ PASS] 04.002 es_cds_test.c:56 - CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize2, Name) (0) == CFE_SUCCESS (0) -> match
[ PASS] 04.003 es_cds_test.c:76 - CFE_ES_RegisterCDS(&CDSHandlePtr2, BlockSize2, Name) (1140850701) == CFE_ES_CDS_ALREADY_EXISTS (1140850701)
[ PASS] 04.004 es_cds_test.c:79 - CFE_ES_RegisterCDS(&CDSHandlePtr2, BlockSize, Name) (0) == CFE_SUCCESS (0)
```

Here is what a "Verbose" CDS test run looks like:

```
[ INFO] es_cds_test.c:46:Testing: CFE_ES_RegisterCDS
[ FLOW] 04.001 es_cds_test.c:56 - CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize2, Name) (0) == CFE_ES_CDS_ALREADY_EXISTS (1140850701) -> not match
[ PASS] 04.002 es_cds_test.c:56 - CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize2, Name) (0) == CFE_SUCCESS (0) -> match
[ PASS] 04.003 es_cds_test.c:76 - CFE_ES_RegisterCDS(&CDSHandlePtr2, BlockSize2, Name) (1140850701) == CFE_ES_CDS_ALREADY_EXISTS (1140850701)
```

And for another example, here is what the "Max Gen Counters" test case looks like in default log level:

```
[ INFO] es_counter_test.c:42:Testing: CFE_ES_RegisterGenCounter
[ PASS] 07.001 es_counter_test.c:47 - CFE_ES_RegisterGenCounter(&Ids[0], NULL) (-1006632950) == CFE_ES_BAD_ARGUMENT (-1006632950)
[ PASS] 07.002 es_counter_test.c:48 - CFE_ES_RegisterGenCounter(NULL, CounterName) (-1006632950) == CFE_ES_BAD_ARGUMENT (-1006632950)
[ PASS] 07.004 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.006 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.008 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.010 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.012 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.014 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.016 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.017 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (-1006632917) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> match
[ PASS] 07.018 es_counter_test.c:70 - NumCounters (7) <= CFE_PLATFORM_ES_MAX_GEN_COUNTERS (8)
```

And here is what the same run looks like with "verbose" log level:

```
[ INFO] es_counter_test.c:42:Testing: CFE_ES_RegisterGenCounter
[ PASS] 07.001 es_counter_test.c:47 - CFE_ES_RegisterGenCounter(&Ids[0], NULL) (-1006632950) == CFE_ES_BAD_ARGUMENT (-1006632950)
[ PASS] 07.002 es_counter_test.c:48 - CFE_ES_RegisterGenCounter(NULL, CounterName) (-1006632950) == CFE_ES_BAD_ARGUMENT (-1006632950)
[ FLOW] 07.003 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> not match
[ PASS] 07.004 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ FLOW] 07.005 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> not match
[ PASS] 07.006 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ FLOW] 07.007 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> not match
[ PASS] 07.008 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ FLOW] 07.009 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> not match
[ PASS] 07.010 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ FLOW] 07.011 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> not match
[ PASS] 07.012 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ FLOW] 07.013 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> not match
[ PASS] 07.014 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ FLOW] 07.015 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> not match
[ PASS] 07.016 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (0) == CFE_SUCCESS (0) -> match
[ PASS] 07.017 es_counter_test.c:57 - CFE_ES_RegisterGenCounter(&Ids[NumCounters], CounterName) (-1006632917) == CFE_ES_NO_RESOURCE_IDS_AVAILABLE (-1006632917) -> match
[ PASS] 07.018 es_counter_test.c:70 - NumCounters (7) <= CFE_PLATFORM_ES_MAX_GEN_COUNTERS (8)
```

jphickey:
Note that the "special sauce" of including an extra match/not match tag at the end is done inside the implementation of `CFE_Assert_STATUS_MAY_BE` -- since this is where we _know_ that the condition is actually describing a boolean expression.  At the UtAssert level (i.e. `UtAssert_Message`), its just a generic informational message.

jphickey:
IMO, its kinda redundant, because the raw info (numbers) are already there, you can already see its not a match.  But if this helps, I'm OK with it.

zanzaben:
Is CFE_Assert_STATUS_MUST_BE also adding that match/not match text. I don't think it should be since it is a hard pass or fail test.

skliper:
Any method you want to report in the log what the test thinks the result is works for me, I think the implementation in the macro is ideal.  You could even change it to just show `!=` within the `FLOW` case type and drop the match/not match.

skliper:
To me then it would basically read as for the MAY_BE test it was either PASS based on the result in the text or was FLOW based on the result shown in the text.  Makes it really clear it's a non-failure and just a decision point.  But really I'm fine with it as-is also since it meets my desired criteria... just brainstorming.

skliper:
> IMO, its kinda redundant, because the raw info (numbers) are already there, you can already see its not a match. But if this helps, I'm OK with it.

This really does help me.  I very much struggle with comparing -1021399484 with -1021399484 with -1021394984 quickly and benefit greatly from knowing what the code thought the result was.  Somewhat trivial here (although I honestly do still struggle w/ this simple example and have to study it for quite a while before I'm really sure), but string comparisons if we eventually support those in a soft test are much harder to spot quickly.

jphickey:
OK - the latest update squashes all the activity, along with the following cleanup:

 - use "==" and "!=" rather than an extra match / no match tag (should be clear enough)
 - Use the "UtAssert_GetCaseTypeAbbrev" function rather than duplicating the case type switch

Should be good to go, however this depends on nasa/osal#1131 (this is why the CI is failing build, as soon as OSAL is merged should be able to re-run and pass).

skliper (approved):

modules/cfe_assert/src/cfe_assert_runner.c
```diff
@@ -89,6 +89,66 @@ bool CFE_UtAssert_StatusCheck(CFE_Status_t Status, bool ExpectSuccess, UtAssert_
...
+    else
+    {
+        Result = (Status == CFE_Assert_Global.StoredStatus);
+        if (Result)
+        {
+            EqualityTag = '=';
+        }
+        else
+        {
+            EqualityTag = '!';
```
zanzaben, line 139: Doesn't this print out =/! instead of ==/!=

jphickey:
OK, the only downside/weirdness that I can see is that when a case of `CFE_Assert_STATUS_MUST_BE` fails, it goes through the same helper function, and also inverts the condition in the string (match/no match did not suffer from this issue).

Example, I hacked the CreatePipe max test to force a failure - basically to check `MUST_BE(CFE_STATUS_EXTERNAL_RESOURCE_FAIL)` instead of `MUST_BE(CFE_SUCCESS)` and this is what the failure line looks like:

```
[ INFO] sb_pipe_mang_test.c:62:Testing: CFE_SB_CreatePipe, maximum pipe limit
[ FLOW] 52.001 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (0) != CFE_SB_MAX_PIPES_MET (-905969660)
[ FAIL] 52.002 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (0) != CFE_STATUS_EXTERNAL_RESOURCE_FAIL (-939524091)
```

jphickey:
To clarify - match/no match was still added/printed in the MUST_BE case, it just made a little more sense than switching to `!=` in this context

modules/cfe_assert/src/cfe_assert_runner.c
jphickey: no, see actual log below.

jphickey:
Updated again to change the `!=` back to an extra `[false]` tag at the end of the line.  The trace now looks like this:

```
[ INFO] sb_pipe_mang_test.c:62:Testing: CFE_SB_CreatePipe, maximum pipe limit
[ FLOW] 52.001 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (0) == CFE_SB_MAX_PIPES_MET (-905969660) [false]
[ FAIL] 52.002 sb_pipe_mang_test.c:75 - CFE_SB_CreatePipe(&PipeId[NumPipes], 10, PipeName) (0) == CFE_STATUS_EXTERNAL_RESOURCE_FAIL (-939524091) [false]
```

I like this better because we are being accurate about the actual condition that was checked for in all cases, but it still should be clear if that had evaluated as false.  Note that I did _not_ add a corresponding `[true]` tag, since most statements should be true, this keeps the [PASS] messages consistent with other pass messages.

skliper (approved):
Looks good to me, I would like to consider the additional [false] marker for the other macros under our control for consistency (fine as a separate/future issue).

jphickey:
Note - the build failure currently reported from CI is fixed by nasa/osal#1131 - currently in IC - so if this gets merged (or if we just finish the OSAL merge) this should work OK.  

I'd like to get this merged in the build before the new RC tag is created, if possible.

PR #1849, opened 2021-08-19
Fix #1846, Disable Unit Tests in CodeQL

### LINKED ISSUE #1846: Disable Unit Tests in CodeQL
Do not analyze unit and functional test code for CodeQL. 

Disable unit tests in the workflow. 

**Requester Info**
Ariel Adams, ASRC Federal

ArielSAdamsNASA:
**Describe the contribution**
Fix #1846 

**Testing performed**
Tested on forked repo: https://github.com/ArielSAdamsNASA/cFE-1/security/code-scanning?query=branch%3Afix-1846-disable-unit-tests-codeql

**Expected behavior changes**
Disabled unit tests for both CodeQL workflows. Alerts should only be shown for flight code. 

Ariel Adams, ASRC Federal

### DIFF
--- a/.github/workflows/codeql-build.yml
+++ b/.github/workflows/codeql-build.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   SIMULATION: native
-  ENABLE_UNIT_TESTS: true
+  ENABLE_UNIT_TESTS: false
   OMIT_DEPRECATED: true
   BUILDTYPE: release
 


skliper:
To capture some of the conversation yesterday - currently the unit tests need to be built to get any of the warnings relative to the other operating systems (specifically VxWorks) since without the unit tests just the single implementation is built (for whatever OS is selected).  Also building the unit tests/coverage allows for getting reports on all source selection options at the same time, since the real build only builds the configured selection but the coverage test builds tests for all selections.

Discussed the possibility of categorizing test code by file name to tell CodeQL it is test code (I think there's a way to indicate with a wildcard).  Then we can still analyze all possible code but reduce or ignore the warnings from the test code.

skliper (changes requested):
Keep unit tests enabled but inform CodeQL as to what is test code.

ArielSAdamsNASA:
@skliper I thought the unit test should only be enabled for osal. Is it for cfe and osal or just cfe?

skliper:
They both have the same concept of testing all "source-select" configurations via unit testing, but only the selected configuration is built when the unit tests aren't built.  So, both should leave the unit-test build in for CodeQL analysis.

astrogeco:
What if we somehow do separate builds per platform so that the alerts themselves get categorized by platform? Not sure if this makes sense, but what I'm visualizing is that the "alerts" would have a tag that says which platform it applies to...

skliper:
Is it possible to shift the focus from adding complexity to the sorting of warnings using this reporting mechanism to just resolving the warnings (and/or quarantining rules we don't chose to comply with)?  We've got at least 3 different ways to generate coding standards warnings now, one of which is really easy to sort and analyze... so I see the purpose of this report to just to show the open source users how we are doing relative to the chosen standards (and maybe catch a few different volations, but not as the primary analysis tool).   It's not the easiest of ways to actually do the analysis (other tools are much better in my experience)... so is it really worth the extra effort here?

astrogeco:
That's a good point, especially since open source users can't even see the code scanning alerts dashboard.

skliper:
> That's a good point, especially since open source users can't even see the github code scanning alerts dashboard. The initial goal here was to categorize the alerts to get a better handle on the 5000+ ones that are not displayed and that are probably mostly in test code

For that case I'd use the tools we have available that have a fairly powerful sorting/filtering capability :)  There may be a handful of different warnings based on the way the tools implement the checks, but that can be resolved once we get the bulk set dispositioned.

astrogeco:
Closing, we decided to keep scans of unit test code as examples of good practices

PR #1913, opened 2021-08-30
Fix #1912, Update time tests to use bitmask check macros

### LINKED ISSUE #1912: Update TIME tests to use bitmask check macros
With nasa/osal#1136, UtAssert now provides a better macro for testing bit fields/masks.  The updated macro provides better/more concise information in the test log.

Replace the UtAssert_UINT32_EQ checks with the new macros - UtAssert_BITMASK_SET/UtAssert_BITMASK_UNSET.

**Additional context**
Depends on nasa/osal#1136 and nasa/cfe#1860 (both currently in IC but not in main, as of this writing).

**Requester Info**

jphickey:
**Describe the contribution**
The new bitmask check macros provide more concise information the resulting log file.  This updates the time test to use those macros when checking the state flags.

**Testing performed**
Build and run all tests, confirm log output is correct

**Expected behavior changes**
Improved test log file info

**System(s) tested on**
Ubuntu

**Additional context**
Note this also changes the code to call `CFE_TIME_GetClockInfo()` repeatedly, rather than storing in a local variable and checking that value.  The intent is to clearly indicate in the log that this function was invoked, not just what its output was.  This helps with tracking/confirming that the function is covered by the functional test.

This just reads a global so it should not be different in subsequent calls, unless the clock actually changes state during the test.  But that was already a risk with the separate call to `CFE_TIME_GetClockState`, this does not really change that - and the flags are not really connected otherwise.

### DIFF
--- a/modules/cfe_testcase/src/time_current_test.c
+++ b/modules/cfe_testcase/src/time_current_test.c
@@ -125,31 +125,30 @@ void TestClock(void)
 {
     UtPrintf("Testing: CFE_TIME_GetClockState, CFE_TIME_GetClockInfo");
 
-    CFE_TIME_ClockState_Enum_t state     = CFE_TIME_GetClockState();
-    uint16                     ClockInfo = CFE_TIME_GetClockInfo();
+    CFE_TIME_ClockState_Enum_t state = CFE_TIME_GetClockState();
 
     if (state >= 0)
     {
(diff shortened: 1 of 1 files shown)


zanzaben (approved):

zanzaben:
Dependent on https://github.com/nasa/osal/pull/1136

astrogeco:
[CCB:2021-09-01](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-01-2021) APPROVED

PR #1943, opened 2021-09-09
Fix #1942, add missing inclusions in CFE API headers

### LINKED ISSUE #1942: Missing some dependency include files in public API headers
**Describe the bug**
As a general rule of thumb, all header files should directly include whatever dependencies they require in order to provide the types/declarations they intend to provide.

However in the CFE headers there remain a couple omissions/mistakes in this regard:

- `cfe_es.h` declares the function `CFE_ES_GetModuleInfo` which accepts a resource ID input, so this depends on `cfe_resourceid_api_typedefs.h`, but it does not directly include this dependency
- `cfe_tbl_api_typedefs.h` defines a `CFE_TBL_Info_t` type, which in turn has a member sized to `CFE_MISSION_MAX_PATH_LEN`, which is provided by `cfe_mission_cfg.h`, but it does not directly include this dependency.

In both cases the current framework sample builds do compile successfully, because the dependent header gets included implicitly (i.e. by some header before it) in all the current use cases, but this could change in other use cases.

**To Reproduce**
Use header files in contexts beyond what the current framework does, such as 3rd party code only including "cfe_es.h" or "cfe_tbl_api_typedefs.h" directly.

**Expected behavior**
Headers should work (compile w/o errors or warnings) when included individually, they should include all dependencies directly rather than relying on inclusion ordering.

**System observed on:**
Ubuntu

**Reporter Info**

jphickey:
**Describe the contribution**
Some CFE API headers were missing dependency inclusions, where the header was referencing a type or symbol but not directly including the header file that provides that type/symbol.

Adding the dependent include allows the headers to work more consistently.

Fixes #1942

**Testing performed**
Build and sanity check CFE

**Expected behavior changes**
None for framework config (these inclusions were already satisfied in existing use cases)
But fixes a broken build for external code that only included these headers directly.

**System(s) tested on**
Ubuntu

**Additional context**
This type of thing was intended to be caught by the "headercheck" concept in the module code, but that is currently not enabled/enforced in the build.

There theoretically could be other instances of stuff like this, but these are the only two noted so far.

### DIFF
--- a/modules/core_api/fsw/inc/cfe_es.h
+++ b/modules/core_api/fsw/inc/cfe_es.h
@@ -41,6 +41,7 @@
 #include "common_types.h"
 #include "cfe_error.h"
 #include "cfe_es_api_typedefs.h"
+#include "cfe_resourceid_api_typedefs.h"
 
 /*
 ** The OS_PRINTF macro may be defined by OSAL to enable
--- a/modules/core_api/fsw/inc/cfe_tbl_api_typedefs.h
+++ b/modules/core_api/fsw/inc/cfe_tbl_api_typedefs.h
@@ -40,6 +40,7 @@
 
 /********************* Include Files  ************************/
 #include "common_types.h" /* Basic Data Types */
+#include "cfe_mission_cfg.h"
 #include "cfe_tbl_extern_typedefs.h"
 #include "cfe_time_extern_typedefs.h"
 


jphickey:
This is the type of stuff I'd suggest to queue up for a "caelum.1" release ... given that caelum is already tagged or soon to be.  probably will continue to slowly find other gotchas as the code is used for more stuff by more users.

astrogeco:
Is there a way to enforce the "headercheck" concept in CI?

zanzaben (approved):

skliper (approved):

astrogeco:
[CCB:2021-09-15](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-15-2021) APPROVED
- Explore using cfe_app_check_intf from modules/CMakeLists.txt in continuous-integration checks

astrogeco:
@skliper can we move this to Caelum?

skliper:
@astrogeco Absolutely!  Moved to Caelum.

PR #1950, opened 2021-09-14
Fix #1949, update msgid testcase to match implementation

### LINKED ISSUE #1949: Mismatch between MSG API and test case in "TestMsgId" functional test
**Describe the bug**
There is a test case in the "TestMsgId" set which passes in `CFE_SB_INVALID_MSG_ID` to `CFE_MSG_SetMsgId()`, and expects `CFE_MSG_BAD_ARGUMENT` return value:

https://github.com/nasa/cFE/blob/64a6a59456fa9e47dc93e4bb9cecacc3d86d1862/modules/cfe_testcase/src/message_id_test.c#L48

However:
- The API does not document that it returns CFE_MSG_BAD_ARGUMENT in response to an invalid MsgId value (in fact it does not say anything about validating the input MsgId at all)
- The implementation is not actually checking if its a valid MsgId anyway.  It is checking if it is `> CFE_PLATFORM_SB_HIGHEST_VALID_MSGID`, which is a different concept.

Although this is currently "passing" - it is only by chance, because CFE_SB_INVALID_MSG_ID has the value of -1, which when converted to an unsigned int, will be greater than `CFE_PLATFORM_SB_HIGHEST_VALID_MSGID` (unless the latter is set to 0xFFFFFFFF). 

**To Reproduce**
Run this test against an alternate MSG module implementation (i.e. one that has different criteria) and/or change the SB definition of "CFE_SB_INVALID_MSG_ID".   The test will now fail.

**Expected behavior**
Test case should still pass, even when run against an alternate MSG implementation.  Should not depend on "chance" values that it does not control.

**Code snips**
Actual implementation is here (same basic check in v1/v2):
https://github.com/nasa/cFE/blob/64a6a59456fa9e47dc93e4bb9cecacc3d86d1862/modules/msg/fsw/src/cfe_msg_msgid_v1.c#L67

**System observed on:**
Ubuntu

**Additional context**
The important concept is nowhere does the documentation say that the `CFE_SB_INVALID_MSG_ID` constant must be greater than the `CFE_PLATFORM_SB_HIGHEST_VALID_MSGID`.  In fact, the latter may not even exist in all implementations.

- If the intent was to reject an invalid MsgId value, the proper function to use is `CFE_SB_IsValidMsgId`, and the API documentation should state that CFE_MSG_BAD_ARGUMENT will be returned in response to an invalid MsgId (it does not currently say this).
- However, in general the MSG module is just supposed to be a getter/setter, not a validator, in its role.  So in that sense, validating the MsgId is superfluous here, and the check against "highest" MsgId should be removed.

**Reporter Info**

jphickey:
**Describe the contribution**
Updates the test case for "CFE_MSG_SetMsgId()" to write a value that is outside the set of storable values for MsgId.  This currently
uses the value "-1" do this, assuming that this translates to a value with all bits set, and at least one of those bits will not correlate with an actual header field.

Caveat is that in other implementations, it is possible that all bits are storeable in the header.  That is, all MsgId values can be stored in a message or vice versa, even those MsgIds which are not route-able in SB.  

This also updates the documentation of CFE_MSG_SetMsgId to clarify if/when the CFE_MSG_BAD_ARGUMENT status code is returned, and that the set of MsgId values which are "settable" really depends on the MSG implementation.

Fixes #1949 

**Testing performed**

**Expected behavior changes**
None with default MSG implementation and current config.

**System(s) tested on**
Ubuntu

**Additional context**
Note that the "CFE_SB_IsValidMsgId" and the related constants `CFE_SB_INVALID_MSG_ID` and `CFE_PLATFORM_SB_HIGHEST_VALID_MSGID` are really pertinent message routing in SB and SBR, more so than MSG.  

The general concept of what constitutes a "valid" msgid is different for MSG than SB/SBR.  For MSG it is just a matter of how many bits the real header has, and how this bits are mapped between the header and CFE_SB_MsgId_t type.  For example there are values which are "storable" in the CCSDS primary header, but are not valid for SB/SBR or usable with CFE.

In particular, it is possible that an application might be serving as an intermediate node in a DTN network - in which case it might need to decode and/or encode messages which are not actually routed on the CFE software bus at all.  It should be SB (and SBR) domain to determine if a MsgId is route-able.  MSG only should care if it is store-able in the header.

### DIFF
--- a/modules/cfe_testcase/src/message_id_test.c
+++ b/modules/cfe_testcase/src/message_id_test.c
@@ -45,7 +45,20 @@ void TestMsgId(void)
     UtAssert_UINT32_EQ(msgid, expectedmsgid);
 
     UtAssert_INT32_EQ(CFE_MSG_SetMsgId(NULL, msgid), CFE_MSG_BAD_ARGUMENT);
-    UtAssert_INT32_EQ(CFE_MSG_SetMsgId(&msg, CFE_SB_INVALID_MSG_ID), CFE_MSG_BAD_ARGUMENT);
+
+    /*
+     * The purpose of this test is to attempt to set a MsgId beyond the set of values that can
+     * be stored in the MSG header. However the criteria for being "storable" depends on the
+     * actual MSG implementation, and may be different in other implementations.  Currently both
+     * "v1" and "v2" implementations do define a specific highest MsgId value, but another
+     * implementation might not have a highest number concept at all.
+     *
+     * By passing the value of -1, when converted to a an unsigned value (either 16 or 32 bit)
+     * it should translate to a MsgId value with all bits being set.  In theory, at least some of
+     * those bits will be not mappable to the packet header bits, and it should therefore elicit
+     * the CFE_MSG_BAD_ARGUMENT response.
+     */
+    UtAssert_INT32_EQ(CFE_MSG_SetMsgId(&msg, CFE_SB_ValueToMsgId(-1)), CFE_MSG_BAD_ARGUMENT);
 
     UtAssert_INT32_EQ(CFE_MSG_GetMsgId(NULL, &msgid), CFE_MSG_BAD_ARGUMENT);
     UtAssert_INT32_EQ(CFE_MSG_GetMsgId(&msg, NULL), CFE_MSG_BAD_ARGUMENT);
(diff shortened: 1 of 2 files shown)


astrogeco:
[CCB:2021-09-15](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-15-2021) APPROVED

PR #1954, opened 2021-09-16
Fix #1952, patch for recursive event loop

### LINKED ISSUE #1952: Potential for recursive loop if Event TLM MsgId is incorrect
**Describe the bug**
If the software bus `CFE_SB_TransmitMsg()` fails to send a message due to a validation failure, it will send an event through EVS.

Event Services, in turn, generates a message (LongEventTlm/ShortEventTlm) which is broadcast via software bus.

However if the Event Telemetry MID value (`CFE_EVS_LONG_EVENT_MSG_MID`) is not set correctly (or some other EVS config is bad) such that EVS tries to send event messages which do not validate, a recursive loop ensues and the software eventually segfaults.

**To Reproduce**
(Mis)configure `CFE_EVS_LONG_EVENT_MSG_MID` to a value which will not pass the CFE_SB_TransmitMsgValidate tests.  Run CFE, it will get in a recursive loop and eventually segfault/crash as soon as any app sends an event.

**Expected behavior**
Should not do a recursive loop

**System observed on:**
Ubuntu

**Additional context**
Should be protection against recursive event loops like this, where if an event fails to send, it should not cause another event to be sent.   This protection appears this is not working correctly right now, at least not for EVS messages.

__NOTE__: To be absolutely clear - the issue described here is a mis-configuration issue.  It will not happen in a properly configured system, so long as EVS generates messages which are "transmittable".  BUT - there are other events that might be triggered by a CFE_SB_TransmitMsg call, such as a MsgLim error, and its not clear of a similar recursive loop might be possible there (have not tested/investigated).

**Reporter Info**

jphickey:
**Describe the contribution**
Adds CFE_SB_RequestToSendEvent/CFE_SB_FinishSendEvent wrappers around all events generated by CFE_SB_TransmitMsgValidate.  This is avoids the potential for a recursive loop if configured improperly.

Fixes #1952

**Testing performed**

**Expected behavior changes**
Tested with the mis-configuration where EVS had an invalid MID and confirmed that the system does not get into a recursive loop and segfault anymore.

**System(s) tested on**
Ubuntu

**Additional context**
This is just a minimal patch that does not change current architecture but avoids a crash.  A better fix for future versions would be to put this recursion protection into EVS in a more generic manner.

### DIFF
--- a/modules/sb/fsw/src/cfe_sb_api.c
+++ b/modules/sb/fsw/src/cfe_sb_api.c
@@ -1503,24 +1503,42 @@ int32 CFE_SB_TransmitMsgValidate(const CFE_MSG_Message_t *MsgPtr, CFE_SB_MsgId_t
         switch (PendingEventID)
         {
             case CFE_SB_SEND_BAD_ARG_EID:
-                CFE_EVS_SendEventWithAppID(CFE_SB_SEND_BAD_ARG_EID, CFE_EVS_EventType_ERROR, CFE_SB_Global.AppId,
-                                           "Send Err:Bad input argument,Arg 0x%lx,App %s", (unsigned long)MsgPtr,
-                                           CFE_SB_GetAppTskName(TskId, FullName));
+                if (CFE_SB_RequestToSendEvent(TskId, CFE_SB_SEND_BAD_ARG_EID_BIT) == CFE_SB_GRANTED)
+                {
+                    CFE_EVS_SendEventWithAppID(CFE_SB_SEND_BAD_ARG_EID, CFE_EVS_EventType_ERROR, CFE_SB_Global.AppId,
+                                               "Send Err:Bad input argument,Arg 0x%lx,App %s", (unsigned long)MsgPtr,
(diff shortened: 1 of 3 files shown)


jphickey:
Pushed to fix clang-format errors.

astrogeco:
https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-September-29-2021 APPROVED
- caused by a misconfiguration
- would benefit from re-architecting

astrogeco:
@jphickey My merge of this one is causing the test failure
See https://github.com/astrogeco/cFE/commit/44312f0395152995317d55618cbd16989caa18fb
```
[  TTF] 127.027 sb_UT.c:4610 - CFE_SB_DeletePipe(PipeId) (0xffffffffca000003) == CFE_SUCCESS
[  END] 127 SB_SpecialCases.TransmitMsgPaths_Nomina TOTAL::27    PASS::26    FAIL::0     MIR::0     TSF::0     TTF::1     WARN::0   
```

PR #1966, opened 2021-09-20
Fix #1965, better message type pointer conversion

### LINKED ISSUE #1965: Header structure name assumptions limit MSG abstraction capability
The MSG library is supposed to offer a complete abstraction of the header format, but in practice there are still some direct references to header structure members in Caelum which limit this capability.

In particular, when converting a local message buffer to a `CFE_MSG_Message_t` pointer, when passing into any CFE_SB API, the code references a sub-member, such as:

https://github.com/nasa/cFE/blob/64a6a59456fa9e47dc93e4bb9cecacc3d86d1862/modules/es/fsw/src/cfe_es_task.c#L396

While this was nice in that it is fully type-safe, it limits the header abstraction in two important ways:
1. It assumes/requires that the buffer itself contains a `TlmHeader` member.
2. It assumes/requires that the `TlmHeader` structure, in turn, contains a `.Msg` member.  

For item (2) above, the TLM header is a structure controlled/defined by the CFE_MSG module, and there is no guarantee that a "Msg" member exists.  This will be a problem if an alternate MSG module is used, where the `CFE_MSG_Message_t` member is not called "Msg".  (The whole point of MSG is to make these structures free-form, and no assumptions should be made as to their internal structure, so this reference somewhat defeats that purpose).

- Convert this to a cast instead, e.g. `(CFE_MSG_Message_t*)&CFE_ES_Global.TaskData.HkPacket`
- Have CFE_MSG provide a macro/inline function to implement this conversion/cast

Just document what the names need to be, and make it a requirement to name things in this manner.

Major issues with that - mainly that it is not friendly to automated tools which might be used to generate these headers/structures from a CMD/TLM database.

**Additional context**
There is no real convention to the names that exist today.  It would be more viable to do that if a specific convention/reasoning was followed, such as the `CFE_MSG_Message_t` member being called `Message` (i.e. without the CFE_MSG_ prefix or _t suffix).  This way a tool would know what name to call things in the generated files.  But as it is, there is simply no naming consistency in these members, a tool would have to hard-code "special" member names for each structure, for no good reason other than that's what a human had used for an abbreviation at one point.

But furthermore, even if a name convention was follwed prevents an additional layer of headers to be added.  For example, the "TlmHeader" is assumed contain a "Msg" member directly.  However in some implementations a user might want an intermediate header, where it would become `.TlmHeader.Intermediate.Msg` instead.  There is no way to accommodate a third layer with the current assumptions in the code.

_Regarding suggested casts_ -- by casting, it can be converted without knowing what the member is specifically named, nor knowing how deep within the structure the Message structure lies.  This is not as bad as it sounds, and not really going back to being type-unsafe, because it is being cast to a `CFE_MSG_Message_t*`, not a `void*` as previous CFE versions had done here.  Because of this, and the fact that CFE is compiled with strict aliasing rules enabled, it will trigger an alias violation if the structure is not actually cast-able to a `CFE_MSG_Message_t` type.  While this is not quite as robust as the current type safety, it is much more flexible, and user errors/mismatches should still be caught.

**Requester Info**

jphickey:
**Describe the contribution**
Provides more abstract conversions between the local buffer pointers and the `CFE_MSG_Message_t*` pointer required for calling CFE_SB API, without assuming specific member names.

Fixes #1965

**Testing performed**

**Expected behavior changes**
None in framework build.
Improves source code compatibility when using an alternate MSG implementation that does not (directly) contain a `Msg` member.

**System(s) tested on**
Ubuntu

### DIFF
--- a/modules/cfe_testcase/src/msg_api_test.c
+++ b/modules/cfe_testcase/src/msg_api_test.c
@@ -54,75 +54,76 @@ void TestMsgApiBasic(void)
 
     /* test msg-init */
     UtAssert_INT32_EQ(CFE_MSG_Init(NULL, CFE_SB_ValueToMsgId(0), sizeof(cmd)), CFE_MSG_BAD_ARGUMENT);
-    UtAssert_INT32_EQ(CFE_MSG_Init(&cmd.Msg, msgId, 0), CFE_MSG_BAD_ARGUMENT);
-    UtAssert_INT32_EQ(CFE_MSG_Init(&cmd.Msg, CFE_PLATFORM_SB_HIGHEST_VALID_MSGID + 1, sizeof(cmd)),
-                      CFE_MSG_BAD_ARGUMENT);
+    UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgId, 0), CFE_MSG_BAD_ARGUMENT);
+    UtAssert_INT32_EQ(
+        CFE_MSG_Init(CFE_MSG_PTR(cmd), CFE_SB_ValueToMsgId(CFE_PLATFORM_SB_HIGHEST_VALID_MSGID + 1), sizeof(cmd)),
+        CFE_MSG_BAD_ARGUMENT);
 
-    UtAssert_INT32_EQ(CFE_MSG_Init(&cmd.Msg, msgId, sizeof(cmd)), CFE_SUCCESS);
+    UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgId, sizeof(cmd)), CFE_SUCCESS);
 
     /* test set-msg-size */
     UtAssert_INT32_EQ(CFE_MSG_SetSize(NULL, 12), CFE_MSG_BAD_ARGUMENT);
-    UtAssert_INT32_EQ(CFE_MSG_SetSize(&cmd.Msg, 0), CFE_MSG_BAD_ARGUMENT);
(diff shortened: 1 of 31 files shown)


skliper (approved):
I think I slightly prefer the cast... would need to propagate through all the framework and GSFC apps.  Approving for Draco.

jphickey:
I'm good with the cast too, the only downside is you may only catch real conversion errors when building with strict aliasing and optimization enabled.  For users doing active development, this is often not the case - typically code is built in debug mode, no optimizations, and then aliasing violations are typically not warned about.  So the macro approach attempts to rectify that, by effectively doing the same member lookup.

That being said, if we do go with a macro, I do have to move its definition to actually be part of MSG  (in this PR its still part of core_api).

jphickey:
After digging deeper and actually attempting to "force" an error via aliasing rules - reality is it just doesn't work that well (sensitive too compiler settings, compilers warn about things that aren't really violations, and don't warn about things that are).

In short - I am now leaning toward keeping the macro, as it can accommodate the EDS/generated header use-case and still provide type safety without relying on compiler features that aren't really that solid.

skliper:
Sounds good to me, that was my very initial concern when first discussed (how well it's really enforced across compilers, etc).  Thanks for doing the analysis, macro works for me.

jphickey:
This should be ready, whenever a "post-caelum" merge strategy is identified.  For now, I'll merge into a baseline in my fork (because this is necessary for EDS demo).

astrogeco:
[CCB:2021-11-03](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-November-03-2021) APPROVED
- What is the motivation of forcing a naming convention for the tool?

PR #1974, opened 2021-09-22
Fix #1945, add CFE_SB_ValueToMsgId/MsgIdToValue wrappers

### LINKED ISSUE #1945: Improve consistency in application of CFE_SB_MsgIdToValue/ValueToMsgId conversions
A `CFE_SB_MsgId_t` value, like other IDs, is supposed to be a unique type/opaque value that identifies a message within the SB application context.   Although it is currently implemented using an integer (`uint32` specifically) application should not assume this.  Instead, a set of macros and inline conversion functions (CFE_SB_MsgIdToValue and CFE_SB_ValueToMsgId) are provided for when the application needs to interpret the value as an integer for a valid purpose.

However, since use of this conversion is not currently enforced - there is no config option that makes MsgId into a strict type, as there is for ResourceID - so much of the newly-added code such as tests and apps, didn't completely apply it.

Add conversions where they are currently missing

Preferably, the MID constants should be defined directly as a `CFE_SB_MsgId_t` type value, which would eliminate the need for conversions all over the runtime code.  However, this breaks the switch/case paradigm used when dispatching - a `case` label needs to be an integer value.  So although this would yield cleaner code it does have a bigger backward compatibility impact which is why it was avoided for Caelum.

**Requester Info**

jphickey:
**Describe the contribution**
Correct code that was not correctly using the CFE_SB_ValueToMsgId or CFE_SB_MsgIdToValue conversion wrappers where required to do so. This should be used whenever the value is intentionally converted to/from an integer.  The CFE_SB_MsgId_t type should not be assumed to be an integer in nature.

Fixes #1945

**Testing performed**

**Expected behavior changes**
None for framework build / default config

**System(s) tested on**
Ubuntu

### DIFF
--- a/modules/cfe_testcase/src/message_id_test.c
+++ b/modules/cfe_testcase/src/message_id_test.c
@@ -42,7 +42,7 @@ void TestMsgId(void)
 
     UtAssert_INT32_EQ(CFE_MSG_SetMsgId(&msg, expectedmsgid), CFE_SUCCESS);
     UtAssert_INT32_EQ(CFE_MSG_GetMsgId(&msg, &msgid), CFE_SUCCESS);
-    UtAssert_UINT32_EQ(msgid, expectedmsgid);
+    CFE_Assert_MSGID_EQ(msgid, expectedmsgid);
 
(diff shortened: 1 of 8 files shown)


astrogeco:
@jphickey do you want to discuss this one at CCB today?

PR #1980, opened 2021-09-27
Fix #1979, implement abstract config registry module

### LINKED ISSUE #1979: Implement generic runtime configuration registry for CFE
Rather than compiling-in various values (e.g. via `cfe_platform_cfg.h` or `cfe_mission_cfg.h`) it would be more flexible and more useful to put many of these configuration items into some type of simple runtime registry.

The registry would be a simple key/value store, where the key is the configurable item ID, and the value is either an integer or a pointer to a global data object (ideally const, to avoid sharing/concurrent access issues).

The registry can be somewhat dynamic, allowing registry items to be easily added/extended.  Default values and text-based descriptions can also be potentially included.  

User API that works something like:
- `uint32 CFE_ES_RegistryGetInt(uint32 Key);`
- `const void* CFE_ES_RegistryGetPointer(uint32 Key);`

**Additional context**
This would reduce dependency on `cfe_platform_cfg.h` and `cfe_mission_cfg.h` for keeping this type of value, and also provide a place for the version info currently in `cfe_version.h`.   Thereby consolidating the current bunch of different types of info with different access methods into one generic registry with a common, simple access method.

Similar model exists in UNIX/POSIX in the `sysconf` facility - so certain platform-specific values do not need to be compiled into binaries, thereby improving binary portability (i.e. compile once for a CPU arch, run anywhere that has that arch).  This would make the long-desired feature of separate CFE core and app builds closer to reality.

Of course Windows has its well-known registry, which is huge and unwieldy - this should _not_ turn into that.  More like UNIX sysconf, but with pointers/strings also supported, not just integers.

Other issues with `#define`-ed config values:
- Value is effectively compiled-into the binary, so if the value ever changes, modular binaries become incompatible (but this is _not_ actively detected - its a subtle, silent incompatibility).
- Works OK for integers, but not as well for strings, even less well for structured data
- Requires that the `cfe_mission_cfg.h` / `cfe_platform_cfg.h` be kept in sync with the source code.  New variables added in new builds must be manually propagated to the user build.

**Requester Info**

jphickey:
**Describe the contribution**
Adds a new module called "config" that tracks simple key/value pairs.  All values are const.

As a proof of concept, all version information is mirrored into this key store, and ES is updated to use this instead
of the GLOBAL_CONFIGDATA when generating events.

Fixes #1979

**Testing performed**

**Expected behavior changes**
Slightly different version display during startup (module names are ALL_CAPS, mainly).  Otherwise pretty transparent.

**System(s) tested on**
Ubuntu

**Additional context**
This attempts to balance the advantages/disadvantages of having a string-based key/value store and an integer-based key/value store.  In this implementation, the table is primarily indexed by an integer, which is assigned at compile time, so its fast (direct lookup from key to table entry).  But the downside to this is that it creates a compile-time dependency on the assigned integer IDs.   So, the keys also have string names associated with them, such that an app can find a key by name, to avoid having a hard compile-time dependency on a particular configuration item, such as if its an optional component.

This also has the side effect of improving branch/line coverage in ES where it was reading the "const" configdata objects.  Due to the fact that configdata is declared "const", it means that coverage test can also only run with one copy.  Therefore, it could not cover all the branches as a result.


modules/config/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+##################################################################
+#
+# cFE Executive Services (ES) module CMake build recipe
```
pepepr08, line 3: "Config module CMake build recipe"

jphickey: Ahh yes, cut and paste evidence I forgot to fix.

Also worth noting that the coverage / functional test aspects are not at all implemented yet, waiting on getting some feedback before going that far....

pepepr08: Yeah I noticed it is still a draft so didn't really review the code

jphickey:
Pushed to fix up a few stray items, also corrects the cut/paste error in CMakeLists.txt

modules/config/CMakeLists.txt
jphickey: cut/paste error fixed in 145cecb5

astrogeco:
[CCB:2021-10-13](https://github.com/nasa/cFS/wiki/Community-CCB-Agenda:-October-13-2021) APPROVED

PR #2025, opened 2022-01-18
Fix #2024, osal_id_t type conversion in es_UT.c

### LINKED ISSUE #2024: Implicit/incorrect OSAL ID type conversion in es_UT.c
**Describe the bug**
The following line in unit tests makes an implicit conversion between `OS_OBJECT_ID_UNDEFINED` (an osal_id_t) and an integer:

https://github.com/nasa/cFE/blob/c161cf04fe7c5fd2f2a533ac1d7522b31191eab6/modules/es/ut-coverage/es_UT.c#L4522

This is allowed with the current permissive type definition, but fails when building with strict types.

**To Reproduce**
Use a type-safe definition for osal_id_t.  This conversion fails and results in a compiler error.

**Expected behavior**
Should use `OS_ObjectIdToInteger()` here - easiest way to make it work without changing logic.

**System observed on:**
Ubuntu 21.10

**Reporter Info**

jphickey:
**Describe the contribution**
Corrects the implicit type conversion from an osal_id_t to an integer, making it explicit by using the OS_ObjectIdToInteger() conversion function.

Fixes #2024

**Testing performed**
Build with strict osal_id_t type definition.  Run all tests.

**Expected behavior changes**
None.

**System(s) tested on**
Ubuntu 21.10

### DIFF
--- a/modules/es/ut-coverage/es_UT.c
+++ b/modules/es/ut-coverage/es_UT.c
@@ -4519,7 +4519,7 @@ void TestAPI(void)
 
     /* Hit error case for NULL TaskRecPtr */
     ES_ResetUnitTest();
-    UT_SetDeferredRetcode(UT_KEY(OS_TaskGetId), 1, OS_OBJECT_ID_UNDEFINED);
+    UT_SetDeferredRetcode(UT_KEY(OS_TaskGetId), 1, OS_ObjectIdToInteger(OS_OBJECT_ID_UNDEFINED));
     UtAssert_INT32_EQ(CFE_ES_GetTaskID(&TaskId), CFE_ES_ERR_RESOURCEID_NOT_VALID);
 }
 


skliper:
Do we have an established convention for ID macros, or at minimum recommended approach?  I seem to recall some of them are the correct type from within the macro (might have just been for unit tests?)... others are "raw" and always need to be converted.  Although it's likely a bigger change if they are already converted all over the place, might be friendlier (backwards compatabile) where macros are used to do the conversion inside?

skliper (approved):

jphickey:
Any initializer should always be of the correct type for the value it initializes.  This should be the default anyway.  It should only have a suffix/extension if this is _NOT_ the case for whatever reason, and that reason should be a good one if we have to have it at all.  (The mess w/MsgId vs MID Value comes to mind, we should be trying to get away from defining a MID as a number, as I've said for years, but in the meantime we can differentiate that a "MID VALUE" is different than a "MID").

However, in this case, the OSAL_OBJECT_ID_UNDEFINED has always been of the `osal_id_t` type since its inception.  It's just that because it wasn't protected against conversion, code like this can sneak by the checks.

astrogeco:
[CCB:2022-01-19](https://github.com/nasa/cFS/wiki/CCB:-2022.01.19) APPROVED

PR #2031, opened 2022-01-18
Fix #2030, array length calculation for perf structs

### LINKED ISSUE #2030: ES PERF array size calculation assumes "uint32" array base type
**Describe the bug**
The ES perf code uses macros to determine the array length based on the overall array size (via `sizeof()`).  Problem is, these calculations hard-code a `uint32` type as the divisor (element size) which may not always be true.

**To Reproduce**
Change the definition of the masks in TLM to be e.g. uint8-based instead of uint32-based.  Array lengths are now calculated incorrectly, because it divides by `sizeof(uint32)` which would now be wrong.

**Expected behavior**
Should use `[0]` to get the actual type of the element, this makes the calculation future-proof, as it does not assume/repeat the array element type (adheres to DRY principle)

**Code snips**
https://github.com/nasa/cFE/blob/c161cf04fe7c5fd2f2a533ac1d7522b31191eab6/modules/es/fsw/src/cfe_es_task.c#L51-L57

**System observed on:**
Ubuntu 21.10

**Reporter Info**

jphickey:
**Describe the contribution**
Cleans up the array length calculation for perf structs to not assume a specific base element type (uint32).  This also makes the definitions much more readable.

Fixes #2030

**Testing performed**
Build and sanity check CFE

**Expected behavior changes**
None w/current structure definitions.  Just a bit of future-proofing and clean up.

**System(s) tested on**
Ubuntu 21.10

**Additional context**
Notably the EDS build actually does use a different base type for this array, so this is not just cosmetic.  But this is a worthwhile cleanup regardless.

### DIFF
--- a/modules/es/fsw/src/cfe_es_task.c
+++ b/modules/es/fsw/src/cfe_es_task.c
@@ -48,13 +48,16 @@
 /*
 ** Defines
 */
+#define CFE_ES_PERF_MASK_ARRAY_SIZE(arr) (sizeof(arr) / sizeof(arr[0]))
+
 #define CFE_ES_PERF_TRIGGERMASK_INT_SIZE \
-    (sizeof(CFE_ES_Global.ResetDataPtr->Perf.MetaData.TriggerMask) / sizeof(uint32))
+    CFE_ES_PERF_MASK_ARRAY_SIZE(CFE_ES_Global.ResetDataPtr->Perf.MetaData.TriggerMask)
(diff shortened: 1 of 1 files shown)


skliper (approved):

astrogeco:
[CCB:2022-01-19](https://github.com/nasa/cFS/wiki/CCB:-2022.01.19) APPROVED
- makes this "type agnostic"

Issue #1300, opened 2021-04-12
Create structure to associated CFE_ES_MemAddress_t and CFE_ES_MemOffset_t


skliper:
**Is your feature request related to a problem? Please describe.**
See https://github.com/nasa/cFE/pull/1283#discussion_r611867642

CFE_ES_MemAddres_t and CFE_ES_MemOffset_t are often used in pairs.

**Describe the solution you'd like**
The two could be combined into (for example) CFE_ES_MemAddOff_t

**Describe alternatives you've considered**
None

**Additional context**
From code review #1283

**Requester Info**
Jacob Hageman NASA/GSFC, submitted by @stashakk

astrogeco:
CCB:2022-01-27, 
- Will wait until next major release to implement

PR #2037, opened 2022-01-25
Fix #2036, Revert breaking change to CFE_ES_AppInfo_t structure

### LINKED ISSUE #236: Clean up unit tests to fully use new UT assert
The unit tests in CFE were ported to use UT assert by "bridging" the existing code structure to the UT assert framework using some adaption layers, keeping the same basic API calls in the tests themselves.

This worked for the time but created a fairly ugly and confusing situation, with a number of wrappers.  Also, many global variables still remain that are no longer necessary, they are only remnants of the bridge logic.

The GHAPS project has done quite a bit of work to clean up the unit test code and remove the unnecessary glue logic, making the test code much cleaner.  It also provides a usable base for CFS application unit tests.

### LINKED ISSUE #2036: Revert app info structure modification (#1300, #1911), relative to Draco development cycle
`CFE_ES_AppInfo_t` is public so the change in #1911/#1300 is breaking.  Shouldn't have been included in Draco.  Makes CS non-portable between Caelum and Draco.

Revert.

Could add accessor functions or some sort of abstraction/layer or back-port, but the change really is very low priority.  Could be considered in a future major update cycle if there's some actual benefit from the additional structure.

Could make CS adapt also (conditional compilation or whatever), but I don't think it's worth it for this sort of change or carrying that option.

**Additional context**
Need to decide if we want to reopen #1300 as future work or leave closed.

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
**Describe the contribution**
- Fix #2036

**Testing performed**
CI (unit test build/execute)

**Expected behavior changes**
No behavior change but reverts breaking software change (public structure modification)

**System(s) tested on**
CI

**Additional context**
Need to decide if we should reopen #1300 as future work for a major update cycle

**Third party code**
None

Jacob Hageman - NASA/GSFC


jphickey (approved):

astrogeco:
CCB:2022-01-27, APPROVED
- Reopen #1300, will to wait until a major change and flesh out the use case

PR #2109, opened 2022-05-25
Fix #2095, Automatic suppression of flooding events

### LINKED ISSUE #2095: Automatic suppression of flooding events
We've observed that overly verbose event messages can take down a processor, especially if the generation of events can induce the production of more events due to increased SB/SBNg traffic. Some ability to rate-limit events regardless
of filter state seems to be desirable, as the default for many apps is often unfiltered, and during an event storm it may not be possible to command an event to be filtered.

In addition, there are concerns about critical event messages being lost amongst a flood of less critical events, especially when not commandable due to being in a temporary loss-of-signal state.

There's also the desire to automatically re-enable an event if it is no longer flooding, and the proposed solution would have that characteristic. This is important if the event notifying that the filter has been activated gets lost due to flooding.
Events could still be permanently suppressed w/ the existing binary filters. 

A filter that is applied per-app, per event type (priority). A token counter for (app, event type)  is incremented every time an event message arrives for (app, event type). Meanwhile, the token counter is decremented at a fixed rate.
If the token counter exceeds a certain threshold, incrementing stops and all incoming events for that (app, event type) are discarded until the counter is below a threshold (could be the same as the suppress threshold, then it'll
effectively rate limit after allowing a small burst).

To simplify the implementation, this could be done globally for all events instead of on a per-app basis. There would be the problem of a misbehaved app flooding out another app sending more critical event messages, but this could
be mitigated somewhat by doing this per-event type or prioritizing by event type.

A more granular solution would filter only excessively verbose events by (app event ID) but this would probably require registration of all events and enforcement. It appears that not all apps register all events w/ EVS (i.e. SBNg).

Another solution that only addresses the auto-reenabling of formerly flooding events would be to add another filter type besides binary filter that uses the above mentioned token scheme. This doesn't help too much since the filter 
type is set by the app during registration, if events are registered at all.

**Additional context**
Initial driver was SBNg doesn't come w/ default filters for floodable events, and one of our other apps also sends off a lot of events on startup. Obvious solution is to modify these apps to setup default filters, but this doesn't address reenabling the events automatically when the transients go away, nor does it address unexpectedly flooding messages due to unforeseen edge cases. 

**Requester Info**
John N Pham, Northrop Grumman

jhnphm:

* [ ] I signed and emailed the appropriate [Contributor License Agreement](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) to GSFC-SoftwareRelease@mail.nasa.gov and copied cfs-program@lists.nasa.gov.

**Describe the contribution**
- Fixes #2095 

**Testing performed**
1. Built with SIMULATION=i686-native ENABLE_UNIT_TESTS=true
1. Ran unit tests, all passing except time due to local time configuration changes
1. Built for VxWorks
1. Ran cFS on vxWorks and observed events being squelched and the squelch event message being emitted

**Expected behavior changes**
A clear and concise description of how this contribution will change behavior and level of impact.
 - API Change: N/A
 - Behavior Change: Events get squelched after 16 events within 1 second per app (default in cpu1_platform_cfg); tokens are returned at a rate of 8/sec (default in cpu1_platform_cfg) to allow a sustained event rate of 8/sec/app. Event is emitted upon squelching, and counters incremented. Counters added to event data stored to file and realtime telemetry. 

**System(s) tested on**
 - Hardware: SP0-s (Simics)
 - OS: VxWorks 6.9
 - Versions: cFE 7.0.0+dev127, OSAL  v6.0.0-rc4+dev66, OSAL  v6.0.0-rc4+dev66 w/ JSC changes
 
**Additional context**
N/A

**Third party code**
N/A

John N Pham, Northrop Grumman, Space Systems


jhnphm:
CLA incoming; will have to submit to my management to fill out

skliper:
@jhnphm We just got a new consolidated form that includes all the apps, I'll email it.

jhnphm:
Updated unit test to improve coverage, and fixed doc errors. Updated commit message to add more detail.

modules/evs/fsw/src/cfe_evs_task.h
```diff
@@ -89,6 +90,9 @@ typedef struct
     uint8  ActiveFlag;           /* Application event service active flag */
     uint8  EventTypesActiveFlag; /* Application event types active flag */
     uint16 EventCount;           /* Application event counter */
+    CFE_TIME_SysTime_t LastSquelchReturnedTime; /* Time of last squelch token return */
+    uint16             SquelchTokens;           /* Application event squelch token counter */
```
jphickey, line 94: Recommend using a signed type for the tokens.  This makes it simpler, if it goes negative, the events are squelched.  As long as its positive, its OK to send events.

modules/evs/fsw/src/cfe_evs_task.h
```diff
@@ -89,6 +90,9 @@ typedef struct
     uint8  ActiveFlag;           /* Application event service active flag */
     uint8  EventTypesActiveFlag; /* Application event types active flag */
     uint16 EventCount;           /* Application event counter */
+    CFE_TIME_SysTime_t LastSquelchReturnedTime; /* Time of last squelch token return */
```
jphickey, line 93: Recommend to use `OS_time_t` here (goes along other comment about using CFE_PSP_GetTime)

modules/evs/fsw/src/cfe_evs_utils.c
```diff
@@ -288,6 +289,92 @@ bool EVS_IsFiltered(EVS_AppData_t *AppDataPtr, uint16 EventID, uint16 EventType)
...
+        }
+        else
+        {
+            AppDataPtr->SquelchTokens++;
+            /* NotSquelched already true */
+        }
+    }
+    return NotSquelched;
+}
+
```
jphickey, line 377: General comments on this whole thing, it still seems really complicated to me.  It also seems a bit paradoxical to send an event to say an app is sending too many events right now.

I understand the desire/intent to use a clock difference to restore credit, but I would argue the units are too coarse, aside from ensuring you use a monotonic clock, not a settable clock, that I already noted.  

Instead of crediting one token per event, make each event "cost" (for example) 1000 tokens.  Then give credit per millisecond instead of per second.

Make sending an event be a decrement, not an increment.  This makes it easy to check by checking the sign of your token credit.  If its negative, you cannot send.  Likewise on the increment side you just need to cap how high it can go (e.g. 1000 * CFE_PLATFORM_EVS_MAX_APP_EVENT_BURST).

I'd also consider make the event notification happen on the transition from negative to positive, rather than positive to negative.  This way, your event will be more useful, in that it will show the actual number of events squelched during the spewing period.  (I am guessing, if I'm reading the code correctly, that if an app starts spewing events the event that comes out will say "1" because it comes out only at the first squelch, even if it was the first of 1000 such issues ... by waiting till it actually stops spewing, you'll report the actual number that were surpressed during the spew).

jphickey (changes requested):
I think overall there are some concerns that need to be resolved here.  Simplify the logic, mainly.
I didn't look too closely at the unit tests but it seems to require modification to every test case.  This is definitely not ideal, I'd try to avoid that if at all possible.  I would think that it could be worked around by giving each app a huge amount of credit during the common setup function (e.g. INT32_MAX if you also use my suggestion of going with a signed credit value and a decrement)

modules/evs/fsw/src/cfe_evs_utils.c
jhnphm: I'm kinda hesitant to have no event message for too many events, even though it seems paradoxical, since otherwise one would have to be actively looking for the event-squelched counter in the per-app telemetry (which isn't currently resettable for some reason) to know why events are mysteriously disappearing.

Emitting it when the token count goes from negative to positive as you mentioned below seems like it'd address this paradox, since you're emitting the event when there's "room" (although you'll never emit it if spamming doesn't stop- I suppose could emit another event message if the 8-bit squelch count is about to roll over, or some other threshold)

It's already giving credit at a rate of 1 per 1s/CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC by dividing out the subseconds. Logic does get simpler if using OS_TimeGetTotalMilliseconds since that's one number to compare rather than whole seconds + subseconds. 

Otherwise I agree w/ everything else, and thanks for the feedback!

skliper: Although I've also debated this a bit internally, I'm in favor of leaving the event message in as an indication an app is sending too many events.  It's the app events that are flooding and causing the problem, not a single additional event that is well behaved.  It also helps in the event "stream" (log or console or ground reporting of event messages) to see why the flooding events went away, becomes immediately identifiable operationally vs an assumption that the issue causing the flooding might have actually stopped.  If a system is really on the margins where +1 event will cause an issue, then I'd suggest just reducing the flooding limit by 1 vs dropping this helpful additional event.

jhnphm:
> I think overall there are some concerns that need to be resolved here. Simplify the logic, mainly. I didn't look too closely at the unit tests but it seems to require modification to every test case. This is definitely not ideal, I'd try to avoid that if at all possible. I would think that it could be worked around by giving each app a huge amount of credit during the common setup function (e.g. INT32_MAX if you also use my suggestion of going with a signed credit value and a decrement)

I simplified it a bit, but still requires resetting the token count to `INT32_MAX` after each successful call to `CFE_EVS_Register`, and I had to put a case in to ignore adding credits if the existing token count was > `CFE_PLATFORM_EVS_MAX_APP_EVENT_BURST`. I guess I could add a flag to CFE_EVS_Global or one of it's substructs to properly disable squelching globally at runtime (and at that point might as well make it a command), but that would add more complexity.

modules/evs/fsw/src/cfe_evs_utils.c
```diff
@@ -288,6 +289,105 @@ bool EVS_IsFiltered(EVS_AppData_t *AppDataPtr, uint16 EventID, uint16 EventType)
...
+                AppDataPtr->SquelchTokens = UPPER_THRESHOLD;
+            }
+            else
+            {
+                AppDataPtr->SquelchTokens += (int32)CreditCount;
+            }
+        }
+
+        /*
+         * Send event messages if we're spamming (need a minimum credit
```
jhnphm, line 357: Don't send

modules/evs/fsw/src/cfe_evs_utils.c
```diff
@@ -288,6 +289,105 @@ bool EVS_IsFiltered(EVS_AppData_t *AppDataPtr, uint16 EventID, uint16 EventType)
...
+        {
+            /* Update last squelch returned time if we credited any tokens*/
+            AppDataPtr->LastSquelchCreditableTime = CurrentTime;
+
+            /*
+             * Add Credits, to a maximum of UPPER_THRESHOLD
+             * Shouldn't rollover, as calculations are done in int64 space due to
+             * promotion rules then bounded before demotion
+             */
+            if (AppDataPtr->SquelchTokens > UPPER_THRESHOLD)
```
jhnphm, line 342: Instead of using CFE_PLATFORM_EVS_MAX_APP_EVENT_BURST directly, set a global variable to this. Solves multiple problems for testability.

modules/evs/fsw/src/cfe_evs_utils.c
```diff
@@ -288,6 +289,105 @@ bool EVS_IsFiltered(EVS_AppData_t *AppDataPtr, uint16 EventID, uint16 EventType)
...
+    {
+        /*
+         * We use a timer here since configurations are not guaranteed to send EVS HK wakeups at 1Hz
+         * Use a non-settable timer to prevent this from breaking w/ time changes
+         */
+        CFE_PSP_GetTime(&CurrentTime);
+        DeltaTimeMs = OS_TimeGetTotalMilliseconds(OS_TimeSubtract(CurrentTime, AppDataPtr->LastSquelchCreditableTime));
+
+        /* Calculate how many tokens to credit in elapsed time since last creditable event */
+        CreditCount = DeltaTimeMs * CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC / 1000;
```
jhnphm, line 330: Use 1000 credits per message rather than dividing to prevent integer division weirdness and inefficiencies.

modules/evs/fsw/src/cfe_evs.c
```diff
@@ -71,6 +71,7 @@ CFE_Status_t CFE_EVS_Register(const void *Filters, uint16 NumEventFilters, uint1
             /* Initialize application event data */
             AppDataPtr->ActiveFlag           = true;
             AppDataPtr->EventTypesActiveFlag = CFE_PLATFORM_EVS_DEFAULT_TYPE_FLAG;
+            AppDataPtr->SquelchTokens        = CFE_PLATFORM_EVS_MAX_APP_EVENT_BURST;
```
jhnphm, line 74: Use intermediate global variable.

modules/evs/fsw/inc/cfe_evs_events.h
```diff
@@ -466,6 +466,19 @@
...
+ *
+ *  \par Type: ERROR
+ *
+ *  \par Cause:
+ *
+ *  Events generated in app at a rate in excess of
+ *  #CFE_PLATFORM_EVS_MAX_APP_EVENT_BURST in one moment or
+ *  #CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC sustained
+ */
+#define CFE_EVS_SQUELCHED_ERR_EID 43
```
jhnphm, line 481: this should be 44

jhnphm:
@skliper @jphickey Added the changes requested/bugs pointed out during today's meeting. I'm using the global EVS mutex if that's okay. Also updated docs and took a first cut at requirements in the CSV.

Not sure if the diagram included is detailed enough; it's a notional state diagram, but there really isn't a state machine in the code. I have this activity diagram but it might be too "busy":
evs_squelch

skliper:
Closed in favor of #2117

justinvvitale:
> Closed in favor of #2117

Damn skliper you really did jhnphm dirty...

skliper:
@justinvvitale - I should have been more verbose in my closure statement.  Transfer was coordinated w/ all parties involved to straighten out contribution ownership based on the various organizations and contracts involved with this work.  There's significant process involved behind the scenes in some external contribution situations, and I realize the optics don't look good w/o context. :)

PR #2117, opened 2022-06-09
Fix #2095, Automatic suppression of flooding events

### LINKED ISSUE #2095: Automatic suppression of flooding events

skliper:

**Describe the contribution**
- Fix #2095 

**Testing performed**
CI

**Expected behavior changes**
Squelches flooding events (configurable)

**System(s) tested on**
CI

**Additional context**
Still a work in progress, needs coverage and functional cleanup

**Third party code**
None, update suggestion from NASA/JSC context.

Jacob Hageman - NASA/GSFC

Ping @jhnphm


TheSwarty (approved):

cmake/sample_defs/cpu1_platform_cfg.h
```diff
@@ -1337,6 +1337,29 @@
...
+**  \cfeevscfg Maximum number of event before squelching
+**
+**  \par Description:
+**       Maximum number of events that may be emitted per app per second.
+**       Setting this to 0 will cause events to be unrestricted.
+**
+**  \par Limits
+**       This number must be less than or equal to INT_MAX/1000
+*/
+#define CFE_PLATFORM_EVS_MAX_APP_EVENT_BURST 16
```
semaldona, line 1350: Do we want this feature enabled by default?

skliper:
@jhnphm - this is planned for CCB on 8/3.  Could you review the recent minor changes?  Thanks for putting this together!

modules/evs/ut-coverage/evs_UT.c
```diff
@@ -1814,6 +1862,126 @@ void Test_InvalidCmd(void)
...
+                                                              1000 / CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC);
+        UT_SetDataBuffer(UT_KEY(CFE_PSP_GetTime), &InjectedTime, sizeof(InjectedTime), false);
+        EVS_Retval = SendEventFuncs[j](EVENT_ID);
+        UtAssert_UINT32_EQ(SnapshotData.Count, CFE_PLATFORM_EVS_MAX_APP_EVENT_BURST + 4);
+        UtAssert_UINT32_EQ(CapturedTlm.Payload.PacketID.EventID, EVENT_ID);
+        UtAssert_UINT32_EQ(EVS_Retval, CFE_SUCCESS);
+        UtAssert_UINT32_EQ(AppDataPtr->SquelchedCount, CFE_EVS_MAX_SQUELCH_COUNT);
+
+        /*
+         * Test boundary condition where SquelchTokens > -EVENT_COST && CreditCount >= EVENT_COST
```
jhnphm, line 1970: Update comment to state `SquelchTokens > |-EVENT_COST| && CreditCount >= EVENT_COST && CreditCount < |SquelchTokens|`

modules/evs/ut-coverage/evs_UT.c
```diff
@@ -1814,6 +1862,126 @@ void Test_InvalidCmd(void)
...
+         */
+        InjectedTime = OS_TimeAssembleFromMilliseconds(1, 150);
+        UT_SetDataBuffer(UT_KEY(CFE_PSP_GetTime), &InjectedTime, sizeof(InjectedTime), false);
+        AppDataPtr->SquelchedCount            = 0;
+        AppDataPtr->SquelchTokens             = -1500;
+        AppDataPtr->LastSquelchCreditableTime = OS_TimeAssembleFromMilliseconds(1, 0);
+        EVS_Retval                            = SendEventFuncs[j](EVENT_ID);
+        UtAssert_UINT32_EQ(EVS_Retval, CFE_EVS_APP_SQUELCHED);
```
jhnphm, line 1978:
```
If CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC == 8:
    150ms * 8 == positive credit count of 1200 which is < 1500
    test passes
however, if CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC == 16:
    150ms * 16 == positive credit count of 2400 which is > 1500
    test fails
 ```   
 Injected time difference should be 1200/CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC  rather than hardcoded 150ms to account for varying values of CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC (up to a point where CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC happens to be greater than an unreasonable 1200 and breaks the test again)

**Review by jhnphm — CHANGES_REQUESTED** [review]
Suggesting the following changes to make the test code more robust against adjustments to `CFE_PLATFORM_EVS_APP_EVENTS_PER_SEC`

File: modules/cfe_testcase/src/evs_send_test.c [hunk 4]
```diff
@@ -44,6 +53,22 @@ void TestSendEvent(void)
 
     UtAssert_INT32_EQ(CFE_EVS_SendEvent(0, CFE_EVS_EventType_CRITICAL, "OK (Critical) Send"), CFE_SUCCESS);
     UtAssert_INT32_EQ(CFE_EVS_SendEvent(0, CFE_EVS_EventType_CRITICAL, NULL), CFE_EVS_INVALID_PARAMETER);
```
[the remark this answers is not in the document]
semaldona: Agreed. Will proceed with documentation update.

modules/evs/ut-coverage/evs_UT.c
semaldona: Agreed. Will proceed with unit test update.

modules/evs/ut-coverage/evs_UT.c
semaldona: This test was added to address branch coverage of the commented case in the if statement, specifically where CreditCount >= EVENT_COST.  The intention was to not cross the threshold or cause SendSquelchEvent = true.

Issue #1511, opened 2021-05-13
Convert to static array sizes for string parameters - `name[static SIZE]` and scrub string handling


skliper:
**Is your feature request related to a problem? Please describe.**
Many APIs and even internal functions require specific minimum string sizes to avoid a potential buffer overrun, yet take simple pointers, i.e. `char *`.  Some tools complain about strcpy/sprintf/etc, but the sized versions aren't any safer if the passed in string is too short.

**Describe the solution you'd like**
Leverage C99 checks on sized parameter arrays, which also will protect at compile time from NULL inputs.  Once minimum sizes are guaranteed, then revisit string comparison/manipulation calls to avoid buffer overruns based on these enforced minimums.

Note not backwards compatible where pointers are passed in vs the array.  Also would REQUIRE the removal of all NULL pointer test cases and checks (would cause compile errors or dead code respectively).

**Describe alternatives you've considered**
None

**Additional context**
Code review, related to #1465

**Requester Info**
Jacob Hageman - NASA/GSFC

skliper:
Should actually confirm all our supported compilers are compliant (check NULL and undersized arrays) before advancing this change.  At minimum the one used for CI should confirm the core software as configured is good.

Issue #1465, opened 2021-04-30
Use sized char array for `CFE_SB_GetAppTskName` parameter


skliper:
**Is your feature request related to a problem? Please describe.**
`CFE_SB_GetAppTskName` takes a char pointer `FullName` without an associated size.  Relies on user to size the buffer correctly.

**Describe the solution you'd like**
Related to #1457, explicitly size as an array.  Note #994 suggests moving this to a generic utility (out of SB scope), maybe ES API.

**Describe alternatives you've considered**
None

**Additional context**
Code review

**Requester Info**
Jacob Hageman - NASA/GSFC

PR #2173, opened 2022-10-18
Fix #2172, Remove last few uses of sprintf()

### LINKED ISSUE #2172: Remove the few remaining stray uses of sprintf()?
I concur that `snprintf` has some caveats/dangers - particularly with respect to the return code and handling that properly - whereas `sprintf()` is a buffer overflow exploit in almost all use-cases, and should rightfully be banned.

_Originally posted by @jphickey in https://github.com/nasa/cFE/issues/824#issuecomment-677851081_
      
There are still a few stray uses of sprintf() across cFS. Should these be removed?

thnkslprpt:
**Checklist**

**Describe the contribution**
- Fixes #2172 
  - Use of sprintf() swapped for snprintf() removing a source of potential buffer overruns.

2 other cases already have their own issues open with further changes being considered, and have not been updated with this PR at this stage (https://github.com/nasa/cFE/issues/1465 and https://github.com/nasa/cFE/issues/1511).

**Testing performed**
Standard cFS build tests (covered the fsw change but not test code change).

**Expected behavior changes**
No impact on behavior expected.

**System(s) tested on**
Intel(R) Celeron(R) N4100 CPU @ 1.10GHz x86_64
Debian GNU/Linux 11 (bullseye)
cFE v7.0.0-rc4+dev193

**Contributor Info**
Avi Weiss @thnkslprpt

### DIFF
--- a/modules/cfe_testcase/src/tbl_content_access_test.c
+++ b/modules/cfe_testcase/src/tbl_content_access_test.c
@@ -118,7 +118,7 @@ void TestGetReleaseAddresses(void)
     TblPtrs[0]    = TblPtrsList;
     for (int i = 1; i < numValidTbls + 1; i++)
     {
-        sprintf(TblName, "%d", i);
+        snprintf(TblName, sizeof(TblName), "%d", i);
         UtAssert_INT32_EQ(
             CFE_TBL_Register(&TblHandles[i], TblName, sizeof(TBL_TEST_Table_t), CFE_TBL_OPT_DEFAULT, NULL),
             CFE_SUCCESS);


thnkslprpt:
Would it be worth updating the other 2 cases remaining to snprintf() with this PR, and leaving their respective issues open for the further changes raised there to be considered?

These would be the changes:
Screenshot 2022-10-18 12 20 13
Screenshot 2022-10-18 12 18 37

dzbaker:
@thnkslprpt Do you know why the Functional Test check is failing?

thnkslprpt:
> @thnkslprpt Do you know why the Functional Test check is failing?

Will figure it out and get back to you.

thnkslprpt:
> @thnkslprpt Do you know why the Functional Test check is failing?

OK it seems to be passing now.
The 2nd parameter (bufsz) of snprintf() was decaying to a pointer causing the function to print an empty string.
By using CFE_MISSION_ES_CDS_MAX_FULL_NAME_LEN as the bufsz parameter, which is the size of the passed-in FullCDSName for the CFE_ES_FormCDSName() fuction, the snprintf() functions correctly.

Ref:
https://github.com/nasa/cFE/blob/be88a07313ff3b579c8abaef4b4a9085e74aabdf/modules/es/fsw/src/cfe_es_cds.h#L536

modules/es/fsw/src/cfe_es_cds.c
```diff
@@ -659,7 +659,7 @@ void CFE_ES_FormCDSName(char *FullCDSName, const char *CDSName, CFE_ES_AppId_t T
     AppName[OS_MAX_API_NAME - 1] = '\0';
 
     /* Complete formation of processor specific table name */
-    sprintf(FullCDSName, "%s.%s", AppName, CDSName);
+    snprintf(FullCDSName, CFE_MISSION_ES_CDS_MAX_FULL_NAME_LEN, "%s.%s", AppName, CDSName);
```
jphickey, line 662: I don't like this one.  In this case the API/function prototype did not provide a length at all, but assuming the user always has provided a buffer of `CFE_MISSION_ES_CDS_MAX_FULL_NAME_LEN` arguably is just as bad as calling `sprintf()` here.  Note that the full name length is computed to be the worst case length of `CFE_MISSION_ES_CDS_MAX_NAME_LENGTH` and `OS_MAX_API_NAME` put together, so if the buffer actually is at least this length, there is no chance of overflow.

The fundamental flaw of `sprintf()` is that it just assumes the buffer is long enough to hold the result.  Switching this to calling `snprintf()` with an "assumed long enough" length parameter doesn't fix the problem, it is still assuming the length is long enough when it might not actually be.

I think this particular function (CFE_ES_FormCDSName) is a broken API, and should be removed (its not public).  It is called in just one place - from CFE_ES_RegisterCDS() - and that can just do an `snprintf` directly.  It actually needs to know the app name anyway for an event later on, so this would also avoid getting it twice.  A win-win.

jphickey (changes requested):

jphickey (approved):
This change is good, however we should also not forget about the other one (as noted, my suggestion is to remove the broken API, clean up the caller).

thnkslprpt:
> This change is good, however we should also not forget about the other one (as noted, my suggestion is to remove the broken API, clean up the caller).

No worries Joe.
https://github.com/nasa/cFE/issues/2193

PR #2192, opened 2022-10-31
Fix #2174, Move CRC types and convert to enum

### LINKED ISSUE #2174: ES CRC type enumerations do not belong in cfe_mission_cfg.h
The mission config header file contains the following definitions:

https://github.com/nasa/cFE/blob/be88a07313ff3b579c8abaef4b4a9085e74aabdf/cmake/sample_defs/sample_mission_cfg.h#L261-L268

These enumeration values are really part of the API to the `CFE_ES_CalculateCRC()` function, as they define the values for the 4th parameter.   

They do not belong in "mission_cfg.h" because are not values that appear in any telemetry, nor in any command, nor would they need to be configured differently as there is no chance of conflict with anything else.

Make a proper enumeration type in the in the `cfe_es_api_typedefs.h` file, and use this for the 4th parameter of the `CFE_ES_CalculateCRC()` function.

**Additional context**
The "DEFAULT" value is the only thing that (theoretically) does belong here:

https://github.com/nasa/cFE/blob/be88a07313ff3b579c8abaef4b4a9085e74aabdf/cmake/sample_defs/sample_mission_cfg.h#L281

However, this is just a placeholder anyway because only the CRC16 algorithm was ever implemented.  Unless the user also has a patched version of ES, no user can set this to anything other than CRC16 or else the system will not function correctly.  Because this is somewhat of a "dead" option, and a case could be made to remove it entirely.

**Requester Info**

thnkslprpt:
**Checklist**

**Describe the contribution**
Fixes #2174 
The macro definitions of the CRC types (8, 16, and 32-bit CRC polynomials) have been moved to `cfe_es_api_typedefs.h` and turned into a typedef'd `enum`.
The previous `#define`'s have been left in with deprecation directives added around them, due to dependencies of the FM app.

**Testing performed**
GitHub CI actions (incl. Build + Run, Unit Tests etc.) all passing successfully.

**Expected behavior changes**
Essentially no change to logic.

Note - Two new issues will need to be opened if/when this is merged:
FM: directly references the types that have been changed here.
CS: refers to (in comments only) `CFE_MISSION_ES_CRC_16`

**Contributor Info**
Avi Weiss @thnkslprpt

### DIFF
--- a/cmake/sample_defs/sample_mission_cfg.h
+++ b/cmake/sample_defs/sample_mission_cfg.h
@@ -258,14 +258,16 @@
 */
 #define CFE_MISSION_EVS_MAX_MESSAGE_LENGTH 122
 
+#ifndef CFE_OMIT_DEPRECATED_6_8
+/* These names have been converted to an enum in cfe_es_api_typedefs.h */
+
 /** \name Checksum/CRC algorithm identifiers */
-/** \{ */
-#define CFE_MISSION_ES_CRC_8  1 /**< \brief CRC ( 8 bit additive - returns 32 bit total) (Currently not implemented) */
-#define CFE_MISSION_ES_CRC_16 2 /**< \brief CRC (16 bit additive - returns 32 bit total) */
-#define CFE_MISSION_ES_CRC_32                                                              \
-    3 /**< \brief CRC (32 bit additive - returns 32 bit total) (Currently not implemented) \
-       */
-/** \} */
+
+#define CFE_MISSION_ES_CRC_8  CFE_ES_CrcType_CRC_8  /* 1 */
+#define CFE_MISSION_ES_CRC_16 CFE_ES_CrcType_CRC_16 /* 2 */
+#define CFE_MISSION_ES_CRC_32 CFE_ES_CrcType_CRC_32 /* 3 */
+
+#endif
 
 /**
 **  \cfeescfg Mission Default CRC algorithm
@@ -276,9 +278,10 @@
 **      Table Image data integrity values.
 **
 **  \par Limits
-**      Currently only CFE_MISSION_ES_CRC_16 is supported (see #CFE_MISSION_ES_CRC_16)
+**      Currently only CFE_ES_CrcType_CRC_16 is supported (see brief in CFE_ES_CrcType_Enum
+**      definition in cfe_es_api_typedefs.h)
 */
-#define CFE_MISSION_ES_DEFAULT_CRC CFE_MISSION_ES_CRC_16
+#define CFE_MISSION_ES_DEFAULT_CRC CFE_ES_CrcType_CRC_16
 
 /**
 **  \cfetblcfg Maximum Table Name Length
--- a/docs/cFE Application Developers Guide.md
+++ b/docs/cFE Application Developers Guide.md
@@ -1293,24 +1293,28 @@ an API for a CRC calculation that can be used by all Applications on a mission.
 This function looks like the following:
 
 ```c
-uint32 CFE_ES_CalculateCRC(void *pData, uint32 DataLength, uint32 InputCRC, uint32 TypeCRC);
+uint32 CFE_ES_CalculateCRC(const void *DataPtr, size_t DataLength, uint32 InputCRC, CFE_ES_CrcType_Enum_t TypeCRC);
 ```
 
-where pData points to the first byte of an array of bytes that are to have
+where DataPtr points to the first byte of an array of bytes that are to have
 the CRC calculated on, DataLength specifies the number of sequential bytes to
 include in the calculation, InputCRC is the initial value of the CRC and
 TypeCRC identifies which of the standard CRC polynomials to be used.  Currently,
 there are the following types available:
 
 ```
-CFE_MISSION_ES_CRC_8 – an 8-bit additive checksum calculation that returns a 32-bit value
-CFE_MISSION_ES_CRC_16 – a 16-bit additive checksum calculation that returns a 32-bit value
-CFE_MISSION_ES_CRC_32 – a 32-bit additive checksum calculation that returns a 32-bit value
-CFE_MISSION_ES_DEFAULT_CRC – the mission specified default CRC calculation
+CFE_ES_CrcType_CRC_8 – an 8-bit additive checksum calculation that returns a 32-bit value
+CFE_ES_CrcType_CRC_16 – a 16-bit additive checksum calculation that returns a 32-bit value
+CFE_ES_CrcType_CRC_32 – a 32-bit additive checksum calculation that returns a 32-bit value
+CFE_MISSION_ES_DEFAULT_CRC – the mission specified default CRC calculation (currently
+                             this is set to CFE_ES_CrcType_CRC_16 in sample_mission_cfg.h)
 ```
 
-Unless there is a specific interface with a specified CRC calculation,
-Applications must use the CFE_MISSION_ES_DEFAULT_CRC type.
+Unless there is a specific interface with a specified CRC calculation, applications
+must use the CFE_MISSION_ES_DEFAULT_CRC type.
+
+Currently only CFE_ES_CrcType_CRC_16 is supported. CFE_ES_CrcType_CRC_8 and CFE_ES_CrcType_CRC_32 are yet
+to be implemented.
 
 ## 5.11 File System Functions
 
--- a/modules/cfe_testcase/src/es_misc_test.c
+++ b/modules/cfe_testcase/src/es_misc_test.c
@@ -41,19 +41,19 @@ void TestCalculateCRC(void)
     UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), 0, CFE_MISSION_ES_DEFAULT_CRC));
     UtAssert_MIR("Confirm mission default CRC of \"%s\" is %lu", Data, (unsigned long)Result);
 
-    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), inputCrc, CFE_MISSION_ES_CRC_16));
+    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), inputCrc, CFE_ES_CrcType_CRC_16));
     UtAssert_MIR("Confirm CRC16 of \"%s\" with input CRC of %lu is %lu", Data, (unsigned long)inputCrc,
                  (unsigned long)Result);
 
-    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), 0, CFE_MISSION_ES_CRC_8));
+    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), 0, CFE_ES_CrcType_CRC_8));
     UtAssert_MIR("Confirm CRC8 of \"%s\" is %lu", Data, (unsigned long)Result);
 
-    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), 0, CFE_MISSION_ES_CRC_32));
+    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), 0, CFE_ES_CrcType_CRC_32));
     UtAssert_MIR("Confirm CRC32 of \"%s\" is %lu", Data, (unsigned long)Result);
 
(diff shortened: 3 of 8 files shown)


docs/cFE Application Developers Guide.md
```diff
@@ -1303,14 +1303,13 @@ TypeCRC identifies which of the standard CRC polynomials to be used.  Currently,
...
-CFE_MISSION_ES_CRC_16 – a 16-bit additive checksum calculation that returns a 32-bit value
-CFE_MISSION_ES_CRC_32 – a 32-bit additive checksum calculation that returns a 32-bit value
-CFE_MISSION_ES_DEFAULT_CRC – the mission specified default CRC calculation
+CFE_ES_CRC_8 – an 8-bit additive checksum calculation that returns a 32-bit value
+CFE_ES_CRC_16 – a 16-bit additive checksum calculation that returns a 32-bit value
+CFE_ES_CRC_32 – a 32-bit additive checksum calculation that returns a 32-bit value
 ```
 
-Unless there is a specific interface with a specified CRC calculation,
-Applications must use the CFE_MISSION_ES_DEFAULT_CRC type.
```
jphickey, line 1313: Given that using/referring to the `CFE_MISSION_ES_DEFAULT_CRC` symbol has been the documented way to call this function, I would not want to change this.  We should continue to define this symbol as it is likely used by apps for this described purpose.

modules/cfe_testcase/src/es_misc_test.c
```diff
@@ -28,32 +28,37 @@
...
+    const char *     Data     = "Random Stuff";
+    uint32           inputCrc = 345353;
+    uint32           Result;
+    CFE_ES_CrcType_t TypeCrc = CFE_ES_CRC_16;
 
     UtPrintf("Testing: CFE_ES_CalculateCRC");
 
     /* CRC is implementation specific, functional just checks that a result is produced and reports */
-    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), 0, CFE_MISSION_ES_DEFAULT_CRC));
+    UtAssert_VOIDCALL(Result = CFE_ES_CalculateCRC(Data, sizeof(Data), 0, TypeCrc));
```
jphickey, line 43: Let's not change this one.

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
```diff
@@ -83,6 +83,18 @@ typedef CFE_ES_TaskEntryFuncPtr_t CFE_ES_ChildTaskMainFuncPtr_t;
...
+ *
+ * Currently only CFE_ES_CRC_16 is supported.
+ */
+typedef enum CFE_ES_CrcType
+{
+    CFE_ES_CRC_8,  /**< \brief CRC ( 8 bit additive - returns 32 bit total) (Currently not implemented) */
+    CFE_ES_CRC_16, /**< \brief CRC (16 bit additive - returns 32 bit total) */
+    CFE_ES_CRC_32  /**< \brief CRC (32 bit additive - returns 32 bit total) (Currently not implemented) */
+} CFE_ES_CrcType_t;
+
```
jphickey, line 97: To adhere to the enum naming convention, the type should be named `CFE_ES_CrcType_Enum_t` and the labels should be `CFE_ES_CrcType_CRC{_ 8,_16,_32}`  (i.e. consistent use of CFE_ES_CrcType_ prefix for both type and labels, with same capitalization)

jphickey (changes requested):
A couple modifications requested here.

Note - this is not just to be pedantic, there is a reason behind both the naming convention as well as putting the things at the right scope and in the right file and named the right way, because eventually some of these headers will become generated, so the names and content needs to match to allow that transition.  That was really the reason this change was needed in the first place.

modules/cfe_testcase/src/es_misc_test.c
thnkslprpt: Removed from the changes.

docs/cFE Application Developers Guide.md
thnkslprpt: No worries.
By the way - I don't take it as pedantic at all. I'm a firm believer in doing things right the first time, and doing them thoroughly (if you're going to do it at all).
Thanks for the additional context.

Re. this change - is it the right approach to leave the `CFE_MISSION_ES_DEFAULT_CRC` #defined and simply setting it to the value of `CFE_ES_CrcType_CRC_16`?
If not, can you advise what would be the best approach?
Cheers

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
thnkslprpt: No worries - I've updated them in the 2nd commit.

docs/cFE Application Developers Guide.md
jphickey: Yes, because the doc said to use `CFE_MISSION_ES_DEFAULT_CRC` and we can assume most/all callers are doing it this way, just by keeping it defined but changing it to `CFE_ES_CrcType_CRC_16` should be relatively safe and low risk of breakage.  There is always the possibility that someone referenced `CFE_MISSION_ES_CRC_16` directly, but that is unlikely enough that I'm not too worried about it.

I do think that the extra option is essentially "dead code" because nobody ever implemented anything else, and thus could be justified in removal, but because it's part of the public ES API we would have to go through the deprecation process (i.e. make a new API and but the old behind an OMIT_DEPRECATED switch, keeping both calls for at least one release cycle, letting users test and fix their app code, before removing the old one outright).  Try to avoid hard cuts/step changes in the API unless absolutely necessary.

cmake/sample_defs/sample_mission_cfg.h
```diff
@@ -35,6 +35,11 @@
 #ifndef SAMPLE_MISSION_CFG_H
 #define SAMPLE_MISSION_CFG_H
 
+/*
+** Includes
+*/
+#include "cfe_es_api_typedefs.h"
+
```
jphickey, line 42: Even though this technically does reference a "CFE_ES_CrcType" enum label now, we don't strictly need this inclusion here, because its just a preprocessor macro, and thus only gets evaluated when it used/referenced.  Anyone who uses the DEFAULT_CRC macro would, by definition, have to include the ES API header to get the prototype for the actual call.  

As a "config file" - I think the preference is to keep these config headers (mission/platform) to be very simple, because they can be used in tools and things like `elf2cfetbl` that might not have access to the ES API header, but can potentially use some of the other values in here.

So while an API or type-defining header should be fully self-contained (i.e. include all its dependencies) that is not really necessary for a pure config header.

There is precedent for this; some of the values in here are true/false which technically depend on `stdbool.h` and also there are examples in the platform config of referencing enum values that are not directly included in the config file.

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
jphickey: Looks good, but one more observation is that the original enum values were 1, 2, and 3 (so CRC_16 was 2).  We should keep the values the same, just in case.

modules/tbl/fsw/src/cfe_tbl_task_cmds.c
```diff
@@ -32,6 +32,7 @@
 */
 #include "cfe_tbl_module_all.h"
 #include "cfe_version.h"
+#include "cfe_es_api_typedefs.h"
```
jphickey, line 35: This inclusion shouldn't be necessary, because "cfe_es.h" should pull in everything needed to call the function.

modules/es/ut-coverage/es_UT.c
```diff
@@ -39,6 +39,7 @@
 #include "es_UT.h"
 #include "target_config.h"
 #include "cfe_config.h"
+#include "cfe_es_api_typedefs.h"
```
jphickey, line 42: Also not needed, as "es_UT.h" pulls in everything for ES, including the API typedefs.

modules/es/fsw/src/cfe_es_task.c
```diff
@@ -38,7 +38,7 @@
 #include "cfe_version.h"
 #include "target_config.h"
 #include "cfe_es_verify.h"
-
+#include "cfe_es_api_typedefs.h"
```
jphickey, line 41: Also not needed ... I think we can back out changes to this file entirely (keep the changeset small and simple).

jphickey (changes requested):
Getting there... but we need to revert back the rest of the `CFE_MISSION_ES_DEFAULT_CRC` references.  I'd back this changeset off to changing only what directly refers to the enum labels (that is, the implementation of the CRC function itself, and the headers that define the labels).  

In addition to keeping this change smaller and simpler, **not** changing other refs also builds some confidence - if the various places that this function is invoked inside CFE core modules (particularly including unit and functional tests) don't have to change, we can be relatively sure that calls from external apps outside CFE core modules probably don't have to change either.

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
thnkslprpt: Yes no worries - I was thinking about that. Will update them now to start at 1.

modules/es/ut-coverage/es_UT.c
thnkslprpt: Good to know - I saw different approaches (including and not including headers recursively).
I think it's better to avoid when not necessary as well.
Will remove where not required.

modules/es/fsw/src/cfe_es_task.c
thnkslprpt: All done - changeset a lot smaller now ;-)

cmake/sample_defs/sample_mission_cfg.h
thnkslprpt: > Getting there... but we need to revert back the rest of the `CFE_MISSION_ES_DEFAULT_CRC` references.

Added them all back - changes are only to the direct references to the enum values.

modules/cfe_testcase/src/es_misc_test.c
```diff
@@ -28,32 +28,37 @@
  */
 
 #include "cfe_test.h"
+#include "cfe_es_api_typedefs.h"
```
jphickey, line 31: Looks like this inclusion is still there.

modules/core_api/fsw/inc/cfe_es_api_typedefs.h
jphickey: Almost there, just a nuance in the GSFC coding standards that says we need to be explicit about every enum label in this case.  In other words, even though ISO C guarantees that an enum label without a value will be assigned the value of the previous label plus 1, the local coding standards say that we must also put this in the source code.

So just add explicit values for 2 and 3 here....

jphickey (changes requested):
Real close now, just a couple minor details still remaining

modules/cfe_testcase/src/es_misc_test.c
thnkslprpt: Removed.

thnkslprpt: Added the explicit enum labes.
Cheers Joe.

jphickey (changes requested):
Reviewed again, but I've come to 2 realizations about this:
1. Unfortunately, the FM app refers to the other (non-implemented) CRC types using the existing names.  That means this change breaks the build of FM, which is a problem.  To get around that we would have to provide both the old and new names (which can be `#define`'d) and follow the deprecation procedure.
2. We should probably back out the "TypeCRC" local variable in the test cases, too.  Reason is that it makes cross referencing the test cases more difficult.  That is, in the log it would previously show the call along with the symbolic name of the last parameter, and we can indicate in our test reports which test case(s) covered those functions.  They are still being covered, of course, but the log now just shows "TypeCRC" for all the calls, so its not clear which call is covering which option, which complicates the reporting/verification that we need to do.

I might be able to update both of those things myself, because this is something I would like to get merged for other reasons.

thnkslprpt:
@jphickey

Thanks Joe. Now I understand when deprecation is required.

I had a go at the changes you described.
Happy to leave these changes to you though if it will be too much back-and-forth. I know you guys are (very) busy. Just let me know mate.
Cheers

jphickey (approved):
Perfect, thanks for doing all those changes.  This looks good to me now, will review with the CFS CCB.

thnkslprpt:
OK no worries @jphickey.
Squashed the commits and added a couple touch-ups (synced up the function declarations).

PR #2263, opened 2023-03-23
Fix #2262, separate dispatcher for messages

### LINKED ISSUE #2262: Move cmd pipe processing into separate dispatch file
Currently the "TaskPipe" function is comprised of a couple big switch statements based on MsgId and FcnCode, along with length validation and potentially other structural checks.   If/When migrating to EDS some of that logic can be handled by common code that check against constraints defined in EDS.

The current "TaskPipe" function and the supporting logic around it should be moved to a separate "dispatch" source unit.  This improves organization in general, but also makes for an easier switch to different logic in the future, should that come to be needed.

The alternative is to add `#ifdef` blocks if validation logic needs to be switched.  

**Additional context**
The recommended solution of moving this logic to a separate source unit permits source-selection in CMake at configuration time rather than relying on preprocessor-based option selection.  This is generally cleaner, and permits all options to be verified in a single unit test build, as opposed to only being able to test the selected option.

**Requester Info**

jphickey:

**Describe the contribution**
Isolate the message verification and dispatch from the general message processing.  Functions in the "task" source file now strictly handle the command content, and do not get involved in general validation.

Also note: this is mostly a simple code move but does add a `const` qualifier to the task pipe function where it was missing before.  All of the handlers were already `const` so this just makes it consistent throughout the handling.

Fixes #2262

**Testing performed**
Build and run CFE and sanity check.  Run all tests

**Expected behavior changes**
None - this just moves code from one source file into another, does not make any change to functionality.

**System(s) tested on**
Debian


jphickey:
Fixed the clang-format issue, this should be good to go now.

Issue #94, opened 2019-09-30
EVS "output ports" should be a function of the PSP


skliper:
Currently the EVS has 4 "output ports" implemented as functions within EVS: {{{EVS_OutputPort1(), EVS_OutputPort2(), EVS_OutputPort3(), EVS_OutputPort4()}}}

These are all **identically hard coded** to {{{OS_printf()}}}

The premise here seems logical -- to have several different destinations that the message may be sent.  But the current implementation does not allow for that to really happen.

The proposal is to change this to a PSP implementation:

{{{CFE_PSP_SendEventToPort(uint32 PortNum, const char *Message);}}}

This single API could be used for all ports, and the PSP could switch based on the "PortNum" value if needed, or simply call {{{OS_printf}}} for all messages as it does in the current code.

skliper:
_Imported from trac issue 63.  Created by jphickey on 2015-06-04T12:43:35, last modified: 2019-07-03T12:58:32_

skliper:
_Trac comment by jphickey on 2015-06-04 12:45:32:_

Candidate for CCB discussion at some point.

skliper:
_Trac comment by jphickey on 2018-05-14 13:39:47:_

I would like to see this revisited for the next CFE release.  The event "output ports" feature has been sitting in its current state for quite some time, where it is there but not really useful in any meaningful way.

We should either allow deployments to actually make use of dynamic output possibilities by directing these through the PSP, or decide that it will only ever be `printf` and therefore remove this extra logic.

(In particular, on Linux systems it might be quite useful to direct these events to the syslog daemon.  Leveraging the PSP could allow that to happen).

skliper:
_Trac comment by jhageman on 2019-04-03 17:47:39:_

CCB 4/3/19 Discussion
 - PSP SendMessageToPort to take a char string and send out port.  Services and apps can format the output however they want (EVS could format events however desired).
 - Clean up EVS to just format and send to PSP.

skliper:
_Trac comment by jhageman on 2019-04-03 17:48:13:_

See also #93, coordinated solution suggested.

skliper:
_Trac comment by jhageman on 2019-07-03 12:58:32:_

Moved unfinished 6.7.0 tickets to next release

skliper:
Related requirement cEVS3017 and cEVS3109:
[...]

PR #2295, opened 2023-04-17
Fix #219, #93, Add EVS port timestamp and simplify port selection

### LINKED ISSUE #93: Clean up EVS ports implementation
Fix a few minor issues with EVS ports implementation:

1. Although a macro is used when checking bits of the "OutputPort" mask, the value is still effectively hard coded with the shift, so the macro cannot change without also breaking these checks (violates the spirit of using the macro to begin with).  A simplification of the conditional will allow them to change independently (just mask and non-zero):
https://github.com/nasa/cFE/blob/73c338d544f85cb0c1a97b743d3ae9f7a35039b5/modules/evs/fsw/src/cfe_evs_task.c#L831-L847
https://github.com/nasa/cFE/blob/73c338d544f85cb0c1a97b743d3ae9f7a35039b5/modules/evs/fsw/src/cfe_evs_task.c#L885-L901
https://github.com/nasa/cFE/blob/73c338d544f85cb0c1a97b743d3ae9f7a35039b5/modules/evs/fsw/src/cfe_evs_utils.c#L469-L502

2. When sending to multiple ports, the (essentially) same snprintf() is done for each one and only one digit in the string changes.  Would be way more efficient to construct the later half of the string once, and change the port number.

The first issue should be resolved prior to integration with EDS as those macros become part of the EDS.  Therefore they should be changeable without breaking the implementation that uses them.  The second issue is just a performance/size enhancement.

### LINKED ISSUE #219: add timestamps to EVS logging to stdout
Often it's useful to know the time a message is generated in the stdout log. Adding a timestamp to EVS_SendViaPorts() as an option would be helpful.

skliper:

**Describe the contribution**
- Fix #93
- Fix #219

**Testing performed**
CI and on local build with time stamp enabled

**Expected behavior changes**
Adds time stamp when configured to do so

**System(s) tested on**
CI, Ubuntu 20.04

**Additional context**
Related but doesn't address (although should make the change simpler):
- #94

**Third party code**
None

Jacob Hageman - NASA/GSFC


skliper:
Only failure is the overly restrictive commit message format check.  Should there be a standard for commits that fix two issies?

CDKnightNASA:
A couple comments:

1) YYYY/MM/DD HH:MM:SS.MMM adds a lot of bytes to the log, it may make sense to just use HH:MM:SS.MMM or even have a "epoch" at the top of the log then all reports are SS.MMM from T0.
2) The "ts" command for Unix adds timestamps to streams (surely not microsecond-accurate.)
3) Instead of compile-time optional, would it be helpful to have it be command-able?

skliper:
> 1. YYYY/MM/DD HH:MM:SS.MMM adds a lot of bytes to the log, it may make sense to just use HH:MM:SS.MMM or even have a "epoch" at the top of the log then all reports are SS.MMM from T0.

I only need the format added.  If someone requires additional formats they could address in a future issue.

> 2. The "ts" command for Unix adds timestamps to streams (surely not microsecond-accurate.)

I don't think this would help my specific need, wouldn't correlate to syslog timing or packet timestamps.

> 3. Instead of compile-time optional, would it be helpful to have it be command-able?

I only need it via config.  If someone requires commandability they could address in a future issue.

Good suggestions, but given my current requirements this change is all I need.  Hopefully the future work can be independent from this ?

skliper:
I also want to get this week's award for addressing the oldest issue (2017!).  So there's that :)

skliper:
@dzbaker @dmknutsen - Is there a CCB action on this one or is it pending review?

dzbaker:
> @dzbaker @dmknutsen - Is there a CCB action on this one or is it pending review?

@jphickey is just giving it a quick review to make sure there are no conflicts with other stuff being developed. It should be merged in at the end of this week.

dmknutsen (approved):

jphickey (commented):
A general comment, in CFE/CFS we have _lots_ of configuration options - configuration option overload, possibly - to the point that we cannot feasibly test them all.

I'm trying to think why we can't just add the time here, and make it "always on"?  Is there a use-case where this would break someone's environment somehow?  I don't think anyone would rely on _not_ having a timestamp in event text.  If we can avoid having yet-another-option that we can't test/maintain both ways, I'd rather just make this the new standard.

skliper:
> ... just add the time ...

I'll do it!  Much preferred from my perspective.

dzbaker:
@jphickey @skliper @dmknutsen I can fast track this and include it in this next bundle if that makes sense?

jphickey (approved):
I like it....

jphickey:
That being said, I dunno if this should close #93 ... The intent with that old idea is that EVS should call into the PSP for low level output.  This is where you can make the 4 "output ports" go to different physical devices (depending on what the board has) and also attach different info per-port.  That was the idea at least.  Having 4 "output ports" in EVS without an mapping to the lower-level hardware seems to be an incomplete implementation of this, its been that way for years though.

skliper:
This is the one to move to the PSP (at least how I interpret them):
- #94 

93 just cleans up how the bits are checked and paves the way for 94:
- #93

jphickey:
> This is the one to move to the PSP (at least how I interpret them):
> 
>     * [EVS "output ports" should be a function of the PSP #94](https://github.com/nasa/cFE/issues/94)
> 
> 
> 93 just cleans up how the bits are checked and paves the way for 94:
> 
>     * [Clean up EVS ports implementation #93](https://github.com/nasa/cFE/issues/93)

Thanks, yes this makes sense.  I'm OK with this closing #93.  Maybe someday will get to implement #94 too, I'd still like to see that done.

PR #2312, opened 2023-04-27
Fix #2311, correct fallback file case

### LINKED ISSUE #2311: New cfe_locate_implementation_file() includes fallback file in results even if override found
**Describe the bug**
When using a "FALLBACK_FILE" option, this is put at the last priority
When using the "ALLOW_LIST" option, this function returns ALL files found that match the pattern being searched for.   

When those two options are used together, it means that the fallback file will be included in the result list, along with the user-supplied files.  This is not the intent; the fallback file should only be returned if _no_ user-supplied files were found.

The result of including both is that one gets duplicate defs and/or unexpected values.

The "generate_include_configfile()" uses it in this way, and thus gets both copies.

**To Reproduce**
Use "generate_config_includefile" with a file that has been overridden by the user.  The generated include file will get both the default and the user-supplied copy.

**Expected behavior**
Only the user-supplied copy should be used.

**Reporter Info**

jphickey:

**Describe the contribution**
Only include the "FALLBACK_FILE" if the normal search came up empty. Do not return a list containing the fallback/default along with the user-supplied files, only return the user-supplied files.  This was an issue when using "ALLOW_LIST" in that it had both.

Fixes #2311

**Testing performed**
Build with normal/default config, and build with overrides

**Expected behavior changes**
Overrides should work correctly again.

**System(s) tested on**
Debian

### DIFF
--- a/cmake/global_functions.cmake
+++ b/cmake/global_functions.cmake
@@ -28,14 +28,17 @@ function(cfe_locate_implementation_file OUTPUT_VAR FILE_NAME)
 
   cmake_parse_arguments(LOCATEIMPL_ARG "OPTIONAL;ALLOW_LIST" "FALLBACK_FILE" "PREFIX;SUBDIR" ${ARGN})
 
-  # Always check for filename matches directly in the MISSION_DEFS dir
-  set(IMPL_SEARCH_BASEDIRS "${MISSION_DEFS}/")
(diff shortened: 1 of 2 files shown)


jphickey:
Tested this fix with cert_testbed and confirmed that it now picks all the same header files that it used to.

PR #2319, opened 2023-05-02
Fix #2315, uniqueness of tbl target names

### LINKED ISSUE #2315: Uniqueness of intermediate table target names
Previously, each table ".c" source file was built into its own static library target based on that source file name.  This had the side effect of creating lots of complexly-named static lib targets.

With recent changes, this is changed to build all listed table source files listed into a single, simpler-named target.

Now a new problem occurs if one tries to build table files separately.  For example, something like this:

```
add_cfe_tables(sc rts10.c)
add_cfe_tables(sc rts11.c)
```

Will both try to register a static lib target named `tblobj_cpu1_sc`, which results in a target name collision error.

Invoking `add_cfe_tables` as above should work without a target name collision.

The `add_cfe_tables` does allow for a name suffix, which is used as part of the lib name - so as long as that suffix is unique, there is no name collision.  For example, this works:

```
add_cfe_tables(sc.rts10 rts10.c)
add_cfe_tables(sc.rts11 rts11.c)
```

And listing mutiple source files works too:

```
add_cfe_tables(sc.custom rts10.c rts11.c)
```

It is only the case of multiple invocations of `add_cfe_tables` with the _same_ first parameter that triggers this conflict.

**Additional context**
Unfortunately, we had previously documented/recommended the first parameter of `add_cfe_tables` to match the app name, so there may be more users that were following the problematic pattern above (1 is confirmed).

**Requester Info**

jphickey:

**Describe the contribution**
If the passed-in table name does not use a qualifier suffix, then generate one by computing a hash of the context information.  This ensures that the generated intermediate target names will be unique.

Fixes #2315

**Testing performed**
Replicated original error by doing additional "add_cfe_tables" for sample app in local project config, got duplicate name error from CMake.  Apply this patch, repeat build, and got all table files successfully.

**Expected behavior changes**
No more target name conflict

**System(s) tested on**
Debian


jphickey:
Latest update adds some more debug/status info messages when running cmake with "VERBOSE=1" flag set.  Ideally, nobody should need to reference the internal target names, but if they do, this flag sheds more light as to what is going on with the internal logic.

PR #2347, opened 2023-05-24
Fix #1143, SB_UT corrections and clear event count history after setup

### LINKED ISSUE #1143: Unit test should clear event count history after setup
Test cases that check for events should focus on the events actually sent by the unit under test.  But many test cases also require some "setup" calls, which also may generate events.

As a result the "assert" is checking for all events from the combination of (setup + test).

This makes tests more fragile, because many tests call the same routine e.g. `CFE_SB_AppInit()` as part of the test setup.  In the current implementation this appears to generate 5 events in its default/nominal mode.  But when (not if) `CFE_SB_AppInit()` changes in any way that affects its nominal events (very easy to do) it breaks nearly every test in SB.

For any call considered "setup" it should discard any events generated by the call, thereby not including them in the set that will be validated later as part of the test case.

**Additional context**
Note that all functions (by definition) have their own test case where events _will_ be verified.  It doesn't need to be verified again when it is used as the setup for something else.  Doing so just causes more maintenance work.

This also makes test cases clearer - as they should only be checking for events that are actually generated by the function under test.

**Requester Info**

thnkslprpt:
**Checklist**

**Describe the contribution**
- Fixes #1143
  - Clears event history after setup calls and updates EVENTCOUNT asserts to the lower number of events only expected from the test under question (rather than including the total events from the setup calls as well)

While looking at this, I realised that lots of tests were checking events related the 'setup', not the actual function that is the focus of the test.

For example, there are dozens of asserts for `CFE_SB_PIPE_ADDED_EID` and `CFE_SB_SUBSCRIPTION_RCVD_EID` - even though `CFE_SB_CreatePipe()` and `CFE_SB_Subscribe()` have their own tests...

These asserts are largely redundant and just add clutter to the tests - I've removed the obvious cases from the tests I was already updating with `UT_ClearEventHistory()`.

`Test_SB_Cmds_MapInfoDef` was checking for `CFE_SB_PIPE_ADDED_EID` and `CFE_SB_SUBSCRIPTION_RCVD_EID` which are part of the setup, not the function under test. I changed this to check for the command counter being incremented instead.

`Test_Unsubscribe_Basic`, `Test_Unsubscribe_AppId` and `Test_Unsubscribe_Local` were checking for `CFE_SB_SUBSCRIPTION_RCVD_EID`. They should really be checking for `CFE_SB_SUBSCRIPTION_REMOVED_EID` given that `CFE_SB_Unsubscribe()` is the focus of these tests.
  - In the case of `Test_Unsubscribe_Local`, it was actually sending a `CFE_SB_UNSUB_NO_SUBS_EID` event. The reason this wasn't noticed may be because `CFE_SB_UnsubscribeLocal()` returns `CFE_SUCCESS` (and sends the `CFE_SB_UNSUB_NO_SUBS_EID` event) if the pipe is not subscribed to the `MsgId`.
    - I've amended the test to subscribe using `CFE_SB_SubscribeLocal()`, instead of `CFE_SB_Subscribe()`, and now the call to `CFE_SB_UnsubscribeLocal()` in the test actually works and returns `CFE_SUCCESS` while sending the `CFE_SB_SUBSCRIPTION_REMOVED_EID` event.

Even though these are coverage tests, they should still be clear and at least check for what they are testing...

**Testing performed**
GitHub CI actions all passing successfully (incl. Build + Run, Unit/Functional Tests etc.).
Local testing confirms net coverage unchanged.

**Expected behavior changes**
No change to code behavior. Test behavior updated as described above.

**System(s) tested on**
Debian GNU/Linux 11 (bullseye)
Current main branch of cFS bundle.

**Contributor Info**
Avi Weiss @thnkslprpt

### DIFF
--- a/modules/sb/ut-coverage/sb_UT.c
+++ b/modules/sb/ut-coverage/sb_UT.c
@@ -570,6 +570,7 @@ void Test_SB_Cmds_Stats(void)
     CFE_UtAssert_SETUP(CFE_SB_CreatePipe(&PipeId1, 4, "TestPipe1"));
     CFE_UtAssert_SETUP(CFE_SB_CreatePipe(&PipeId2, 4, "TestPipe2"));
     CFE_UtAssert_SETUP(CFE_SB_CreatePipe(&PipeId3, 4, "TestPipe3"));
+    UT_ClearEventHistory(); /* Clear events generated during test setup */
 
     /* Generic command processing - The dispatch must be set up FIRST */
     UT_SetupBasicMsgDispatch(&UT_TPID_CFE_SB_CMD_SEND_SB_STATS_CC, sizeof(SendSbStats.Cmd), false);
@@ -583,7 +584,7 @@ void Test_SB_Cmds_Stats(void)
     CFE_SB_ProcessCmdPipePkt(&SendSbStats.SBBuf);
 
     /* No subs event and command processing event */
-    CFE_UtAssert_EVENTCOUNT(5);
+    CFE_UtAssert_EVENTCOUNT(2);
 
     CFE_UtAssert_EVENTSENT(CFE_SB_SND_STATS_EID);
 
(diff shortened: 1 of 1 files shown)


skliper:
> Even though these are coverage tests, they should still be clear and at least check for what they are testing...

Love it, great catch on the UT bug in `Test_Unsubscribe_Local`!

thnkslprpt:
> Love it, great catch on the UT bug in `Test_Unsubscribe_Local`!

Cheers Jake.
Yeah `SubscribeLocal()`/`UnsubscribeLocal()` are fully tested with functional tests, but still best to keep the coverage tests correct and up-to-date anyway of course.

thnkslprpt:
> I rebased this one and confirmed its working.

Thanks for updating it Joe.

PR #2367, opened 2023-06-01
Fix #2362, Add source routing APIs to SB

### LINKED ISSUE #2362: Add source-routing feature to Software Bus
Normally, software bus messages are fully assembled by clients, then passed to the software bus via e.g. `CFE_SB_TransmitMsg()` for delivery to subscribers.  This routing to subscribers is currently done based on the `MsgId` value that is present in the message - that is, the MsgId is looked up in the routing table, which is in turn translated to a list of destinations (subscribers) to deliver that message to.

Add an alternative API that allows the message to be routed to destination(s) that are are given explicitly in the API call.  That is, allow the caller to specify the `MsgId` that SB should use to route and deliver the message.  Specifically - this passed-in `MsgId` for routing may be different than the MsgId value contained in the message.

N/A

**Additional context**
The use-case for this feature has to do with complex systems with distributed software bus services across many instances of CFE.  In this context the destination may not be directly reachable from the source, but reachable through some sort of intermediate hop.  This feature gives the needed flexibility to work with such an architecture, by allowing messages to be routed to an intermediate delivery assistant app that may not be the final destination of the message.

**Requester Info**

jphickey:

**Describe the contribution**
Add two new APIs that allow more control over the message routing. These allow the caller to directly specify the route (MsgId) and size of the content, and the message will be delivered based on the passed in values vs. the values inside the message itself.

This restructures the existing/historical API calls to use the same underlying mechanism.  Send/Receive actions have a transaction object associated and this tracks the status and events.  Common event reporting is done based on this transaction object.

Fixes #2362

**Testing performed**
Build and run CFE including functional test app
Build and run sanity tests

**Expected behavior changes**
Adds two new APIs.  Existing APIs should be backward compatible, with some possible changes to the expected event IDs that might be generated under some off-nominal conditions.

**System(s) tested on**
Debian

**Additional context**
The new API allows the passed object to be any arbitrary data -- it does not need to be a CFE/CCSDS message at all.  Therefore this could be more suitable for things like CFDP PDU packets, etc. as it would not need to be forced to look like a CMD/TLM message.

### DIFF
--- a/modules/sb/ut-coverage/sb_UT.c
+++ b/modules/sb/ut-coverage/sb_UT.c
@@ -3515,9 +3515,48 @@ void Test_TransmitMsg_API(void)
     SB_UT_ADD_SUBTEST(Test_TransmitTxn_FindDestinations);
     SB_UT_ADD_SUBTEST(Test_TransmitTxn_PipeHandler);
     SB_UT_ADD_SUBTEST(Test_TransmitTxn_Execute);
+    SB_UT_ADD_SUBTEST(Test_TransmitBufferWithRoute);
 
     SB_UT_ADD_SUBTEST(Test_AllocateMessageBuffer);
     SB_UT_ADD_SUBTEST(Test_ReleaseMessageBuffer);
+}
+
+void Test_TransmitBufferWithRoute(void)
+{
+    /* Test case for:
+     * CFE_Status_t CFE_SB_TransmitBufferWithRoute(CFE_SB_Buffer_t *BufPtr, CFE_MSG_Size_t BufSize, CFE_SB_MsgId_t
+     * RoutingMsgId, bool EnableUpdate, int32 TimeOut);
+     */
+
+    CFE_SB_Buffer_t       *SendPtr;
+    const CFE_SB_Buffer_t *ReceivePtr = NULL;
+    CFE_SB_PipeId_t        PipeId     = CFE_SB_INVALID_PIPE;
+    CFE_SB_MsgId_t         MsgId      = SB_UT_TLM_MID;
+    CFE_SB_MsgId_t         RecvMsgId  = CFE_SB_INVALID_MSG_ID;
+    size_t                 RecvSize;
+
+    static const uint8_t RandomBytes[] = { 0x4a, 0x75, 0x62, 0x34, 0x1d, 0xa5, 0x48, 0x5b, 0xc6, 0xb4, 0x40, 0x3f };
+
+    CFE_UtAssert_SETUP(CFE_SB_CreatePipe(&PipeId, 3, "TestPipe"));
+    CFE_UtAssert_SETUP(CFE_SB_Subscribe(MsgId, PipeId));
+
+    UtAssert_NOT_NULL(SendPtr = CFE_SB_AllocateMessageBuffer(sizeof(SB_UT_Test_Tlm_t)));
+    memcpy(SendPtr, RandomBytes, sizeof(RandomBytes));
+
+    /* Test a successful pass */
+    CFE_UtAssert_SUCCESS(CFE_SB_TransmitBufferWithRoute(SendPtr, sizeof(RandomBytes), MsgId, false, CFE_SB_POLL));
+    CFE_UtAssert_SUCCESS(CFE_SB_ReceiveBufferWithRoute(PipeId, &ReceivePtr, &RecvSize, &RecvMsgId, false, CFE_SB_POLL));
+
+    UtAssert_ADDRESS_EQ(SendPtr, ReceivePtr);
+    UtAssert_EQ(size_t, RecvSize, sizeof(RandomBytes));
+    CFE_UtAssert_MSGID_EQ(RecvMsgId, MsgId);
+
+    /* Test bad input */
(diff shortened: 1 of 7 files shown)


jphickey:
Still have some (known) coverage and test workflow issues to solve in here, but wanted to get this going for CCB review today.

skliper:
I don't understand the driver behind adding complexity to SB to handle blobs of data that aren't self identifiable, this doesn't seem like a "natural" fit for pub/sub.  Why not just wrap it with CCSDS if pub/sub is necessary, or implement point-to-point if it isn't?  Simple tools for simple jobs vs trying to expand SB to do everything.  Is this introducing a pattern of extending a core service beyond original intent, and if so what are the bounds?  There's no longer embedded size, identification, common handling supported by binary blob messages, it's not clear why such an extended change makes sense in the open source context.  Maybe just add a library that does wrapping/unwrapping for you and can be used by anyone who needs to send blobs, or some other way to avoid such major rewrites?  My concern is now receivers need to know if it's a blob or not, which none of the standard receivers handle, so as soon as this is introduced there's an interoperability issue all over (not every subscribed MID can be checked for size or whatever).

jphickey:
Yeah, arbitrary data blobs are not really the emphasis/driving need behind this.  Its more about the routing flexibility. 

We already had the notion of the message itself as well as its "envelope" -- the `CFE_SB_BuffertD_t` that wrapped every message as it traversed the bus.  The latter was used for routing (mostly) but mainly for convenience, some code would still peek at the message.  This change makes the envelope address the authoritative one - that is, once a message gets put into its envelope, that envelope address becomes the one (and only) thing that is used for routing on the bus.

The fact that this permitted the message to be a binary blob was a side benefit - if nothing else, proving that delivery of a random binary blob works ensures that the delivery is being done based only on its envelope information.

This opens up a whole bunch of possibilities for future features:

 - Messages can be sent through intermediate delivery agent(s) - in case the dest is not directly reachable on the same SB subscription domain - they can go through one or more such hops if needed, on the way to their internal destination.  **This is the current need of the stakeholders and the driving reason behind this PR**
 - Envelope size could be bigger than the internal message size, with the extra space for things like a CRC and/or cryptographic signature for authentication.  This is also currently needed by the stakeholders, although there are other was of doing this that have been used, it is more complex and have been a big maintenance issue.  This would simplify.
 - Envelope content could be encrypted (might not be useful till we get isolated memory spaces, but still...).  Although this is not currently needed by the stakeholders right now, its been on the roadmap for some time.

If the concern has to do with the (possibility) of sending a binary blob and confusing a subscriber, we could conceivably restrict content to "Valid messages" even with this new API, and add some checks that the message attempting to be sent seems like an actual message.  I think all we could do is confirm that the `CFE_MSG_GetSize()` and `CFE_MSG_GetMsgId()` return reasonable values, which is really the only checks the currently calls do (that is, today one could generate a message that is complete garbage but as long as these two calls output values that are within range, the blob can be sent).   That isn't a particularly robust test of a message sanity, but its all we can do, and all we ever really did up to this point even with the current APIs.

skliper:
While I get you could do these things all within SB using the SB internal "envelope" concept, I don't agree with such a temporary nature or adding complexity to SB to do things that also could require changes to every sender/receiver.  This concept seems to tie directly with SB, where it's dependent on these new APIs, only SB has the envelope information, it's gone as soon as delivered, etc.  If the envelope is part of the message (wrapped) that dependency is eliminated, the CCSDS spacepacket header is already an envelope and you can layer as needed to add all the functionality above yet maintain independences from SB "envelope" magic.  Wrapping w/ additional protocols (or encapsulating protocols in CCSDS space packets) has been common practice and eliminates the strict dependency on SB to get this all done.  You could keep the SB/MSG independence and just update MSG to handle whatever envelope/wrapper information is needed and it can be customized/localized to the specific projects that need it.  If folks really want this to work seamlessly across the common apps I just don't see how adding APIs and requiring special handling on the receiving side is the preferred approach.

modules/sb/ut-coverage/sb_UT.c
```diff
@@ -4160,6 +4199,89 @@ void UT_CFE_MSG_Verify_CustomHandler(void *UserObj, UT_EntryKey_t FuncKey, const
...
+    /* Test case for:
+     * CFE_Status_t CFE_SB_ReceiveBufferWithRoute(CFE_SB_Buffer_t **BufPtr, CFE_MSG_Size_t *BufSize, CFE_SB_PipeId_t
+     * RxRoute, bool EnableVerify, int32 TimeOut);
+     */
+
+    CFE_SB_BufferD_t       BufDsc[2];
+    const void            *BufDscP[2];
+    CFE_SB_PipeId_t        PipeId;
+    CFE_SB_PipeD_t        *PipeDscPtr;
+    // CFE_SB_MsgId_t   TestMsgId  = SB_UT_TLM_MID;
```
chillfig: Can we omit this line?

chillfig (dismissed):
Looks good. 1 nit with commented test code. Issuing approval

modules/sb/ut-coverage/sb_UT.c
jphickey: Fixed in c190bfc5

chillfig (approved):

PR #2404, opened 2023-07-17
Revert "Fix #2395, do not directly use cfe_test_msgids.h"

### LINKED ISSUE #2395: Testcase code should not directly include "cfe_test_msgids.h" (yet)
**Describe the bug**
The module-specific header file is a relatively new addition.  But for a user that has overridden the previous all-inclusive `cfe_msgids.h` file, including the test_msgids.h directly will cause conflict with the overrides.

**To Reproduce**
Override (only) `cfe_msgids.h` in a build, and change the msgids in here.  Do not override cfe_test_msgids.h.
Attempt to build cfe_testcase -- the `cfe_test_msgids.h` inclusion will bypass the modified msgids, and get defaults, which will conflict / mismatch.

**Expected behavior**
Should be backward compatible with an existing override of `cfe_msgids.h`

**System observed on:**
Debian

**Reporter Info**

jphickey:
Reverts nasa/cFE#2399

### DIFF
--- a/modules/cfe_testcase/src/sb_sendrecv_test.c
+++ b/modules/cfe_testcase/src/sb_sendrecv_test.c
@@ -29,6 +29,7 @@
 
 #include "cfe_test.h"
 #include "cfe_msgids.h"
+#include "cfe_test_msgids.h"
 
 #define CFE_FT_STRINGBUF_SIZE 12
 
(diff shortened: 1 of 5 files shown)


skliper:
@jphickey - why is this being reverted?  This fixed an issue for us.

jphickey:
Turns out, this only works if you have `cfe_testcase` included in the application set.  If you don't build the testcase app, it breaks the build ..... see https://github.com/nasa/cFS/actions/runs/5579990936/jobs/10196292874

PR #2424, opened 2023-08-10
Fix #2423, avoid use of abspath make function

### LINKED ISSUE #2423: Older versions of make do not support "abspath"
**Describe the bug**
The table build procedure uses the GNU make function `$(abspath <file>)` as documented here: https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html#index-abspath-1

However, older versions of make (e.g. 3.8) appear to not support this function as described - although make itself does not throw an error about this function, the result is an empty string.  This in turn causes the command being executed to fail because its missing the argument, which is supposed to be a file name.

**To Reproduce**
Build the software using an older version of GNU make, an error such as this occurs:

```
cd elf/lx_es1_stub/ && /usr/bin/ar x  cfe_test_tbl.c.o
/usr/bin/ar: cfe_test_tbl.c.o: No such file or directory
```

**Expected behavior**
There should be a library filename between the "x" and the object filename.

**Code snips**
The command in the makefile here is: https://github.com/nasa/cFE/blob/8852743153ddafcc7b12054f59bc86a361480f99/cmake/tables/elf2cfetbl_rules.mk#L5

Note that the `$(abspath $(<))` evaluated to the empty string.  Whereas on GNU make 4.x this works as described in the documentation.

**System observed on:**
VxWorks 6.9, which ships a binary version of GNU make v3.8.

**Reporter Info**

jphickey:

**Describe the contribution**
The abspath function does not work in GNU make v3.80, which is still used as it is packaged with the Wind River tools as part of VxWorks 6.x.

The `$(abspath ...)` function was used in two places related to generating tables - first for extraction of the ELF file from an intermediate lib, second for conversion of that ELF file to a table file.  In the first case, the path should already be absolute, as it comes from CMake. In the second case, the path should always be relative, because it is intended to match another pattern rule starting with "elf/%".  For the second rule, prefixing with $(CURDIR) should achieve the same effect.

Fixes #2423

**Testing performed**
Build with both older (v3.80) and newer (v4.3) versions of GNU make

**Expected behavior changes**
Successfully builds using make v3.80

**System(s) tested on**
GSFC Vxworks development machine


jphickey:
Note this PR also adds some quoting.  This should make it a bit more robust just in case a filename contains spaces.  But more importantly, it makes it more evident in the console if an argument evaluated to an empty string -- that is, one can easily see the quotes with an empty string vs. no argument at all - which doesn't really look wrong.

PR #2489, opened 2023-12-19
Fix #2488, separate bad argument test

### LINKED ISSUE #2488: MSG testcase passes in two bad arguments at once
**Describe the bug**
The `CFE_MSG_Init()` test is passing in both an invalid pointer as well as an invalid MsgID into the same invocation.  While this should return an error code either way, its not clear which bad argument is producing the error.

**To Reproduce**
N/A

**Expected behavior**
Should pass in only one "bad" argument at a time.

**Code snips**
The test case in this line:
https://github.com/nasa/cFE/blob/7f5ebcd15032fe31e7ef0e5ff4ef624d51d66e16/modules/cfe_testcase/src/msg_api_test.c#L52

**System observed on:**
Debian

**Additional context**
Passing in only one bad value at a time ensures that each one is checked.

**Reporter Info**

jphickey:

**Describe the contribution**
Pass in CFE_SB_INVALID_MSG_ID separately from a NULL pointer. This also required updates to the way the MSG module checks the MsgID, as it was not using the IsValidMsgID check.

Fixes #2488

**Testing performed**
Build and run all tests

**Expected behavior changes**
CFE_MSG_Init() will now return the BAD_ARGUMENT code if passed CFE_SB_INVALID_MSG_ID.  Oddly, this was **not** the case previously, it would still return CFE_SUCCESS.  This is because the MSG implementation wasn't using the same check function (CFE_SB_IsValidMsgID).  This PR corrects that.

**System(s) tested on**
Debian

**Additional context**
This is correct behavior now, but its one of those things where I can't be totally sure if some app was relying on the incorrect behavior.

### DIFF
--- a/modules/cfe_testcase/src/msg_api_test.c
+++ b/modules/cfe_testcase/src/msg_api_test.c
@@ -26,6 +26,7 @@
  */
 
 #include "cfe_test.h"
+#include "cfe_test_msgids.h"
 #include <string.h>
 
 void TestMsgApiBasic(void)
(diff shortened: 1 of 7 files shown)


jphickey:
@skliper  --- do you know why the original "SetMsgId()" only checked against HIGHEST_VALID_MSGID instead of using the check implemented in SB?  Was there a justification for having a different/lesser validation of MsgId in this particular code?

I'm curious because I don't see any real reason it should be more permissive, yet there seemed to be coverage tests checking for the fact that a msgID value of "0" can be successfully set.

skliper:
I think that's just left over from when `CFE_SB_INVALID_MSG_ID` was not zero.  I think at some point in history it was in a config file.  With it now in the core_api the current tests to check for `0` being valid don't make sense to me.

skliper (approved):

PR #2517, opened 2024-02-14
Fix #2516, propagate stack pointer for child tasks

### LINKED ISSUE #2516: CFE_ES_CreateChildTask ignores its StackPtr argument
**Describe the bug**
The implementation of `CFE_ES_CreateChildTask()` does not correctly pass through the stack pointer, if this is used.  As a result, the child task always uses a dynamically allocated stack, not the passed-in pointer.

**To Reproduce**
Call `CFE_ES_CreateChildTask()` with a non-NULL stack pointer.  
The `CFE_ES_StartAppTask()` underlying function always uses `OSAL_TASK_STACK_ALLOCATE`, not the passed-in pointer.

**Expected behavior**
Should use the stack pointer provided by the user, if they pass one in.

**System observed on:**
Debian

**Reporter Info**
Joseph Hickey, Vantage Systems, Inc. (reporting on behalf of user/stakeholder bug report)

jphickey:

**Describe the contribution**
Update CFE_ES_CreateChildTask to propagate the user-supplied stack pointer to the underlying OS_TaskCreate call.  Also adds a functional test to check that the memory address of a local variable within a child task resides within the expected stack buffer.

NOTE: this requires an additional fix to POSIX OSAL to make it work on that platform.

Fixes #2516

**Testing performed**
Build and run all tests

**Expected behavior changes**
CFE Child tasks created will adhere to passed-in stack pointer.

**System(s) tested on**
Debian, RTEMS

**Additional context**
The newly added functional test will fail on POSIX due to a similar bug in the OS_TaskCreate() implementation.  This didn't propagate the stack pointer, either.

The biggest risk with "fixing" this is that it might expose stack size problems that were otherwise hidden.  Many apps use very small stack sizes for child tasks, and on POSIX (at least Linux/Glibc) it puts TLS storage in the stack buffer.  This TLS takes about 10kB off the top of the stack.  If the entire stack was only 4kB, this is now a problem.  It was hidden because POSIX created a stack anyway, and when it created the stack it was at least 16kB.  But by fixing that problem, the stack size really will be only 4kB, and thus not be big enough, and memory corruption will result.

### DIFF
--- a/modules/cfe_testcase/src/es_task_test.c
+++ b/modules/cfe_testcase/src/es_task_test.c
@@ -90,6 +90,66 @@ void TaskExitFunction(void)
     }
 }
 
+#define UT_LOCAL_STACK_SIZE 4096
+static unsigned long UT_LOCAL_STACK[UT_LOCAL_STACK_SIZE];
+
+void TestCheckStackPointer(void)
+{
+    int32   LocalVar;
+    cpuaddr VarAddress;
+    cpuaddr StackAddress;
+
+    OS_TaskDelay(10);
+
+    VarAddress   = (cpuaddr)&LocalVar;
+    StackAddress = (cpuaddr)UT_LOCAL_STACK;
+
+    UtAssert_GT(cpuaddr, VarAddress, StackAddress);
+    UtAssert_LT(cpuaddr, VarAddress, StackAddress + sizeof(UT_LOCAL_STACK));
+
+    CFE_ES_ExitChildTask();
+}
+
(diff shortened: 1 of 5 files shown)


jphickey:
Related OSAL issue is fixed in nasa/osal#1450 (dependency)

jphickey:
Checked the failure reported in the workflows -- confirmed that it is the newly added test, and it should be fixed by the linked OSAL PR.

jphickey:
Pushed the same content in new commit, in order to re-run workflows.

PR #2539, opened 2024-03-28
Fix #2538, table transaction initial implementation

### LINKED ISSUE #2538: Introduce transaction object for CFE TBL
Currently the table services implementation has a number of likely race conditions.  Most API calls do not lock the registry - generally only handle registration does so, and only for the specific part of the operation that finds the empty slot.

This is not an easy fix, as most functions are complex and scattered with inline "return" calls.

Introduce the concept of a transaction.  A similar concept is used in OSAL and other CFS apps. 

A transaction will track the overall state of the change.  Instead of independently tracking the Handle, `DescPtr` and `RegRegPtr` in the code, the transaction object wraps it all into one struct.  It can also track what _was_ changed, such that it can be reliably un-done if something later in the process did not work.  For example, allocating a new table requires both a registry entry and an access descriptor.  If the registry entry is successfully allocated first, and then the access descriptor fails to allocate, then the registry entry must be released to put the global tables back into their initial state.

Importantly, if a global object is locked during a transaction, it must be unlocked at the end of the transaction.  The transaction object can track this state as well.

**Additional context**
This would be just one step toward fixing race conditions.  In order to properly fix these, logging and event generation will likely need to be moved to the end of the transaction (after unlock).  Importantly, we can't just extend locks across the entire operation, if the operation involves calls into other subsystems to write to syslog or send an event.  But this will likely help reveal where the problem areas are and make it easier to fix.

**Requester Info**

jphickey:

**Describe the contribution**
Initial implementation of transaction concept for table API calls. The basic transaction object wraps all of the relevant pointers and state information for the request being processed.

Fixes #2538

**Testing performed**
Build and run all tests

**Expected behavior changes**
Internal code structure more manageable.  
Behavior of external/public API calls should remain the same.

**System(s) tested on**
Debian

**Additional context**
Does not really "fix" the problem of race conditions alone, needs more work to actually move/reorg the problem areas.  But this doesn't break anything that wasn't already broken (i.e. race conditions in here were there before).  They are just more evident (and fixable) now.

Unit and functional tests (mostly) did not need to change for this, aside from a few minor details.

### DIFF
--- a/modules/tbl/CMakeLists.txt
+++ b/modules/tbl/CMakeLists.txt
@@ -10,6 +10,8 @@ project(CFE_TBL C)
 set(tbl_SOURCES
     fsw/src/cfe_tbl_api.c
     fsw/src/cfe_tbl_internal.c
+    fsw/src/cfe_tbl_resource.c
+    fsw/src/cfe_tbl_transaction.c
     fsw/src/cfe_tbl_task.c
     fsw/src/cfe_tbl_task_cmds.c
     fsw/src/cfe_tbl_dispatch.c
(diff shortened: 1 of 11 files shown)


jphickey:
Documentation updates should be resolved now, pending successful workflow run.

Issue #954, opened 2020-10-16
EVS coverage test fails when CFE_PLATFORM_EVS_DEFAULT_TYPE_FLAG is set to 0xF


skliper:
**Describe the bug**
When CFE_PLATFORM_EVS_DEFAULT_TYPE_FLAG is set to 0xF there are two failures:
```
[ FAIL] 14.001 evs_UT.c:2675 - CFE_EVS_WriteLogDataFileCmd - Write log data - successful
[ FAIL] 14.002 evs_UT.c:2685 - CFE_EVS_SetLogModeCmd - Set logging mode - successful
```

**To Reproduce**
1. Set CFE_PLATFORM_EVS_DEFAULT_TYPE_FLAG to 0xF
2. Build unit tests, run cfe-core_evs_UT, observe failure

**Expected behavior**
Test should pass with debug events enabled (shouldn't care).

**Code snips**
https://github.com/nasa/cFE/blob/dc3d62b37d2bacd8fe23f66ff7b64025ac7d0751/fsw/cfe-core/unit-test/evs_UT.c#L2670-L2688

**System observed on:**
 - Hardware: cFS Dev Server
 - OS: Ubuntu 18.04
 - Versions: 6.8.0-rc1+dev129

**Additional context**
N/A

**Reporter Info**
Jacob Hageman - NASA/GSFC

skliper:
I haven't dug into the logic, but I suspect it's considering no events a success.  This doesn't seem very resilient, check should be that the dispatch logic was followed as expected not zero events.  If all events were disabled, this would always pass.

PR #2691, opened 2026-04-13
Fix #2692, add size check for each header field

### LINKED ISSUE #2692: [SECURITY] Integer overflow in CFE_TBL_ValidateCodecLoadSize bypasses bounds check on 32-bit targets
* [x] I performed a cursory search to see if the bug report is relevant, not redundant, nor in conflict with other tickets.

**Describe the bug**
Integer overflow in `CFE_TBL_ValidateCodecLoadSize()` bypasses the table load bounds check on 32-bit targets. The function computes `ProjectedSize = Offset + NumBytes` using `uint32` arithmetic. When the sum exceeds 2^32, the result wraps to a small value that passes the bounds check. The original unwrapped `Offset` is then used for the destination pointer computation in `CFE_TBL_LoadContentFromFile()`, enabling out-of-bounds memory writes on 32-bit flight processors (LEON3, RAD750).

This is mitigated on 64-bit targets where pointer arithmetic does not wrap at 2^32.

**To Reproduce**
1. Craft a table load file with `Offset=0xFFFFFF00` and `NumBytes=0x200` targeting a 256-byte table
2. The `uint32` addition wraps: `0xFFFFFF00 + 0x200 = 0x100` (256), which passes the `ProjectedSize > ActualSize` check
3. The subsequent `memcpy` uses the original `Offset` (0xFFFFFF00) for the destination pointer, writing 512 bytes 4GB beyond the table buffer
4. On 32-bit flight hardware, this wraps into other memory regions — confirmed with three overflow variants

Proof-of-concept source and output are available upon request.

**Expected behavior**
The bounds-check addition should be performed in 64-bit arithmetic to prevent wraparound, regardless of target architecture.

**Code snips**
`cfe/modules/tbl/fsw/src/cfe_tbl_passthru_codec.c` (line 163):
```c
uint32 ProjectedSize;
ProjectedSize = HeaderPtr->Offset + HeaderPtr->NumBytes;  // uint32 wraps at 2^32
if (ProjectedSize > ActualSize)
{
    /* reject as too large */
}
```

Example overflow:
```
Offset       = 0xFFFFFF00  (4,294,967,040)
NumBytes     = 0x00000200  (512)
True sum     = 0x100000100 (4,294,967,552)
uint32 wrap  = 0x00000100  (256)  <- passes check for TableSize=256
```

**System observed on:**
 - Hardware: x86-64 (arithmetic overflow demonstrated via `uint32_t`; exploitable OOB write requires 32-bit address space as on LEON3/RAD750)
 - OS: Linux (Ubuntu 22.04)
 - Versions: cFS Draco release, commit `83c735e`

**Additional context**
The primary cFS deployment targets are 32-bit flight processors (LEON3, RAD750) where this vulnerability is exploitable. The fix (casting to `uint64_t` before addition) is trivial and has zero impact on 64-bit targets. Requires ground operator command authority to issue table load commands.

jphickey:

**Describe the contribution**
Adds a size check on each of the table header size fields.  Previously only the sum (offset + numbytes) was checked against the size limit but if that sum overflows a uint32, then the check could pass when it should have failed.

Fixes nasa/cFE#2692

**Testing performed**
Build and run tests, confirm described case is now caught with error code (test added)

**Expected behavior changes**
Described condition is caught and error event sent

### DIFF
--- a/modules/tbl/fsw/src/cfe_tbl_eds_codec.c
+++ b/modules/tbl/fsw/src/cfe_tbl_eds_codec.c
@@ -421,7 +421,7 @@ CFE_Status_t CFE_TBL_ValidateCodecLoadSize(CFE_TBL_TxnState_t *Txn, const CFE_TB
     if (Status == CFE_SUCCESS)
     {
         ProjectedSize = HeaderPtr->Offset + HeaderPtr->NumBytes;
-        if (ProjectedSize > ActualSize)
+        if (ProjectedSize < HeaderPtr->Offset || ProjectedSize > ActualSize)
         {
             Status = CFE_TBL_ERR_FILE_TOO_LARGE;
             CFE_TBL_TxnAddEvent(Txn, CFE_TBL_LOAD_EXCEEDS_SIZE_ERR_EID, ProjectedSize, ActualSize);
--- a/modules/tbl/fsw/src/cfe_tbl_passthru_codec.c
+++ b/modules/tbl/fsw/src/cfe_tbl_passthru_codec.c
@@ -161,7 +161,7 @@ CFE_Status_t CFE_TBL_ValidateCodecLoadSize(CFE_TBL_TxnState_t *Txn, const CFE_TB
     /* Because the file sizes and in-memory size is the same, this check is simple */
     ActualSize    = CFE_TBL_RegRecGetSize(CFE_TBL_TxnRegRec(Txn));
     ProjectedSize = HeaderPtr->Offset + HeaderPtr->NumBytes;
-    if (ProjectedSize > ActualSize)
+    if (ProjectedSize < HeaderPtr->Offset || ProjectedSize > ActualSize)
     {
         Status = CFE_TBL_ERR_FILE_TOO_LARGE;
         CFE_TBL_TxnAddEvent(Txn, CFE_TBL_LOAD_EXCEEDS_SIZE_ERR_EID, ProjectedSize, ActualSize);
--- a/modules/tbl/ut-coverage/tbl_ut_support_eds.c
+++ b/modules/tbl/ut-coverage/tbl_ut_support_eds.c
@@ -357,8 +357,40 @@ void UT_TBL_ValidateCodecLoadSize_Test(void)
     UtAssert_ZERO(Txn.NumPendingEvents);
 
     Txn.RegRecPtr = &RegRec;
+
+    /* Case where the load exactly will fill the buffer */
+    RegRec.Config.Size = 100;
+    Header.NumBytes    = RegRec.Config.Size;
     UtAssert_INT32_EQ(CFE_TBL_ValidateCodecLoadSize(&Txn, &Header), CFE_SUCCESS);
     UtAssert_ZERO(Txn.NumPendingEvents);
+
+    /* Case where the numbytes is OK but offset is too large */
+    Header.NumBytes = RegRec.Config.Size / 2;
+    Header.Offset   = RegRec.Config.Size * 2;
+    UtAssert_INT32_EQ(CFE_TBL_ValidateCodecLoadSize(&Txn, &Header), CFE_TBL_ERR_FILE_TOO_LARGE);
+    UT_TBL_EVENT_PENDING(&Txn, CFE_TBL_LOAD_EXCEEDS_SIZE_ERR_EID);
+    CFE_TBL_TxnClearEvents(&Txn);
+
+    /* Case where the offset is OK but numbytes is too large */
+    Header.NumBytes = RegRec.Config.Size * 2;
+    Header.Offset   = RegRec.Config.Size / 2;
+    UtAssert_INT32_EQ(CFE_TBL_ValidateCodecLoadSize(&Txn, &Header), CFE_TBL_ERR_FILE_TOO_LARGE);
+    UT_TBL_EVENT_PENDING(&Txn, CFE_TBL_LOAD_EXCEEDS_SIZE_ERR_EID);
+    CFE_TBL_TxnClearEvents(&Txn);
+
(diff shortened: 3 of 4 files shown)


ivanperez-keera:
@jphickey  Why not compress all three conditions in one to avoid code duplication?

jphickey:
> @jphickey Why not compress all three conditions in one to avoid code duplication?

Because that would involve changing APIs, I didn't want to do that for this.   I was thinking about it.  Although the checks are similar the size comes from a different place between EDS and non-EDS.   To combine into a single impl would've necessitated a new/different API to get the value and it just seemed to not meet the size/benefit tradeoff to doing that.

ivanperez-keera:
Could you have the case where neither of the two values overflows a `uint32` and they are both smaller than `ActualSize` but the addition does overflow a `uint32` and is greater than `ActualSize` but, because of the overflow, it appears smaller, so you end up with `Status == CFE_SUCCESS`?

I just don't understand how this check protects you in case of an overflow.

Do you need to instead check if:

```C
HeaderPtr->Offset > ActualSize - HeaderPtr->NumBytes
```

as the third condition, which you know won't underflow because you already checked that `HeaderPtr->NumBytes` is not greater than `ActualSize`?

Or prefix the condition with:

```c
(UINT32_MAX - HeaderPtr->Offset < HeaderPtr->NumBytes) || ... rest of the condition ...
```

jphickey:
> Could you have the case where neither of the two values overflows a `uint32` and they are both smaller than `ActualSize` but the addition does overflow a `uint32` and is greater than `ActualSize` but, because of the overflow, it appears smaller, so you end up with `Status == CFE_SUCCESS`?
> 
> I just don't understand how this check protects you in case of an overflow.
> 
> Do you need to instead check if:
> 
> ```c
> HeaderPtr->Offset > ActualSize - HeaderPtr->NumBytes
> ```
> 
> as the third condition, which you know won't underflow because you already checked that `HeaderPtr->NumBytes` is not greater than `ActualSize`?
> 
> Or prefix the condition with:
> 
> ```c
> (UINT32_MAX - HeaderPtr->Offset < HeaderPtr->NumBytes) || ... rest of the condition ...
> ```

CFE Tables cannot be that large such that they are anywhere near able to overflow a uint32.  The max table size is 16k by default.  This can be raised, but we don't advertise supporting anywhere near the table sizes that would fit this hypothetical situation.

If we want to be really future-proof and paranoid, we can check that the sum / Projected Tail didn't come out to a value less than Header->Offset.  But -- this is bordering on "dead code" (against a different coding standard) because it simply cannot happen given the other constraints on tables (the fact that a valid table cannot be so large that this is possible).

If anything, I would just note in the configuration for the max table size that it cannot be greater than 2 GB (i.e. INT32_MAX)

ivanperez-keera (approved):
It would help to leave a comment so that future devs know the number cannot overflow because table sizes are far too small for that to happen.

Swapping the logic to protect against overflow could help make the code more obviously correct/safe without having to keep the fact that table sizes are much smaller in mind (meaning that even a static analyzer would probably not flag this as a potential overflow).

The commit message points to nasa/cFE#954 (which I believe is unrelated) instead of nasa/cFS#954 (which became #2692 in the cFE repo).

ivanperez-keera (approved):

modules/tbl/fsw/src/cfe_tbl_eds_codec.c
```diff
@@ -421,7 +421,7 @@ CFE_Status_t CFE_TBL_ValidateCodecLoadSize(CFE_TBL_TxnState_t *Txn, const CFE_TB
     if (Status == CFE_SUCCESS)
     {
         ProjectedSize = HeaderPtr->Offset + HeaderPtr->NumBytes;
-        if (ProjectedSize > ActualSize)
+        if (ProjectedSize < HeaderPtr->Offset || ProjectedSize > ActualSize)
```
ivanperez-keera, line 424: Just as a note, not a blocker:

Although this way of checking for an overflow is **not** unidiomatic, it might not hurt to leave a comment saying that we check for overflows. It's not obvious unless you are thinking you can overflow the `uint32` why you'd want to check the result of the sum against `HeaderPtr->Offset`.

Alternative that requires one more operation in the worse case but the intention is clearer:

```C
(HeaderPtr->NumBytes > (UINT32_MAX - HeaderPtr->Offset))
  || (ProjectedSize > ActualSize)
```

Alternative that requires the same operations in the worse case and one less in the best case:
```C
(HeaderPtr->Offset > ActualSize)
  || (HeaderPtr->NumBytes > (ActualSize - HeaderPtr->Offset))
```

This is probably a common enough case that we might want to consider having a macro that checks if addition of two numbers would overflow. That would make our code more declarative.

modules/tbl/fsw/src/cfe_tbl_passthru_codec.c
```diff
@@ -161,7 +161,7 @@ CFE_Status_t CFE_TBL_ValidateCodecLoadSize(CFE_TBL_TxnState_t *Txn, const CFE_TB
     /* Because the file sizes and in-memory size is the same, this check is simple */
     ActualSize    = CFE_TBL_RegRecGetSize(CFE_TBL_TxnRegRec(Txn));
     ProjectedSize = HeaderPtr->Offset + HeaderPtr->NumBytes;
-    if (ProjectedSize > ActualSize)
+    if (ProjectedSize < HeaderPtr->Offset || ProjectedSize > ActualSize)
```
ivanperez-keera, line 164: Same as above.

ivanperez-keera (approved):
Consider comments at your discretion.

ivanperez-keera (approved):
Approved with comments for your consideration.

PR #2694, opened 2026-04-21
Fix #2417, add size check in CFE_MSG_Init

### LINKED ISSUE #2417: Default CFE_MSG_Init function will initialize a message with size of 0
**Describe the bug**
The 3rd argument to `CFE_MSG_Init()` indicates the actual/real size of the structure that should be initialized.  All of the operations that deal with a primary header assume it must be _at least_ this size, and write to the primary header.

However, CFE_MSG_Init() does not confirm that the size is at least the minimum required.  It will still write fields (e.g. MsgID) into the primary header, even if the size passed in is 0.

**To Reproduce**
Call default `CFE_MSG_Init()` with size specified as 0.  The MsgID is still written into the primary header.  

This is basically a buffer overflow, as the function wrote to memory that was beyond the end of the buffer.

**Expected behavior**
The function should never write beyond the end of the buffer.  Passing a size of 0 should not initialize/write anything because there is no buffer to write to.

**Code snips**
Implementation of CFE_MSG_Init - note this proceeds to "InitDefaultHdr" regardless of size.
https://github.com/nasa/cFE/blob/a8b20ac7a0bdc2b765d824b3ff49e78934ce03c9/modules/msg/fsw/src/cfe_msg_init.c#L34-L45

**System observed on:**
Debian

**Additional context**
Related issue in CF that was relying on this bug -- this is being fixed in nasa/CF#401

**Reporter Info**

heathdutton:

**Describe the contribution**
Fixes #2417

`CFE_MSG_Init` only NULL-checks `MsgPtr`, then calls `CFE_MSG_InitDefaultHdr` and `CFE_MSG_SetMsgId` before `CFE_MSG_SetSize` performs any size validation. If the caller passes `Size` smaller than the header, those earlier calls have already written out-of-bounds by the time `SetSize` returns `CFE_MSG_BAD_ARGUMENT`.

Extended the existing guard to also reject `Size < sizeof(*MsgPtr)`. Using `sizeof(*MsgPtr)` naturally covers both the pri-only (6 bytes) and priext (10 bytes) header configurations without any ifdef.

**Testing performed**
Added one assertion in `test_cfe_msg_init.c` for `Size = sizeof(CFE_MSG_Message_t) - 1`, exercising the new branch. The pre-existing `Size = 0` assertion also now exits via this guard instead of coincidentally via `SetSize`. `clang-format --dry-run --Werror` clean.

**Expected behavior changes**
Callers passing a buffer smaller than the CCSDS header will now receive `CFE_MSG_BAD_ARGUMENT` with no writes to the buffer. Previously the same callers received `CFE_MSG_BAD_ARGUMENT` but only after the function had already corrupted memory past the end of their buffer. The downstream `SetSize` range check (`Size < CFE_MSG_SIZE_OFFSET` and upper bound) is unchanged.

The related downstream bug in CF (nasa/CF#401) that relied on this behavior is already closed.

**System(s) tested on**
Code review and clang-format only (macOS does not support cFS native build). Upstream CI will exercise the unit tests on Linux.

**Contributor Info**
Heath Dutton / Personal

### DIFF
--- a/modules/msg/fsw/src/cfe_msg_init.c
+++ b/modules/msg/fsw/src/cfe_msg_init.c
@@ -35,7 +35,7 @@ CFE_Status_t CFE_MSG_Init(CFE_MSG_Message_t *MsgPtr, CFE_SB_MsgId_t MsgId, CFE_M
 {
     int32 status;
 
-    if (MsgPtr == NULL)
+    if (MsgPtr == NULL || Size < sizeof(*MsgPtr))
     {
         return CFE_MSG_BAD_ARGUMENT;
     }
--- a/modules/msg/ut-coverage/test_cfe_msg_init.c
+++ b/modules/msg/ut-coverage/test_cfe_msg_init.c
@@ -57,6 +57,7 @@ void Test_MSG_Init(void)
     UT_SetDefaultReturnValue(UT_KEY(CFE_SB_IsValidMsgId), true);
     UtAssert_INT32_EQ(CFE_MSG_Init(NULL, msgid_act, sizeof(cmd)), CFE_MSG_BAD_ARGUMENT);
     UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgid_act, 0), CFE_MSG_BAD_ARGUMENT);
+    UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgid_act, sizeof(CFE_MSG_Message_t) - 1), CFE_MSG_BAD_ARGUMENT);
     UT_SetDefaultReturnValue(UT_KEY(CFE_SB_IsValidMsgId), false);
     UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgid_act, sizeof(cmd)), CFE_MSG_BAD_ARGUMENT);
 


sylvesterkaczmarek (commented):
This adds the lower-bound check, but `memset(MsgPtr, 0, Size)` still runs before any upper-bound validation. `CFE_MSG_SetSize()` rejects oversized values only after the memset, so an oversized `Size` can still drive a write beyond the valid message range before the function returns an error. Could the same pre-memset validation also enforce the maximum size accepted by `CFE_MSG_SetSize()`?

heathdutton:
@sylvesterkaczmarek thanks for looking. Upper bound isn't a memory issue though. `memset` writes exactly the Size the caller declared. If `Size` is wrong nothing here catches it anyway. `0xFFFF + 7` is the CCSDS length field limit, buffer bounds are SB's job.

Lower bound is different. Header setters write at fixed offsets regardless of Size, so we know when we'd overrun (#2417).

Early upper check just buys "no mutation on error". Is a separate issue, and the bound is private to `cfe_msg_ccsdspri.c` I think. Could be a follow-up.

sylvesterkaczmarek:
That distinction makes sense. I was conflating the representable CCSDS message size with the caller’s actual buffer extent, which `CFE_MSG_Init` cannot infer. The lower bound prevents the known fixed-header writes from exceeding a too-small declared message, while an upper bound would be a separate API/no-mutation-on-error constraint rather than a memory-safety check here. I withdraw the upper-bound concern.

PR #2709, opened 2026-05-08
Fix #2707, #2710 Hardening/memsafety and tbl overflow

### LINKED ISSUE #2707: Integer overflow in cfe_tbl_load.c
**Describe the bug**
The fix in PR #2691 for CVE-2026-5476 addresses the integer overflow in the codec validation layer (Site 1), but the same overflow pattern exists downstream in `cfe_tbl_load.c` (Site 2) and was not patched. 

In `LoadContentFromFile`, `Offset + NumBytes` is computed using `size_t`, which is 32-bit on ARM/SPARC/PPC RTEMS targets. The result wraps the same way as the codec site, bypassing the bounds check. OS_read then writes NumBytes bytes at `DestPtr + Offset`, where the large `Offset` wraps the pointer backward through BSS on 32-bit systems, producing an out-of-bounds write to an attacker-controlled location.

While the patched codec check currently gates this code path, the downstream overflow should be fixed for defense in depth.

**To Reproduce**
I've created a PoC that chains site 1 & 2 overflows for Remote Code Execution (RCE) on a live target. I'm withholding it from this public issue, but I can share it with maintainers through a private channel.

At a high level:
1. Craft a table file with Offset and NumBytes values that sum to exactly 2³² (wrapping to 0)
2. Both bounds checks (Site 1 codec, Site 2 `LoadContentFromFile`) see 0 and pass
3. `OS_read` writes `NumBytes` bytes at `DestPtr + Offset`, where the large `Offset` wraps the pointer backward on 32-bit
4. Attacker-controlled data overwrites BSS memory at a predictable location

**Expected behavior**
Table files with `Offset + NumBytes` exceeding the table size should be rejected at all validation sites, including `cfe_tbl_load.c`.

**Code snips**
`modules/tbl/fsw/src/cfe_tbl_load.c`:252-266
```c
LoadTailSize = Offset + NumBytes; 									// size_t=4 on 32-bit, wraps
if (LoadTailSize > CFE_TBL_LoadBuffGetAllocSize(WorkingBufferPtr))	// bypassed
{ ... } 
else
{
	DestPtr = CFE_TBL_LoadBuffGetWritePointer(WorkingBufferPtr);
	DestPtr += Offset;           									// pointer wraps backward on 32-bit
	OsStatus = OS_read(FileDescriptor, DestPtr, NumBytes);			// OOB write
	...
}
```
The same pattern exists in v6.x at `cfe_tbl_internal.c`:968-988 and `cfe_tbl_task_cmds.c`:431-443.

**System observed on:**
 - Hardware: x86_64 host, QEMU user-mode ARM32 emulation
 - OS: Linux (ARM32 cross-compiled binary)
 - Versions: cFE v6.6.0a through v7.0.0. PoC targets Draco v7.0.0, commit 70b80172

**Additional context**
- While uplink access is already privileged, normal commands don't allow for complete system takeover the way RCE does.
- `Offset` and `NumBytes` are read directly from the table file header and are not bound by `CFE_PLATFORM_TBL_MAX_DBL_TABLE_SIZE` or any other preconfigured values.
- In v6.x, where the codec layer does not gate the load path, `TBL_LOAD_CC` reaches the overflow directly, making it a single-command RCE vector with no specific requirements.

**Reporter Info**
Magnus Punga, Equinox Security

Not4right:
* [ ☑️ ] I reviewed the [Contributing Guide](https://github.com/nasa/cFE/blob/main/CONTRIBUTING.md).
* [ ☑️ ] I signed and emailed the appropriate [Contributor License Agreement](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) to GSFC-SoftwareRelease@mail.nasa.gov and copied cfs-program@lists.nasa.gov.

**Describe the contribution**

Closes: #2707, #2710

Fixes an unsigned integer overflow in `CFE_TBL_ValidateCodecLoadSize` plus related memory-safety cleanups. `Offset + NumBytes` (both from the untrusted file header) is computed as `uint32` with no overflow check, then compared to the table size. Crafted input (`Offset=0xFFFFFFF0, NumBytes=0x100`) wraps to `0xF0`, passes the check, and `CFE_TBL_LoadContentFromFile` writes attacker-controlled bytes to an attacker-controlled offset on 32-bit `size_t` targets. Masked on x86_64 dev/CI, exposed on every 32-bit flight target.

| # | File | Class |
|---|------|-------|
| 1 | `cfe_tbl_passthru_codec.c` | **Critical** — integer overflow → OOB write |
| 2 | `cfe_tbl_eds_codec.c` | **Critical** — same as #1 |
| 3 | `cfe_tbl_load.c` | **Critical** — defense-in-depth on `size_t` path |
| 4 | `cfe_tbl_internal.c` | High — `MarkNameAsModified`: NULL-deref UB, `size_t` underflow, banned `strcpy` |
| 5 | `cfe_es_mempool.c` | Medium — log `OS_MutSemTake/Give` failures (matches SB/FS/TBL convention) |
| 6 | `cfe_es_cds.c` | Medium — `sprintf` → `snprintf` |
| 7 | `cfe_sb_priv.c` | Medium — `sprintf` → `snprintf` |

Including #2707, #2710

**Testing performed**

1. Build: `make SIMULATION=native ENABLE_UNIT_TESTS=true prep && make` clean.
2. `make test`: 115/117 pass. The two failures (`osal-core-test`, `queue-test`) reproduce on unpatched `main` in the same environment POSIX mqueue / signal limits in the container, not regressions. None of the patched files are exercised by either.
3. Synthetic harness mirroring the validation logic, built 64-bit and `-m32`:
   - Unpatched: accepts malicious input (bug confirmed).
   - Patched: rejects malicious input, accepts legitimate input. No regression.

A unit test in `cfe/modules/tbl/ut-coverage/` covering wrap-around inputs should land with this fix. Happy to add it here or as a follow-up.

**Expected behavior changes**

 - API Change: none.
 - Behavior Change: malformed table files whose `Offset + NumBytes` would overflow are now rejected with `CFE_TBL_ERR_FILE_TOO_LARGE`. `MarkNameAsModified` is a no-op on buffers too small for the suffix (previously UB). Mempool mutex failures now log to syslog. Nominal flight paths unchanged.

**System(s) tested on**

 - Hardware: x86_64 (Linux container)
 - OS: Ubuntu 22.04, GCC 11.4 (also `-m32` via `gcc-multilib`)
 - Versions: cFS @ `main` HEAD at audit time

A 32-bit flight-target build (RTEMS or VxWorks) is recommended before merge — that is the configuration where the original load-path bug actually triggers.

**Additional context**

Audit scope: memory safety and mission-assurance bugs in cFE core framework (no apps). The matching security issue contains the conceptual reproducer.

**Third party code**

None.

**Contributor Info**

[Montather] — Personal

### DIFF
--- a/modules/es/fsw/src/cfe_es_cds.c
+++ b/modules/es/fsw/src/cfe_es_cds.c
@@ -672,8 +672,11 @@ void CFE_ES_FormCDSName(char *FullCDSName, const char *CDSName, CFE_ES_AppId_t T
     /* Ensure that AppName is null terminated */
     AppName[OS_MAX_API_NAME - 1] = '\0';
 
-    /* Complete formation of processor specific table name */
-    sprintf(FullCDSName, "%s.%s", AppName, CDSName);
+    /* Complete formation of processor specific table name.
+     * Use snprintf with the documented full-name length so that any future
+     * widening of OS_MAX_API_NAME or CFE_MISSION_ES_CDS_MAX_NAME_LENGTH
+     * cannot silently overflow the caller-supplied buffer. */
+    snprintf(FullCDSName, CFE_MISSION_ES_CDS_MAX_FULL_NAME_LEN, "%s.%s", AppName, CDSName);
 }
 
 /*----------------------------------------------------------------
--- a/modules/es/fsw/src/cfe_es_cds.c
+++ b/modules/es/fsw/src/cfe_es_cds.c
@@ -672,11 +672,8 @@ void CFE_ES_FormCDSName(char *FullCDSName, const char *CDSName, CFE_ES_AppId_t T
     /* Ensure that AppName is null terminated */
     AppName[OS_MAX_API_NAME - 1] = '\0';
 
-    /* Complete formation of processor specific table name.
-     * Use snprintf with the documented full-name length so that any future
-     * widening of OS_MAX_API_NAME or CFE_MISSION_ES_CDS_MAX_NAME_LENGTH
-     * cannot silently overflow the caller-supplied buffer. */
-    snprintf(FullCDSName, CFE_MISSION_ES_CDS_MAX_FULL_NAME_LEN, "%s.%s", AppName, CDSName);
+    /* Complete formation of processor specific table name */
+    sprintf(FullCDSName, "%s.%s", AppName, CDSName);
 }
 
 /*----------------------------------------------------------------
--- a/modules/es/fsw/src/cfe_es_mempool.c
+++ b/modules/es/fsw/src/cfe_es_mempool.c
@@ -61,49 +61,6 @@
 
 /*----------------------------------------------------------------
  *
- * Internal helper — take a pool mutex if defined, log and proceed on failure.
- * Matches the "log and continue" convention used by CFE_SB_LockSharedData /
- * CFE_FS_LockSharedData / CFE_TBL_LockRegistry so a transient OSAL error does
- * not silently mask an unprotected critical section.
- *
- *-----------------------------------------------------------------*/
-static inline void CFE_ES_PoolLock(const CFE_ES_MemPoolRecord_t *PoolRecPtr, const char *FuncName)
-{
-    int32 OsStatus;
-
-    if (OS_ObjectIdDefined(PoolRecPtr->MutexId))
-    {
-        OsStatus = OS_MutSemTake(PoolRecPtr->MutexId);
-        if (OsStatus != OS_SUCCESS)
-        {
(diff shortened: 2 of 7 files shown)


sylvesterkaczmarek (commented):
`CFE_ES_PoolLock()` logs an `OS_MutSemTake()` failure but then returns `void`, so callers still enter the pool critical section unprotected; `CFE_ES_PoolUnlock()` later calls `OS_MutSemGive()` even though the lock may never have been acquired. Could the lock helper return status and have callers abort, and only unlock after a successful take? Logging alone does not preserve the pool concurrency invariant.

PR #2741, opened 2026-05-21
fix nasa#2740, return value of OS_write() called by UT_BSP_WriteLogFi…

### LINKED ISSUE #2740: Return value of OS_write() called by UT_BSP_WriteLogFile() in modules/cfe_assert/src/cfe_assert_io.c is ignored.

myc-yang:

* [ ] I reviewed the [Contributing Guide](https://github.com/nasa/cFE/blob/main/CONTRIBUTING.md).
* [ ] I signed and emailed the appropriate [Contributor License Agreement](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) to GSFC-SoftwareRelease@mail.nasa.gov and copied cfs-program@lists.nasa.gov.

**Describe the contribution**
Fixes nasa#2740 - return value of OS_write() called by UT_BSP_WriteLogFile() in modules/cfe_assert/src/cfe_assert_io.c casted to `(void)` in order to suppress warning.

**Testing performed**

**Expected behavior changes**
 - None

**System(s) tested on**
 - Hardware: PC
 - OS: Ubuntu 18.04
 - Versions: [e.g. cFE 6.6, OSAL 4.2, PSP 1.3 for mcp750, any related apps or tools]

**Additional context**

**Third party code**
If included, identify any third party code and provide text file of license

Michael Yang NASA/GSFC

### DIFF
--- a/modules/cfe_assert/src/cfe_assert_io.c
+++ b/modules/cfe_assert/src/cfe_assert_io.c
@@ -88,7 +88,10 @@ void UT_BSP_WriteLogFile(osal_id_t FileDesc, uint8 MessageType, const char *Pref
     if (MsgEnabled & 1)
     {
         snprintf(LogFileBuffer, sizeof(LogFileBuffer), "[%5s] %s\n", Prefix, OutputMessage);
-        OS_write(FileDesc, LogFileBuffer, strlen(LogFileBuffer));
+
+        /* SAD: Suppress Ignored Return Value warning from CodeSonar.  There is no alternative action to take if
+         * OS_write fails  */
+        (void)OS_write(FileDesc, LogFileBuffer, strlen(LogFileBuffer));
     }
 }
 


jphickey (changes requested):
Ignoring the return value here is intentional because there is no "recovery", its a best-effort approach.

Calling another function that does terminal output to report that a function failed on terminal output is probably not a good idea, this could create an endless loop.

myc-yang:
Updated to suppress codeSonar warning with comment and casting to a (void)

jphickey (dismissed):
I'm generally OK with it, but I do think the comment should include something about WHY we are suppressing it (i.e. the write is best effort, there is no alternative action if it fails) ... not just that we are suppressing it.

myc-yang:
> I'm generally OK with it, but I do think the comment should include something about WHY we are suppressing it (i.e. the write is best effort, there is no alternative action if it fails) ... not just that we are suppressing it.

I've updated the comment with an explanation.

jphickey (approved):

PR #2743, opened 2026-05-24
ci: declare workflow-level `contents: read` on 4 check workflows

arpitjain099:
Four lint/analysis workflows (`format-check`, `codeql-build`, `static-analysis`, `mcdc`) just run pure checks. No GitHub API writes from the workflows, so workflow-level `contents: read` is the right cap.

Note: `codeql-build` writes via `security-events: write` on the analyze job, so I left that job-level permissions alone (this PR only adds the workflow-level cap, jobs can still override upward where they need to).

Same post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/codeql-build.yml
+++ b/.github/workflows/codeql-build.yml
@@ -15,6 +15,9 @@ on:
     # 11:15 PM UTC every Sunday
     - cron:  '15 23 * * 0'
 
+permissions:
+  contents: read
+
 jobs:
   codeql:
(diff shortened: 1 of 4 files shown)


arpitjain099:
Closing in favor of #2730 to avoid duplicate/overlapping permissions PRs in this repo. The extra workflow this PR touched needs write scopes (CodeQL, auto-update, or a user/issue responder), so it should not get a blanket contents: read; I am leaving that to the maintainers. #2730 hardens the read-only workflows here.

PR #2748, opened 2026-06-01
Fix #2746, cfe use of strcmp

### LINKED ISSUE #2746: Use of strcmp
Static analyzer found several instances of 'strcmp' in cfe. The use of strcmp could potentially result in buffer overflows. The instances are:

1. cfe_config_get.c : CFE_Config_GetIdByName
2. cfe_es_apps.c : CFE_ES_ParseFileEntry (2 instances)
3. cfe_es_apps.c : CFE_ES_LoadModule
5. cfe_es_cds.c : CFE_ES_LocateCDSBlockRecordByName
6. cfe_es_resource.c : CFE_ES_LocateAppRecordByName
7. cfe_es_resource.c : CFE_ES_LocateLibRecordByName
8. cfe_es_resource.c : CFE_ES_LocateCounterRecordByName
9. cfe_tbl_load.c : CFE_TBL_ValidateFileIsLoadable
10. cfe_tbl_regrec.c : CFE_TBL_LocateRegRecByName

jlombay:

**Describe the contribution**
Added null check to argument in function CFE_Config_GetIdByName. 
- Fixes #2746 

**Testing performed**
1. make native_std.install
1. make native_std.runtest

**Expected behavior changes**
No impact to behavior. 

**System(s) tested on**
 - Hardware: PC
 - OS: Ubuntu
 - Versions: cFE v7.0.1

José Lombay, NASA GRC


jphickey (changes requested):
In general this is another case where it seems like CodeSonar is just pointing out some general info about "strcmp" and is not indicating that the code is flawed or needs to be changed.
In all these cases, the strings were already ensured to be null terminated.  This is the only requirement for safe use of `strcmp()`.  Blindly changing this to `strncmp()` may have unintended consequences where one string is longer than the other but the fixed-length portion is the same.
My recommendation -- so long as the strings being passed to `strcmp` are guaranteed to be null terminated (and they should be, this was established when these data structures were instantiated) just document that and leave the code alone.  Do not change to `strncmp` unless there is a valid reason to do so.

jlombay:
> In general this is another case where it seems like CodeSonar is just pointing out some general info about "strcmp" and is not indicating that the code is flawed or needs to be changed. In all these cases, the strings were already ensured to be null terminated. This is the only requirement for safe use of `strcmp()`. Blindly changing this to `strncmp()` may have unintended consequences where one string is longer than the other but the fixed-length portion is the same. My recommendation -- so long as the strings being passed to `strcmp` are guaranteed to be null terminated (and they should be, this was established when these data structures were instantiated) just document that and leave the code alone. Do not change to `strncmp` unless there is a valid reason to do so.

Thanks for the feedback. After our conversation, will undo the changes back to the `strcmp()`. And will look to suppress the warning from the static analysis tool.

jphickey (changes requested):
Something seems amiss here, there are 74 files in this commit?  Many diffs seem unrelated.  Seems like an incorrect merge or rebase.

jlombay:
> Something seems amiss here, there are 74 files in this commit? Many diffs seem unrelated. Seems like an incorrect merge or rebase.

You're correct. I think I should just close this PR and Issue #2746 and make a new one, because the change has nothing to do with the stncmp anymore.

PR #2751, opened 2026-06-01
Fix #2750, add in Task Affinity TA module to cFE

### LINKED ISSUE #2750: Add TA Task Affinity to CFE modules
Add Task Affinity module to CFE. This is to assign different tasks to specific cores in an SMP system

arodarrow:

**Describe the contribution**
Adding SMP API to OSAL for POSIX, VxWorks, RTEMS. Also added a no-impl for those configurations that don't support SMP or that there isn't an implementation for it yet.

**Testing performed**
Built with the 3 different OS Linux, RTEMS, VxWorks. Checked for functionality of API using the new cFE module TA. 

**Expected behavior changes**
Able to change task affinity and get core num information

**System(s) tested on**
 - Hardware: PC, VxSIM(on a linux VM), GR740
 - OS: Debian, VxWorks, RTEMS

**Additional context**

**Third party code**
If included, identify any third party code and provide text file of license

Adrian Rodriguez NASA GSFC


modules/ta/fsw/inc/cfe_ta_tasklist.h
```diff
@@ -0,0 +1,163 @@
...
+            (uint8)(((val) >> 8) & 0xFF),  \
+            (uint8)(((val) >> 16) & 0xFF), \
+            (uint8)(((val) >> 24) & 0xFF), \
+            (uint8)(((val) >> 32) & 0xFF), \
+            (uint8)(((val) >> 40) & 0xFF), \
+            (uint8)(((val) >> 48) & 0xFF), \
+            (uint8)(((val) >> 56) & 0xFF)  \
+        }                                  \
+    }
+#endif
```
jphickey: This long `#if`/`#elif` chain has to be simplified.  This is way too complicated.  The `affinity_mask` should be a simple array of uint8 values and it should use OSAL abstractions to get/set bits in that.

This also could be done with a function and a for loop.

jphickey (changes requested):
Just need to clean up the macro definition.  We cannot have so many different options.  Also we should be able to painlessly go above 64.

modules/ta/fsw/inc/cfe_ta_tasklist.h
arodarrow: I moved the helper macro into the affinity list file and removed those ifdefs

jphickey (approved):

PR #2754, opened 2026-06-02
fix: guard CFE_SB_GetUserDataLength against unsigned underflow

### LINKED ISSUE #2697: [SECURITY] Unsigned integer underflow in CFE_SB_GetUserDataLength returns SIZE_MAX-class value for malformed messages, enabling OOB read in downstream consumers
* [x] I performed a cursory search to see if the bug report is relevant, not redundant, nor in conflict with other tickets.

**Describe the bug**
`CFE_SB_GetUserDataLength()` performs an unsigned subtraction `TotalMsgSize - HdrSize` without verifying that `TotalMsgSize >= HdrSize`. When a Software Bus message arrives whose CCSDS Length field encodes a total smaller than the message's secondary header (e.g. Length=0 on a Cmd or Tlm message), the subtraction underflows to a near-`SIZE_MAX` value (`0xFFFFFFFFFFFFFFFF` on 64-bit, `0xFFFFFFFF` on 32-bit). The function silently violates its documented contract (`@retval 0 if an error occurs`, [`cfe_sb.h:707-708`](modules/core_api/fsw/inc/cfe_sb.h#L707-L708)).

Any caller that follows the documented contract and uses the return value to drive a `memcpy`, `memset`, fixed-step loop, or buffer allocation performs the operation with a multi-GB length. The in-tree consumer pattern at [`io_lib/fsw/src/services/tm_sdlp.c:68`](fsw/apps/io_lib/fsw/src/services/tm_sdlp.c#L68) demonstrates exactly this idiom (`for (byte = 0; byte < idleDataLength; ++byte) pIdleData[byte] = ...`), and downstream `cfe_testcase` validates the same usage at [`sb_sendrecv_test.c:532`](modules/cfe_testcase/src/sb_sendrecv_test.c#L532) without exercising the malformed-input case.

The prior fix #1192 / commit `d5bd7ca` only initialised `TotalMsgSize` to 0 to silence a static-analysis warning for the `MsgPtr == NULL` path; the `MsgPtr != NULL && malformed` underflow remains unaddressed on both `main` and `origin/dev`.

**To Reproduce**
1. Construct a CCSDS Space Packet with Length field = `0` (`Length[0]=0, Length[1]=0`), Sec-Hdr-Flag = 1, valid MsgID for a Cmd or Tlm topic the target subscribes to.
2. Deliver the packet to the local Software Bus by any in-process or external means (`CFE_SB_TransmitMsg`, CI ingest, SBN peer, …).
3. Have an application call `CFE_SB_GetUserDataLength(msg)`.

| Input Length field | TotalMsgSize | HdrSize (Cmd hdr) | Returned |
|---|---|---|---|
| 50 (benign) | 57 | 8 | 49 (correct) |
| 0 (malformed) | 7 | 8 | **18,446,744,073,709,551,615** (= SIZE_MAX) |
| 0 (malformed, Tlm) | 7 | 16 | **18,446,744,073,709,551,607** |

A standalone reproduction using verbatim cFE constants and the actual function body is available; output verified on `main @ 432c9302`. PoC source available on request.

**Expected behavior**
`CFE_SB_GetUserDataLength()` should return `0` for any malformed message where `TotalMsgSize < HdrSize`, matching the documented contract. The fix is local and trivial:

```c
if (CFE_MSG_GetSize(MsgPtr, &TotalMsgSize) != CFE_SUCCESS) return 0;
HdrSize = CFE_SB_MsgHdrSize(MsgPtr);
if (TotalMsgSize < HdrSize) return 0;            /* contract: 0 on error */
return TotalMsgSize - HdrSize;
```

**Code snips**
[`modules/sb/fsw/src/cfe_sb_util.c:67-81`](modules/sb/fsw/src/cfe_sb_util.c#L67-L81) (identical on `main` and `origin/dev`):
```c
size_t CFE_SB_GetUserDataLength(const CFE_MSG_Message_t *MsgPtr)
{
    CFE_MSG_Size_t TotalMsgSize = 0;
    size_t         HdrSize;

    if (MsgPtr == NULL)
    {
        return TotalMsgSize;
    }

    CFE_MSG_GetSize(MsgPtr, &TotalMsgSize);   /* attacker-controlled value */
    HdrSize = CFE_SB_MsgHdrSize(MsgPtr);

    return TotalMsgSize - HdrSize;            /* underflow when Total < Hdr */
}
```

Documented contract — [`modules/core_api/fsw/inc/cfe_sb.h:707-708`](modules/core_api/fsw/inc/cfe_sb.h#L707-L708):
```
\return The size (in bytes) of the user data in the software bus message.
\retval 0 if an error occurs, such as if the MsgPtr argument is not valid.
```

Source of attacker-controlled value — [`modules/msg/fsw/src/cfe_msg_ccsdspri.c:376-386`](modules/msg/fsw/src/cfe_msg_ccsdspri.c#L376-L386).

**System observed on:**
 - Hardware: x86-64 (effect demonstrated; SIZE_MAX-class return value also occurs on 32-bit targets as `0xFFFFFFFF`)
 - OS: Linux (Ubuntu 24.04 / Docker), reproduction is OS-independent
 - Versions: cFE `main` @ commit `432c9302` (HEAD as of 2026-04-26); same code present on `origin/dev`

**Additional context**
- Existing unit tests `Test_CFE_SB_SetGetUserDataLength` ([`sb_UT.c:4518`](modules/sb/ut-coverage/sb_UT.c#L4518)) and `TestMessageUserDataAccess` ([`sb_sendrecv_test.c:516`](modules/cfe_testcase/src/sb_sendrecv_test.c#L516)) only exercise the well-formed case; the underflow path is uncovered.
- Issue #1683 already requests broader functional tests for these APIs but does not identify the underflow.
- Distinct from but related to two companion findings filed separately: a missing lower-bound check on the SB transmit path (`MessageTxn_SetContentSize`, `TransmitMsgValidate`) and an unbounded `PktLen` loop in `CFE_MSG_ComputeCheckSum`. All three share the root cause that the cFE Software Bus / Message layer trusts the attacker-controlled CCSDS Length field without lower-bound validation; each requires its own patch.
- Adding regression cases for `Length=0` (Cmd) and `Length=0` (Tlm) to the existing UTs would prevent recurrence.

stark256-spec:
Fixes #2697. When a SB message's CCSDS Length field encodes a total smaller than the header size, the unsigned subtraction `TotalMsgSize - HdrSize` wraps to SIZE_MAX. Every downstream caller using this value as a buffer size or loop bound then performs an unbounded OOB read. Adds a `TotalMsgSize <= HdrSize` guard returning 0 for malformed messages, consistent with the existing NULL-pointer guard.

### DIFF
--- a/modules/sb/fsw/src/cfe_sb_util.c
+++ b/modules/sb/fsw/src/cfe_sb_util.c
@@ -77,6 +77,15 @@ size_t CFE_SB_GetUserDataLength(const CFE_MSG_Message_t *MsgPtr)
     CFE_MSG_GetSize(MsgPtr, &TotalMsgSize);
     HdrSize = CFE_SB_MsgHdrSize(MsgPtr);
 
+    /* Guard against unsigned underflow: a malformed message whose CCSDS
+     * Length field encodes a total smaller than the header size would wrap
+     * the subtraction to SIZE_MAX, causing downstream OOB reads in every
+     * caller that uses the returned length as a buffer size or loop bound. */
+    if (TotalMsgSize <= HdrSize)
+    {
+        return 0;
+    }
+
     return TotalMsgSize - HdrSize;
 }
 


stark256-spec:
Looks like @linerfan5114's PR #2703 was opened for this issue before mine — I wasn't aware of it when I submitted #2754. Same root cause (unsigned underflow in `CFE_SB_GetUserDataLength`), so no need for two fixes here.

@linerfan5114 — has #2703 merged, or is it still pending review? Happy to close #2754 as a duplicate either way, or if there's something my version handles differently that's useful, glad to fold it into yours instead. Just let me know.

PR #2758, opened 2026-06-03
fix: perf log DataStart offset, %u format specifiers, ResourceId narrowing conversion

### LINKED ISSUE #2637: Invalid MetaData in CFE_es_perf
At some point the perf code that writes the data buffer to a file was rewritten. At this time they changed the logic to reorder the data in the buffer. This is fine except they did not update the MetaData to match the new data buffer order. 

The MetaData is used to parse and process the data file. It includes a data start index and a data end index. Since this is a circular buffer the start and end can be anywhere in the buffer, it all depends on when the data collection was stopped. So the data start and data end indicates where in buffer the data starts and ends.

The code that writes the buffer to a file used to just dump the buffer as is to the file. The code now uses the data start index to initialize the first entry in the loop that writes each entry to the file, so the data in the file is now ordered starting with the start entry. 

The data start and end index in the metadata is still initialized to the original start and end index values and so now does not match the actual order of the data in the file. This results in the existing applications that read the file will only process entries from the start index to the end of the buffer, skipping entries before the start index. The number of entries that are loaded will vary depending on the location of the start index in the buffer when the data collection was stopped.

Solution: Don't reorder the entries in the buffer

cfe_es_perf.c
Function: CFE_ES_RunPerfLogDump
Change:
                case CFE_ES_PerfDumpState_WRITE_PERF_ENTRIES:
                    State->DataPos      = Perf->MetaData.DataStart;
                    State->StateCounter = Perf->MetaData.DataCount;
                    break;

to:
                case CFE_ES_PerfDumpState_WRITE_PERF_ENTRIES:
                    State->DataPos      = 0;
                    State->StateCounter = Perf->MetaData.DataCount;
                    break;

Steven Slegel

### LINKED ISSUE #2638: Implicit type conversion in CFE_ES_PerfLogAdd())
There is an implicit type conversion in the string formatting where the variable 'Marker' is defined as an unsigned integer and is formatted for printing as a signed integer %d. 

Fix is to change string formatting from %d to %u.

https://github.com/nasa/cFE/blob/4b3fedd971d704ce4782f62ad92f9f0a56ff8eaa/modules/es/fsw/src/cfe_es_perf.c#L589

Reporter Info
Mark Foster
NASA / Johnson Space Center / ER6

### LINKED ISSUE #2664: Narrowing conversion warning for CFE_ResourceId_FromInteger
**Describe the bug**
When compiling with `-Werror` and `OMIT_DEPRECATED=ON` with C++ code,  the code does not compile because there is a narrowing warning in the `cfe_resourceid.h` file. Turning off the warning globally is not preferable.

Problem [code link](https://github.com/nasa/cFE/blob/fc8c48ad09460ce0277ad2bfcd81d376634aa2c6/modules/core_api/fsw/inc/cfe_resourceid.h#L121)

Compiler error output for gcc:
```
cFS/cfe/modules/core_api/fsw/inc/cfe_resourceid.h: In function 'CFE_ResourceId_t CFE_ResourceId_FromInteger(long unsigned int)':
cFS/cfe/modules/core_api/fsw/inc/cfe_resourceid.h:123:50: error: narrowing conversion of 'Value' from 'long unsigned int' to 'uint32' {aka 'unsigned int'} [-Werror=narrowing]
  123 |     return (CFE_ResourceId_t)CFE_RESOURCEID_WRAP(Value);
      |                                                  ^~~~~
cFS/cfe/modules/resourceid/option_inc/cfe_resourceid_strict.h:87:9: note: in definition of macro 'CFE_RESOURCEID_WRAP'
   87 |         x                      \
```

**To Reproduce**
I made an easy to see example on [compiler explorer](https://godbolt.org/z/6rrxW66j6).
Otherwise, you'll need to

1. `-Werror` and `OMIT_DEPRECATED=ON` 
2. Compile a C++ file with gcc including "cfe.h"
3. Should see the warning being treated as an error

**Expected behavior**
No warning.

**Code snips**
This is the simplified code I put in [compiler explorer](https://godbolt.org/z/6rrxW66j6).
```cpp
#include <cstdint>
typedef struct
{
    uint32_t id;
} CFE_ResourceId_t;

#define CFE_RESOURCEID_WRAP(x) \
    {                          \
        x                      \
    }

static inline CFE_ResourceId_t CFE_ResourceId_FromInteger(unsigned long Value)
{
    return (CFE_ResourceId_t)CFE_RESOURCEID_WRAP(Value);
}

int main() {}
```

**Potential Fix**
One option is to insert an explicit cast to silence this issue ([Works in C](https://godbolt.org/z/xxzTba1M5) & [Works in C++](https://godbolt.org/z/9GjY9xshv)). Not the best given this now has to assume `uint32` but maybe the simple and strict header files can provide the type used within `CFE_ResourceId_t`.
```c
static inline CFE_ResourceId_t CFE_ResourceId_FromInteger(unsigned long Value)
{
    return (CFE_ResourceId_t)CFE_RESOURCEID_WRAP((uint32)Value);
}
```

**System observed on**
This code is in develop of cFE

**Reporter Info**
Philip Cooksey @ NASA/ARC

stark256-spec:
## Summary

Three independent defects across two files.

---

### Fix 1 — Perf log metadata `DataStart` mismatches file layout (closes #2637)

**File:** `modules/es/fsw/src/cfe_es_perf.c`

`CFE_ES_RunPerfLogDump` rewrites the dump buffer in order starting at `DataStart`, so the output file always begins at index 0. However, `State->DataPos` was still initialised to `Perf->MetaData.DataStart` when entering `WRITE_PERF_ENTRIES`, leaving `MetaData.DataStart` unchanged in the file. Parsers that honour this field skip every entry before the original start index — the number of skipped entries varies with wherever the circular buffer happened to stop.

**Fix:** set `State->DataPos = 0` to match the actual file layout.

---

### Fix 2 — `%d` format specifiers for unsigned values (closes #2638)

**File:** `modules/es/fsw/src/cfe_es_perf.c`

`CFE_ES_PerfLogAdd` formats `Marker` (uint32) and `CFE_MISSION_ES_PERF_MAX_IDS` with `%d` (signed). Changed both to `%u` and added an explicit `(unsigned int)` cast on the macro expression to satisfy `-Wformat`.

---

### Fix 3 — Narrowing conversion in `CFE_ResourceId_FromInteger` (closes #2664)

**File:** `modules/core_api/fsw/inc/cfe_resourceid.h`

In strict mode `CFE_RESOURCEID_WRAP(x)` expands to a compound-literal struct initializer `{ x }` where the field is `uint32`. Passing an `unsigned long` directly is a narrowing conversion that becomes a hard error when compiling C++ with `-Werror=narrowing` on 64-bit Linux (where `unsigned long` is 64 bits).

**Fix:** add an explicit `(uint32)` cast on `Value` before the wrap, consistent with the suggestion in the issue.

---

## Test plan

- [ ] Existing unit tests pass (no logic change for Fix 2 and Fix 3)
- [ ] For Fix 1: verify perf log dump output parses cleanly from index 0
- [ ] For Fix 3: confirm no `-Werror=narrowing` with `OMIT_DEPRECATED=ON` C++ build

### DIFF
--- a/modules/core_api/fsw/inc/cfe_resourceid.h
+++ b/modules/core_api/fsw/inc/cfe_resourceid.h
@@ -147,7 +147,7 @@ static inline unsigned long CFE_ResourceId_ToInteger(CFE_ResourceId_t id)
  */
 static inline CFE_ResourceId_t CFE_ResourceId_FromInteger(unsigned long Value)
 {
-    return (CFE_ResourceId_t)CFE_RESOURCEID_WRAP(Value);
+    return (CFE_ResourceId_t)CFE_RESOURCEID_WRAP((uint32)Value);
 }
 
 /**
--- a/modules/es/fsw/src/cfe_es_dispatch.c
+++ b/modules/es/fsw/src/cfe_es_dispatch.c
@@ -287,10 +287,12 @@ void CFE_ES_TaskPipe(const CFE_SB_Buffer_t *SBBufPtr)
 
     if (CFE_SB_MsgId_Equal(MessageID, SEND_HK_MID))
     {
-        /*
-        ** Housekeeping telemetry request
-        */
-        CFE_ES_SendHkCmd((const CFE_ES_SendHkCmd_t *)SBBufPtr);
+        /* Housekeeping telemetry request — verify length before casting,
+         * consistent with every command code in CFE_ES_ProcessGroundCmd(). */
+        if (CFE_ES_VerifyCmdLength(&SBBufPtr->Msg, sizeof(CFE_ES_SendHkCmd_t)))
+        {
+            CFE_ES_SendHkCmd((const CFE_ES_SendHkCmd_t *)SBBufPtr);
+        }
     }
     else if (CFE_SB_MsgId_Equal(MessageID, CMD_MID))
     {
--- a/modules/es/fsw/src/cfe_es_perf.c
+++ b/modules/es/fsw/src/cfe_es_perf.c
@@ -365,7 +365,7 @@ bool CFE_ES_RunPerfLogDump(uint32 ElapsedTime, void *Arg)
                     break;
 
                 case CFE_ES_PerfDumpState_WRITE_PERF_ENTRIES:
-                    State->DataPos      = Perf->MetaData.DataStart;
+                    State->DataPos      = 0;
                     State->StateCounter = Perf->MetaData.DataCount;
                     break;
 
@@ -614,10 +614,10 @@ void CFE_ES_PerfLogAdd(uint32 Marker, uint32 EntryExit)
         /* if marker has not been reported previously ... */
         if (Perf->MetaData.InvalidMarkerReported == false)
         {
-            CFE_ES_WriteToSysLog("%s: Invalid performance marker %d,max is %d\n",
+            CFE_ES_WriteToSysLog("%s: Invalid performance marker %u,max is %u\n",
                                  __func__,
                                  (unsigned int)Marker,
-                                 (CFE_MISSION_ES_PERF_MAX_IDS - 1));
+                                 (unsigned int)(CFE_MISSION_ES_PERF_MAX_IDS - 1));
             Perf->MetaData.InvalidMarkerReported = true;
         }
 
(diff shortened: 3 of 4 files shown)


stark256-spec:
This branch is rebased on current `main`. Three independent static-analysis findings bundled here (perf log offset calculation, format specifier mismatches, and a narrowing conversion on ResourceId) — happy to split into separate PRs if that's easier to review, or keep as-is if bundling small related fixes is fine. Let me know either way and I can request `CCB:Ready`.

PR #2759, opened 2026-06-03
fix: guard against NULL return from CFE_ES_LocateTaskRecordByID in StartAppTask

### LINKED ISSUE #2491: Possible NULL ptr deference in CFE_ES_StartAppTask
**Describe the bug**
Static code analysis finding from the Klocwork tool.  Line numbers based on the draco-rc4 tag.  On line 606 in the cfe_es_apps.c file, the function CFE_ES_LocateTaskRecordByID could return a value of NULL. This will then set TaskRecPtr to NULL. Then on line 607 TaskRecPtr will be dereferenced.

**To Reproduce**
Run Klocwork SCA analysis

**Reporter Info**
Beth Geist/NASA GSFC

stark256-spec:
## Summary

Fixes #2491

### Problem

In `CFE_ES_StartAppTask`, the return value of `CFE_ES_LocateTaskRecordByID` was used without a null check. The function can return NULL when the resource ID is out of range, leading to a null pointer dereference at the subsequent `CFE_ES_TaskRecordIsUsed()` call and `memset()`. This was flagged by Klocwork static analysis.

### Fix

Add a null check immediately after `CFE_ES_LocateTaskRecordByID`, following the same pattern used in `CFE_ES_AppCreate` (where `LocateAppRecordByID` is already guarded):

- If `TaskRecPtr == NULL`: log the error, call `OS_TaskDelete` on the newly created OSAL task to avoid leaking the OS resource, set `CFE_ES_NO_RESOURCE_IDS_AVAILABLE`, and return.
- Otherwise: continue with existing logic (in-use check, memset, field initialization).

## Test plan

- [ ] Existing ES unit tests pass
- [ ] Static analysis (Klocwork) no longer flags this path

### DIFF
--- a/modules/es/fsw/src/cfe_es_apps.c
+++ b/modules/es/fsw/src/cfe_es_apps.c
@@ -680,34 +680,47 @@ int32 CFE_ES_StartAppTask(CFE_ES_TaskId_t                *TaskIdPtr,
          */
         LocalTaskId = CFE_ES_TaskId_FromOSAL(OsalTaskId);
         TaskRecPtr  = CFE_ES_LocateTaskRecordByID(LocalTaskId);
-        if (CFE_ES_TaskRecordIsUsed(TaskRecPtr))
+
+        if (TaskRecPtr == NULL)
         {
-            CFE_ES_SysLogWrite_Unsync("%s: Error: ES_TaskTable slot for ID %lx in use at task creation!\n",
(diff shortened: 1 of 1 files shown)


stark256-spec:
Checking in on this one — static analysis flagged a real NULL-deref path here and the fix is small and self-contained. If there's a preferred reviewer for ES-module changes, happy to tag them directly. Let me know if anything needs adjusting before this goes to CCB.

PR #2762, opened 2026-06-08
Fix #2752, cFE use UtAssert_StrCmp replaced with UtAssert_STRINGBUF_EQ.

### LINKED ISSUE #2752: Use of UtAssert_StrCmp macro in unit test
Replace the use of "UtAssert_StrCmp" macros in unit testing with "UtAssert_STRINGBUF_EQ". Functionally equivalent but the latter is preferred and will also address static analysis reports.

jlombay:

**Describe the contribution**
Replaced UtAssert_StrCmp calls with UtAssert_STRINGBUF_EQ.  
- Fixes #2752 

**Testing performed**
1. make native_std.install
1. make native_std.runtest

**Expected behavior changes**
No impact to behavior

**System(s) tested on**
 - Hardware: PC
 - OS: Ubuntu 
 - Versions: cFE v7.0.1

José Lombay, NASA GRC

### DIFF
--- a/modules/cfe_testcase/src/es_cds_test.c
+++ b/modules/cfe_testcase/src/es_cds_test.c
@@ -98,7 +98,7 @@ void TestCDSName(void)
     UtAssert_INT32_EQ(CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name), CFE_ES_CDS_ALREADY_EXISTS);
 
     UtAssert_INT32_EQ(CFE_ES_GetCDSBlockName(CDSNameBuf, CDSHandlePtr, sizeof(CDSNameBuf)), CFE_SUCCESS);
-    UtAssert_StrCmp(CDSNameBuf, CDSName, "CFE_ES_GetCDSBlockName() = %s", CDSNameBuf);
+    UtAssert_STRINGBUF_EQ(CDSNameBuf, sizeof(CDSNameBuf), CDSName, sizeof(CDSName));
     UtAssert_INT32_EQ(CFE_ES_GetCDSBlockIDByName(&IdByName, CDSNameBuf), CFE_SUCCESS);
     CFE_Assert_RESOURCEID_EQ(CDSHandlePtr, IdByName);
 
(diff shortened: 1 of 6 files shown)


modules/cfe_testcase/src/es_cds_test.c
```diff
@@ -98,7 +98,7 @@ void TestCDSName(void)
     UtAssert_INT32_EQ(CFE_ES_RegisterCDS(&CDSHandlePtr, BlockSize, Name), CFE_ES_CDS_ALREADY_EXISTS);
 
     UtAssert_INT32_EQ(CFE_ES_GetCDSBlockName(CDSNameBuf, CDSHandlePtr, sizeof(CDSNameBuf)), CFE_SUCCESS);
-    UtAssert_StrCmp(CDSNameBuf, CDSName, "CFE_ES_GetCDSBlockName() = %s", CDSNameBuf);
+    UtAssert_STRINGBUF_EQ(CDSNameBuf, strlen(CDSNameBuf), CDSName, strlen(CDSName));
```
jphickey: Rather than using `strlen()` you should use the supplied constant:
`UTASSERT_STRINGBUF_NULL_TERM`
This indicates that the buffer is null terminated and it should use the string length.

Using `strlen` might cause unexpected results if one string is a substring of the other.

jphickey (changes requested):

modules/cfe_testcase/src/es_info_test.c
```diff
@@ -176,10 +174,10 @@ void TestGetLibInfo(void)
     UtAssert_INT32_EQ(CFE_ES_GetLibIDByName(&LibId, LibName), CFE_SUCCESS);
     UtAssert_INT32_EQ(CFE_ES_GetLibInfo(&LibInfo, LibId), CFE_SUCCESS);
     UtAssert_INT32_EQ(CFE_ES_GetLibName(LibNameBuf, LibId, sizeof(LibNameBuf)), CFE_SUCCESS);
-    UtAssert_StrCmp(LibNameBuf, LibName, "CFE_ES_GetLibName() = %s", LibNameBuf);
+    UtAssert_STRINGBUF_EQ(LibNameBuf, strlen(LibNameBuf), LibName, strlen(LibName));
     UtAssert_True(LibInfo.Type == CFE_ES_AppType_LIBRARY, "Lib Info -> Type = %d", (int)LibInfo.Type);
-    UtAssert_StrCmp(LibInfo.Name, LibName, "Lib Info -> Name = %s", LibInfo.Name);
-    UtAssert_StrCmp(LibInfo.EntryPoint, "CFE_Assert_LibInit", "Lib Info -> EntryPt  = %s", LibInfo.EntryPoint);
+    UtAssert_STRINGBUF_EQ(LibInfo.Name, strlen(LibInfo.Name), LibName, strlen(LibName));
+    UtAssert_STRINGBUF_EQ(LibInfo.EntryPoint, strlen(LibInfo.EntryPoint), "CFE_Assert_LibInit", -1);
```
jlombay: @jphickey I had seen the "-1" being used on other tests wherre the STRINGBUF_EQ macro was being used when passing a char array like here. 

Is it fair to assume that using "UTASSERT_STRINGBUF_NULL_TERM" for this case be preferred as well? so it'll look like:

`UtAssert_STRINGBUF_EQ(LibInfo.EntryPoint, UTASSERT_STRINGBUF_NULL_TERM, "CFE_Assert_LibInit", UTASSERT_STRINGBUF_NULL_TERM);`

jphickey (approved):

modules/cfe_testcase/src/es_info_test.c
jphickey: Yeah -1 accomplishes the same thing (it will get translated to SIZE_MAX which is also what UTASSERT_STRINGBUF_NULL_TERM is defined as).
It is preferred to use UTASSERT_STRINGBUF_NULL_TERM as it is clearer what the intent is -- the value "-1" doesn't really have any meaning.

jlombay: Thank you for the explanation!

PR #2768, opened 2026-06-23
Fix #2737, cFE use of strcpy

### LINKED ISSUE #2737: Use of strcpy

elizabash:

* [x] I signed and emailed the appropriate [Contributor License Agreement](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) to GSFC-SoftwareRelease@mail.nasa.gov and copied cfs-program@lists.nasa.gov. (This issue was discussed in a cfs meeting. I think that counts!)

**Describe the contribution**
Fix #2737
After discussion with Joe and Jose about similar instances where CodeSonar flagged certain functions (like strcpy) for not having bounds checking (though bounds checking was done right before the function) I'm adding a note in the code to ignore this in CodeSonar. I've also marked on our CodeSonar website to ignore these lines.

**Background Info Into Why This Became An Issue**
CodeSonar flagged the following error: BADFUNC.BO.STRCPY : Use of strcpy
CodeSonar Summary: A use of strcpy() or a similar function (see full list below), which are vulnerable to buffer overflows.
CodeSonar Resolution Suggestion: Use a function that does bounds checking, such as strncpy(), StrCbCopy(), or StrCchCopy().

The issue seems to be a concern over buffer overflow. The source string being copied over to a destination string could potentially cause issues if, for example, the source string doesn't have a null terminator for some reason. That would mean garbage could potentially be added into the destination string.
However, in this case, though strcpy itself isn't bounds checking, there is bounds checking being done beforehand.

**Testing performed**
1. make distclean
2. make native_std.prep
3. make native_std.install
4. make native_std.runtest
5. cd build-native_std/native/default_cpu1/
6. ctest

**Expected behavior changes**
There should be no changes in behavior to the code itself. Though CodeSonar should ignore this use of strcpy in these particular lines in the future.


elizabash:
Today during our cFS + DARPA meeting, we discussed this pull request. The PR includes only two comments to tell CodeSonar to ignore the line. This comment needs to be on every individual line that needs to be ignored. 
We decided during the meeting that instead of having a million comments everywhere telling CodeSonar to ignore certain lines, we'll instead log into CodeSonar and tell it to ignore false positives from there. That way, if the project decides to stop using CodeSonar in favor of using a different static analyzer, we won't have a bunch of useless comments everywhere in the code.

PR #2769, opened 2026-06-23
Fix #2725, Updates SB requirements

### LINKED ISSUE #2725: Scrub SB Requirements
- Review existing requirements for module
- Analyze code for module, code changes/PRs since baseline
- Compare code with requirements; identify differences

For discrepancies:
- Determine whether requirement or code contains desired behavior, and update the incorrect one (or create ticket to resolve code discrepancy)
- May involve team discussions

chillfig:

**Describe the contribution**
- Fixes #2725, Adds some grammatical updates to the SB requirements

**Testing performed**
N/A

**Expected behavior changes**
Better readability

**System(s) tested on**
N/A

**Additional context**
N/A

**Third party code**
N/A

Justin Figueroa, Vantage Systems Inc

### DIFF
--- a/docs/cFE_FunctionalRequirements.csv
+++ b/docs/cFE_FunctionalRequirements.csv
@@ -1,4 +1,4 @@
-Summary,Custom field (Requirement ID),Description,Custom field (Requirement Rationale)
+Summary,Requirement ID,Description,Requirement Rationale
 ES: Housekeeping Message,cES1000,"Upon receipt of a Message, the cFE shall generate a housekeeping message that includes the following Executive Services items:
 
 - Number of Registered Applications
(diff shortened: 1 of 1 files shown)


docs/cFE_FunctionalRequirements.csv
```diff
@@ -328,17 +330,26 @@ SB: Housekeeping Message,cSB4005,"Upon receipt of a Message, the cFE shall gener
...
+- Queue read or write error counter
+- Create Pipe error counter
+- Subscribe error counter
+- Pipe options error counter
+- Duplicate subscriptions counter
+- Get pipe ID by name error counter
+- Pipe Overflow error counter
+- MsgId-to-pipe limit error counter
+- Memory pool handle
+- Memory in use
```
czogby-nasa: "Memory in use" - Is this a boolean?  Or what?  As written, it's ambiguous what this represents.

docs/cFE_FunctionalRequirements.csv
```diff
@@ -306,17 +306,19 @@ EVS: Suppress Flooding Event Messages - Increment counter,cEVS3220.3,The cFE sha
...
 - Valid command counter
 - Invalid command counter
 - No subscriptions counter
+- Duplicate subscriptions counter
 - Message send error counter
 - Message receive error counter
+- Queue read or write error counter
 - Create Pipe error counter
 - Subscribe error counter
 - Pipe Overflow error counter
```
czogby-nasa, line 320: Looking at the rendered CSV file: Throughout the file, there are many itemized lists such as this (using either dashes or numbered lines).  Each one of these renders poorly.  Each list item should be on a new line - but instead all items are on a single line, separated by only a space.  This causes each of these lists to be extremely long horizontally (and difficult to read through), which makes it so the reader has to scroll extremely far to the right to see the other columns (and it looks unprofessional).   Is there a way to add a newline between each item?

docs/cFE_FunctionalRequirements.csv
chillfig: It is the bytes added/removed from the SB buffer memory pool allocations. I'll update to "- Memory in use (bytes)"

docs/cFE_FunctionalRequirements.csv
chillfig: Yeah that was my concern with adding all the items. There's no standard CSV escape for a forced line break that GitHub's renderer would honor. If this was a markdown file, we could do something.

czogby-nasa: Hmm.  In that case, I think it would be easier to read if you made each dash be two dashes (--).

Also, I just noticed: two of the column names contain "Custom field".  Did you mean to remove these when you gave them their actual names?

czogby-nasa (approved):

PR #2787, opened 2026-08-03
Fix #2786, include fragment in syslog dump

### LINKED ISSUE #2786: ES - Write Sys Log File Size inconsistent in Rare Case
**Describe the bug**
There may be unwritten or lost data in the ES system log. In rare cases the ES Write sys log command writes less than it was expected to write.

This bug can only be seen under certain circumstances and was pursued after seeing seemingly random instances of tests failing. 

The processor reset counter increments every time the cFS starts in our development VMs, despite shutting previous runs down with ctrl-c. 

If the processor reset count has rolled over and the system starts with a power-on start type in ES HK telemetry due to that, the System Log Mode is set to OVERWRITE instead of DICARD as it normally is. This may not contribute to this bug, but I am providing as much info as I can.

If that condition occurs, a Write Sys Log command to write the sys log to a file that is sent will produce a file size in FM file info packet and event message text reporting the size of the file that is inconsistent with processor reset starts. 

In processor reset starts, the log mode is set to DISCARD, and it is usually full (as reported by ES HK) after the system settles. When a write system log command is sent in these conditions, it reports the HK reported system log size in bytes + the header of 64 bytes in the event message as the size of the written file. A FM file info command on that file also reports the file size to be this in it's packet. Adding 64 (header size) to the ES HK tlm SYS_LOG_BYTES_USED matches these items (the event message text and the file info telemetry point). 

However, when the processor reset count rolls over due to reaching the max reset count, when the cFS starts, the System Log Mode is now of type OVERWRITE, and in this case the log bytes used in HK telemetry is usually not full, but close to it. When a Write System Log command is sent in these conditions, the size of the file reported in the event message and is the FM file info packet for the file report a size that is smaller than expected. 

The following are 3 runs where this occurred. I was able to reproduce this any time I wanted to.
### Run 1:
Curated log file: [write_sys_log_fail_1.txt](https://github.com/user-attachments/files/30521544/write_sys_log_fail_1.txt)
HK tlm at the start:
`SYS_LOG_BYTES_USED       :             3019`
Event Message Received:
`CFS-1      1970/01/12 14:07:48.064 scid:66  procid:1  DEBUG CFE_ES         18: /cf/test_syslog.txt written:Size=3055,Entries=45`
Expected Size of file based on HK Tlm: 
3083   (3019+64 header)
FM File info packet reported size:
`FM_FILE_INFO SIZE == 3055`
Final tlm size after test:
`SYS_LOG_BYTES_USED       :             3019`

difference in EM and expected: 28

### Run 2:
**This file has a size written that is LESS than the sys log size reported in HK Tlm**

Curated log file: [write_sys_log_fail_2.txt](https://github.com/user-attachments/files/30521809/write_sys_log_fail_2.txt)
HK tlm at the start:
`SYS_LOG_BYTES_USED       :             3034`
Event Message Received:
`CFS-1      1970/01/12 16:18:57.032 scid:66  procid:1  DEBUG CFE_ES         18: /cf/test_syslog.txt written:Size=3031,Entries=52`
Expected Size of file based on HK Tlm: 
3098   (3034+64 header)
FM File info packet reported size:
`FM_FILE_INFO SIZE == 3031`
Final tlm size after test:
`SYS_LOG_BYTES_USED       :             3034`

difference in EM and expected: 67  **Bigger than the size of the header!**

### Run 3:
**Despite being the same size as run 1, The file size produced for this run was different!**

Curated log file: [write_sys_log_fail_3.txt](https://github.com/user-attachments/files/30522042/write_sys_log_fail_3.txt)
HK tlm at the start:
`SYS_LOG_BYTES_USED       :             3019`
Event Message Received:
`CFS-1      1970/01/12 14:04:33.065 scid:66  procid:1  DEBUG CFE_ES         18: /cf/test_syslog.txt written:Size=3058,Entries=39`
Expected Size of file based on HK Tlm: 
3083   (3019+64 header)
FM File info packet reported size:
`FM_FILE_INFO SIZE == 3058`
Final tlm size after test:
`SYS_LOG_BYTES_USED       :             3019`

difference in EM and expected: 25

The following are examples of 2 system log files, the HK reported size for their creation is stated in the file name:
[test_syslog_good_run_HK_size_is_3072.txt](https://github.com/user-attachments/files/30522179/test_syslog_good_run_HK_size_is_3072.txt)
[test_syslog_bad_run_HK_size_is_3019.txt](https://github.com/user-attachments/files/30522182/test_syslog_bad_run_HK_size_is_3019.txt)

**To Reproduce**
1. Start the system
2. Check the ES Processor reset count
3. Set Max Reset count = processor reset count
4. Stop and start the system
5. Processor reset count should = 0 and sys log type should read OVERWRITE in ES HK tlm
6. Record the reported Bytes Used of the system log
7. the expected file size is Bytes used in ES HK Tlm + the size of the header, 64
8. Send the write sys log command 
9. Note the reported bytes written does not match the expected number of bytes

There has to be data being not written or lost in this set of conditions.

**Expected behavior**
In a normal run, the bytes used as reported in ES HK + 64 (the header size) is the size of the file. It is only under this rare condition that it is not.

**Code snips**
None

**System observed on:**
 - VM
 - Dev branch and all submodules a cFS d911c3f15d5c18b83f8808bc2ca9bd13ae52ee48

**Reporter Info**
Damon Stewart - NASA/GSFC

jphickey:

**Describe the contribution**
Do not skip the initial fragment when dumping syslog.  Write it to the log with a tag indicating it is a fragment.

Also increases the default log size to 8k, so that all messages in a normal boot can be stored.

Fixes #2786

**Testing performed**
Issue sys log write command, confirm behavior

**Expected behavior changes**
Initial fragment (after wrap) is written to the file, along with a fixed-length tag indicating that it is a fragment.

**System(s) tested on**
Linux

### DIFF
--- a/modules/es/fsw/inc/cfe_es_internal_cfg.h
+++ b/modules/es/fsw/inc/cfe_es_internal_cfg.h
@@ -164,7 +164,7 @@
 **       verified.
 */
 #define CFE_PLATFORM_ES_SYSTEM_LOG_SIZE         CFE_PLATFORM_ES_CFGVAL(SYSTEM_LOG_SIZE)
-#define DEFAULT_CFE_PLATFORM_ES_SYSTEM_LOG_SIZE 3072
+#define DEFAULT_CFE_PLATFORM_ES_SYSTEM_LOG_SIZE 8192
 
 /**
 **  \cfeescfg Define Max Number of Generic Counters
--- a/modules/es/fsw/src/cfe_es_log.h
+++ b/modules/es/fsw/src/cfe_es_log.h
@@ -110,6 +110,7 @@ typedef struct
     size_t BlockSize;  /**< Size of content currently in the "Data" member */
     size_t EndIdx;     /**< End of the syslog buffer at the time reading started */
     size_t LastOffset; /**< Current Read Position */
+    bool   IsFragment; /**< Flag indicating that the data is NOT a complete message */
 
     char Data[CFE_ES_SYSLOG_READ_BUFFER_SIZE]; /**< Actual syslog content */
 } CFE_ES_SysLogReadBuffer_t;
--- a/modules/es/fsw/src/cfe_es_syslog.c
+++ b/modules/es/fsw/src/cfe_es_syslog.c
@@ -97,27 +97,13 @@ void CFE_ES_SysLogReadStart_Unsync(CFE_ES_SysLogReadBuffer_t *Buffer)
 {
     size_t ReadIdx;
     size_t EndIdx;
-    size_t TotalSize;
 
-    ReadIdx   = CFE_ES_Global.ResetDataPtr->SystemLogWriteIdx;
-    EndIdx    = CFE_ES_Global.ResetDataPtr->SystemLogEndIdx;
-    TotalSize = EndIdx;
+    ReadIdx = CFE_ES_Global.ResetDataPtr->SystemLogWriteIdx;
+    EndIdx  = CFE_ES_Global.ResetDataPtr->SystemLogEndIdx;
 
-    /*
-     * Ensure that we start reading at the start of a message
-     * Likely pointing to an old fragment right now -- find the end of it
-     */
-    while (TotalSize > 0 && ReadIdx < EndIdx)
-    {
-        ++ReadIdx;
-        --TotalSize;
-        if (CFE_ES_Global.ResetDataPtr->SystemLog[ReadIdx - 1] == '\n')
-        {
-            break;
-        }
-    }
-
-    Buffer->SizeLeft   = TotalSize;
+    /* If the write position is NOT at the tail of the buffer, then its a fragement */
+    Buffer->IsFragment = (ReadIdx != EndIdx);
+    Buffer->SizeLeft   = EndIdx;
     Buffer->LastOffset = ReadIdx;
     Buffer->EndIdx     = EndIdx;
     Buffer->BlockSize  = 0;
@@ -286,9 +272,21 @@ int32 CFE_ES_SysLogWrite_Unsync(const char *SpecStringPtr, ...)
  *-----------------------------------------------------------------*/
 void CFE_ES_SysLogReadData(CFE_ES_SysLogReadBuffer_t *Buffer)
 {
-    size_t BlockSize;
+    size_t            BlockSize;
+    static const char FRAGMENT_TAG[] = "<<FRAGMENT>> ";
 
-    Buffer->BlockSize = 0;
+    if (Buffer->IsFragment)
+    {
+        /* Add a tag indicating that this block of data does not begin w/a complete message */
+        memcpy(Buffer->Data, FRAGMENT_TAG, sizeof(FRAGMENT_TAG) - 1);
+        Buffer->BlockSize  = sizeof(FRAGMENT_TAG) - 1;
+        Buffer->IsFragment = false;
+    }
+    else
+    {
+        Buffer->BlockSize = 0;
+    }
+
     while (Buffer->SizeLeft > 0 && Buffer->BlockSize < sizeof(Buffer->Data))
     {
         /*
--- a/modules/es/ut-coverage/es_UT.c
+++ b/modules/es/ut-coverage/es_UT.c
@@ -5276,9 +5276,9 @@ void TestSysLog(void)
 
     UtPrintf("Begin Test Sys Log");
 
-    /* Test loop in CFE_ES_SysLogReadStart_Unsync that ensures
-     * reading at the start of a message */
+    /* Test conditional in CFE_ES_SysLogReadStart_Unsync that checks for a fragment */
     ES_ResetUnitTest();
+    memset(&SysLogBuffer, 0, sizeof(SysLogBuffer));
     CFE_ES_Global.ResetDataPtr->SystemLogWriteIdx = 0;
     CFE_ES_Global.ResetDataPtr->SystemLogEndIdx   = sizeof(CFE_ES_Global.ResetDataPtr->SystemLog) - 1;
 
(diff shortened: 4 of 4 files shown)


ddstewar (approved):
Proven via retest that the 13 bytes added in are consistent with the expected result. The contents and size of files are now fully predictable.

Only anomaly I had was when I set the mode the OVERWRITE, cleared the SYS log, sent 110 start_application commands for the sample_app, which was already running, to have 74 bytes entered into the log each command. After 110, the next one would cause an overwrite.

When sending the last command to overwrite the log (same as all the other commands, so this behavior is expected), the resultant file marked a full entry as a fragment. This is not unexpected, just interesting. Attached are the log files from that test:

[test_b4roll.txt](https://github.com/user-attachments/files/30835443/test_b4roll.txt)
[test_afterroll.txt](https://github.com/user-attachments/files/30835448/test_afterroll.txt)

I do not believe this would warrant further change, but it is left to the developer to determine if this is the behavior they intended or not.

As-is, I am approving this.

ddstewar:
> Proven via retest that the 13 bytes added in are consistent with the expected result. The contents and size of files are now fully predictable.
> 
> Only anomaly I had was when I set the mode the OVERWRITE, cleared the SYS log, sent 110 start_application commands for the sample_app, which was already running, to have 74 bytes entered into the log each command. After 110, the next one would cause an overwrite.
> 
> When sending the last command to overwrite the log (same as all the other commands, so this behavior is expected), the resultant file marked a full entry as a fragment. This is not unexpected, just interesting. Attached are the log files from that test:
> 
> [test_b4roll.txt](https://github.com/user-attachments/files/30835443/test_b4roll.txt) [test_afterroll.txt](https://github.com/user-attachments/files/30835448/test_afterroll.txt)
> 
> I do not believe this would warrant further change, but it is left to the developer to determine if this is the behavior they intended or not.
> 
> As-is, I am approving this.

After continuing this line of testing, I followed up sending 130 start_app commands to start CS, which was already running. After rolling the file over again the bytes used reached max and the fragment in the file was a true fragment, but the behavior was the same and the file size was as predictable as the previous test. So I am fully confident in the behavior to be repeatable here.

jphickey:
> When sending the last command to overwrite the log (same as all the other commands, so this behavior is expected), the resultant file marked a full entry as a fragment. This is not unexpected, just interesting. Attached are the log files from that test:

Yeah, this is expected behavior.  The reality is that the marker is really just indicating that the log has overwritten data, there is no reliable way to tell if the line/message is complete or not, because we don't know what _was_ in it. 

Perhaps there is a better word than `FRAGMENT` here .... maybe just "clipped" or something.  I don't know.  The goal of this is to convey to the user that the first line they are looking at _MAY_ be incomplete, and also that there were lines before it which are not being shown at all because they are simply lost due to overwrite.

jphickey:
To clarify this is just a chance operation -- i.e. that writing a new log message in overwrite mode happened to end at exactly the same character position as the start of an old message.  The vast majority of times this does not happen and will result in the beginning of the old message being lost .... but every now and then things will line up perfectly and mean that oldest message still happens to be complete.  But the FSW can't tell the difference.

ddstewar:
> To clarify this is just a chance operation -- i.e. that writing a new log message in overwrite mode happened to end at exactly the same character position as the start of an old message. The vast majority of times this does not happen and will result in the beginning of the old message being lost .... but every now and then things will line up perfectly and mean that oldest message still happens to be complete. But the FSW can't tell the difference.

Yes I agree, The way the test is structured forces this overwrite of the exact same size. Like I said I found this to be expected once I thought about it. no issues here. As you said most of the time in almost all situations the messages will not line up so nicely and it will be a partial message. I think "FRAGMENT" is fine for this, not need to change it.

ddstewar:
It appears that it is possible that the FRAGMENT string gets written in rare cases when in DISCARD Mode.

If the log previously wrapped in OVERWRITE mode causing the FRAGMENT string to be written, then the mode is switched to DISCARD but the log is not cleared.

Is this expected or an anomaly? I can go either way.

ddstewar:
Just some more info on the state of how that occurred:

tlm points at the start of the test:
SYS_LOG_BYTES_USED       :             8140
SYS_LOG_SIZE             :             8192
SYS_LOG_ENTRIES          :              162
SYS_LOG_MODE             :                1 (DISCARD)

command sent to write sys log
Event: 
CFS-1      1970/01/12 14:54:02.001 scid:66  procid:1  DEBUG CFE_ES         18: /cf/test_syslog.txt written:Size=**8217**,Entries=162


When calculating from the reported tlm size, in DISCARD mode, no FRAGMENT 13 bytes were expected, so the test here fails:

 <!> Failed to verify system Log HK tlm:
     SYS_LOG_SIZE       verified: True
         hk_sys_log_size=8192 should == CFE_PLATFORM_ES_SYSTEM_LOG_SIZE=8192
     SYS_LOG_ENTRIES    verified: True
         hk_sys_log_entries=162 should == em_sys_log_entries=162
     SYS_LOG_BYTES_USED verified: False
         hk_sys_log_bytes_used=8140 + header_size=64 = 8204 should == em_sys_log_file_size=8217

ddstewar:
Log file generated from the DISCARD mode write sys log command:
[test_syslog.txt](https://github.com/user-attachments/files/30914642/test_syslog.txt)

jphickey:
> It appears that it is possible that the FRAGMENT string gets written in rare cases when in DISCARD Mode.
> 
> If the log previously wrapped in OVERWRITE mode causing the FRAGMENT string to be written, then the mode is switched to DISCARD but the log is not cleared.
> 
> Is this expected or an anomaly? I can go either way.

Yes --- when you switch from OVERWRITE into DISCARD mode, the log **not** cleared ... so if it had already wrapped (and thus had overwritten entries) you will still get the fragment.

PR #2796, opened 2026-08-20
Fix #2574, document software bus timeout clock behavior

### LINKED ISSUE #2574: CFE_SB_ReceiveBuffer times out immediately if the system time changes on POSIX
**Describe the bug**
When pending on `CFE_SB_ReceiveBuffer` with a timeout on a POSIX system, a change to the system time causes the function to immediately return with a timeout error code or pend longer than specified depending on the direction of the time change.

**To Reproduce**
`CFE_SB_ReceiveBuffer` with a timeout of 10 seconds on a Linux system.  Change the system time ahead by 1 hour.  The function will return immediately with code CFE_SB_TIME_OUT regardless of the actual elapsed time.

**Expected behavior**
The function should always wait for the specified timeout regardless of any change to the system time.  If this is not practical, it would be helpful to document this behavior in the function doc string. 

**System observed on:**
ARM Cortex A9
Linux
Versions Draco-RC5

**Additional context**
The implementation calls OSAL function `OS_QueueGet`.  The POSIX implementation calls `mq_timedreceive` which is known to [suffer from this problem](https://e2e.ti.com/support/processors-group/processors/f/processors-forum/548380/mq_timedreceive-and-changing-the-real-time-clock). 

**Reporter Info**
Andrew Hatstat, Geost

sylvesterkaczmarek:
Fix #2574

Document the platform-specific behavior of finite `CFE_SB_ReceiveBuffer` and `CFE_SB_ReceiveBufferWithRoute` timeouts when the underlying POSIX queue uses an absolute real-time clock.

The Software Bus known-problem note now explains that wall-clock adjustments while a finite receive is pending can cause an early timeout or a longer wait, and advises applications that require elapsed-time semantics to account for this platform limitation.

Testing:
- documentation-only change
- verified the branch contains only the intended update to `docs/src/cfe_sb.dox`

Expected behavior changes:
- none; this documents existing platform behavior

Contributor: `sylvesterkaczmarek`

### DIFF
--- a/docs/src/cfe_sb.dox
+++ b/docs/src/cfe_sb.dox
@@ -391,6 +391,13 @@
 /**
   \page cfesbugknwnprob Known Problem
 
+  Finite timeouts passed to #CFE_SB_ReceiveBuffer and #CFE_SB_ReceiveBufferWithRoute
+  are implemented by the underlying OS queue. On POSIX systems that use an absolute
+  real-time clock for timed queue receives, changing the system wall clock while a
+  receive is pending may cause the call to time out early or wait longer than requested.
+  Applications requiring timeout behavior independent of wall-clock adjustments should
+  account for this platform-specific limitation.
+
   The software bus may perform unexpectedly under an unlikely corner-case scenario. This
   scenario was revealed in a stress test. The stress test was designed to deplete the Software
   Bus memory pool by having a high priority application continuously send 1000 byte packets


sylvesterkaczmarek:
This documentation change remains current and mergeable, and #2574 is still open. Could a maintainer review when convenient?

PR #2819, opened 2026-08-29
Fix #2818, CFE_SB_MessageStringSet leaves destination unterminated

### LINKED ISSUE #2818: [SECURITY] CFE_SB_MessageStringSet() can leave destination buffer without a NUL terminator

- [x] I reviewed the [[Contributing Guide](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md)](https://github.com/nasa/cFS/blob/main/CONTRIBUTING.md).
- [x] I performed a cursory search to see if the bug report is relevant, not redundant, nor in conflict with other tickets.

**Describe the bug**

`CFE_SB_MessageStringSet()` in `modules/sb/fsw/src/cfe_sb_util.c` does not always null terminate its destination buffer. The function's own header comment says it exists specifically because "neither buffer is required to be null terminated" and that it pads the destination "to ensure the entire destination buffer is set." That only holds true when `DestMaxSize` leaves enough room for the terminator after the copy. When `DestMaxSize <= SourceMaxSize` and the source has no NUL byte anywhere in the first `DestMaxSize` bytes, the copy loop consumes the whole destination buffer and the padding loop after it never runs, since `DestMaxSize` is already 0 by the time execution reaches it. The destination ends up completely filled with copied data and has no NUL anywhere inside it.

Compare this with `CFE_SB_MessageStringGet()` in the same file, which reserves one byte for the terminator before its copy loop (`--DestMaxSize;`). `Set()` has no equivalent reservation, so it doesn't actually deliver on the guarantee described in its own doc comment.

This matters because the destination gets passed around afterward as if it's a normal C string. In `fm_cmds.c`, `FM_MonitorFilesystemSpaceCmd()` copies a table entry name into a telemetry payload field using `CFE_SB_MessageStringSet()` with `sizeof(dest) == sizeof(source)`. If that copy loses its terminator, anything downstream that calls `strlen()` or formats the field with `%s` (event logging, ground telemetry decoding, etc.) can read past the end of the buffer.

**To Reproduce**

1. Declare `char dest[8];` and `char src[8];`
2. Fill `src` with 8 non-zero bytes so it contains no NUL anywhere, e.g. `memset(src, 'A', sizeof(src));`
3. Call `CFE_SB_MessageStringSet(dest, src, sizeof(dest), sizeof(src));`
4. Inspect `dest`: it now holds 8 `'A'` characters with no NUL terminator anywhere in the buffer

For a heap-based repro that catches this under AddressSanitizer instead of just inspecting the buffer by hand, see the attached `asan_repro.c`. It `#include`s `cfe_sb_messagestringset_verbatim.c`, an unmodified extraction of the current upstream function (lines 189-228 of `cfe_sb_util.c` at commit `f881799`, byte-identical, not retyped or paraphrased), calls it the same way `fm_cmds.c` does (`sizeof(dest) == sizeof(source)`, both heap allocated at exactly `CFE_MISSION_MAX_PATH_LEN`), then calls `strlen()` on the result the way any normal downstream consumer would.

**Expected behavior**

The destination should always be left NUL terminated within `DestMaxSize` bytes, even if that means truncating the copy by one character to make room, the same way `CFE_SB_MessageStringGet()` already handles it and the same way the function's own documentation describes it.

**Code snips**

Current implementation, `modules/sb/fsw/src/cfe_sb_util.c`:
```c
while (SourceMaxSize > 0 && *SourceStringPtr != 0 && DestMaxSize > 0)
{
    *DestStringPtr = *SourceStringPtr;
    ++DestStringPtr;
    ++SourceStringPtr;
    ++Result;
    --DestMaxSize;
    --SourceMaxSize;
}

/* Pad the remaining space with NUL chars, but this should NOT be included in the final size */
while (DestMaxSize > 0)
{
    *DestStringPtr = 0;
    ++DestStringPtr;
    --DestMaxSize;
}
```

Real call site that can trigger this, `apps/fm/fsw/src/fm_cmds.c` line 893:
```c
CFE_SB_MessageStringSet(ReportPtr->Path,
                        MonitorPtr->Name,
                        sizeof(ReportPtr->Path),
                        sizeof(MonitorPtr->Name));
```
Both `ReportPtr->Path` and `MonitorPtr->Name` are sized `CFE_MISSION_MAX_PATH_LEN`, so `DestMaxSize` equals `SourceMaxSize` here. FM's own table validation (`FM_ValidateTable`) does check `Name` for a terminator, but only when `CFE_TBL_VALIDATE_CC` is actually issued before the table is activated, which cFE does not require, so it's not a hard guarantee at the table load level.

The existing unit test at `modules/sb/ut-coverage/sb_UT.c` around line 5291 already exercises this exact path (`DestMaxSize=4` with a longer source), but checks the result with `UtAssert_STRINGBUF_EQ()` comparing 4 raw bytes, which doesn't require a terminator to be present, so the missing NUL was never caught there.

**System observed on:**
- Hardware: x86-64
- OS: Ubuntu 24.04 (Linux)
- Versions: cFE dev branch, commit `f881799`, cFS bundle, native_std config with `ENABLE_UNIT_TESTS=TRUE`
- Verified with AddressSanitizer

**Additional context**

Severity (self-assessed): CVSS 3.1 base 5.4 (Medium), vector `AV:N/AC:L/PR:L/UI:N/S:U/C:L/I:N/A:L`. Attack Vector is Network under the assumption that whatever gets an unterminated fixed-size string into a call site like the FM one above (a table load, typically) arrives over a network-facing ground command interface, which varies by mission and isn't guaranteed by cFE core itself. Privileges Required is Low because it needs the ability to load a table, which is normally restricted to an authorized ground role. Impact is Low/None/Low: the over-read walks forward through adjacent memory until it happens to hit a zero byte, which is primarily a confidentiality concern for whatever ends up in that gap (could be far enough to be meaningful, not necessarily just the next few bytes), with a secondary availability concern if it walks off the end of a mapped region.

Dynamic verification, output from the attached `asan_repro.c` running against the actual unmodified upstream function:
```
==540==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x506000000060
READ of size 74 at 0x506000000060 thread T0
    #0 in strlen ...
    #1 in main /asan_repro.c:82
0x506000000060 is located 0 bytes after 64-byte region [0x506000000020,0x506000000060)
allocated by thread T0 here:
    #0 in malloc ...
    #1 in main /asan_repro.c:51
SUMMARY: AddressSanitizer: heap-buffer-overflow ... in strlen
```
Full output is in the attached `asan_output.txt`. Running the identical harness against the fixed version of the function (`cfe_sb_messagestringset_fixed.c`, also attached) exits cleanly with no sanitizer error and `strlen()` correctly returning 63, see `asan_output_fixed.txt`.

I have a fix ready along with updated unit test coverage for both the truncation case and the no-NUL-at-all case. Checked against the full SB and TBL unit test suites (1025/1025 and 1358/1358 passing), `clang-format --style=file` makes no changes to the modified files, and `cppcheck --enable=warning,performance,portability` reports nothing on the changed file. I'll open a PR referencing this issue, titled "Fix #2818, CFE_SB_MessageStringSet leaves destination unterminated".

Attached:
- [cfe_sb_messagestringset_verbatim.c](https://github.com/user-attachments/files/31588031/cfe_sb_messagestringset_verbatim.c), unmodified extraction of the current vulnerable function, used as-is by the ASan harness
- [asan_repro.c](https://github.com/user-attachments/files/31588032/asan_repro.c), standalone AddressSanitizer reproduction
- [asan_output.txt](https://github.com/user-attachments/files/31588094/asan_output.txt), AddressSanitizer output from the reproduction, heap-buffer-overflow in strlen
- [cfe_sb_messagestringset_fixed.c](https://github.com/user-attachments/files/31588033/cfe_sb_messagestringset_fixed.c), the fixed function, for reference
- [asan_output_fixed.txt](https://github.com/user-attachments/files/31588034/asan_output_fixed.txt), same harness run against the fix, clean exit

**Reporter Info**
- Name: Muhammad Bilal
- Organization: Independent contributor

0nsec:
## Description of Change

`CFE_SB_MessageStringSet()` did not always leave its destination buffer NUL terminated, contrary to its own documented contract. When `DestMaxSize <= SourceMaxSize` and the source has no NUL byte within the first `DestMaxSize` bytes, the copy loop consumed the entire destination and the padding loop that follows never ran, so no terminator was ever written. See issue #2818 for the full writeup, a concrete real call site (`fm_cmds.c`), and an AddressSanitizer reproduction against the unmodified function.

The fix mirrors the existing pattern already used in the sibling function `CFE_SB_MessageStringGet()`: reserve one byte for the terminator before the copy loop runs, by changing the loop guard from `DestMaxSize > 0` to `DestMaxSize > 1`. This guarantees the padding loop that follows always writes at least one NUL byte, regardless of whether the source had one. Also added a `DestMaxSize == 0` check that returns `CFE_SB_BAD_ARGUMENT`, matching the existing behavior of `CFE_SB_MessageStringGet()` (previously `Set()` would silently no-op on a zero-size destination instead of reporting an error).

No behavior change for any source string that is already NUL terminated within `DestMaxSize` bytes, which is the common case. The only observable difference is in the truncation case: the destination is now correctly truncated to `DestMaxSize - 1` copied characters plus a terminator, instead of `DestMaxSize` copied characters with no terminator.

## Linked Issue

Closes #2818

## Requirements Impact

- Requirement ID(s): none identified, this is an internal utility function with no direct requirement mapping
- [x] Requirements updated as necessary
- [x] Existing requirements are still satisfied by this change

### Unit Tests (UT Assert)

Built and ran the full SB and TBL coverage test binaries locally (native_std config, `ENABLE_UNIT_TESTS=TRUE`), both modules exercise this function:

```
SB: 1025/1025 tests pass
TBL: 1358/1358 tests pass
```

Updated `modules/sb/ut-coverage/sb_UT.c`:
- Added a `DestMaxSize == 0` bad-argument case for `CFE_SB_MessageStringSet()`, mirroring the existing case already present for `CFE_SB_MessageStringGet()`.
- Fixed the existing truncation test (`DestMaxSize=4`, longer source) to assert the correct truncated-and-terminated result instead of 4 raw unterminated bytes, plus an explicit check that the last byte in that 4-byte allotment is NUL.
- Added a new case for a source with no NUL at all within `DestMaxSize` bytes, asserting the destination's last byte is NUL.

### AddressSanitizer

Not part of CI here, but since this is a memory-safety fix I built a standalone harness that `#include`s the unmodified upstream function verbatim and calls it the same way `fm_cmds.c` does (matching heap-allocated source/dest sizes), then calls `strlen()` on the result the way any normal caller would. Before the fix this reliably reproduces a heap-buffer-overflow read under ASan; after the fix it exits clean with the correct truncated length. Full harness and both outputs are attached to issue #2818.

### COSMOS Test Suite

N/A, no telemetry/command definitions or interfaces changed, this is an internal string-copy utility fix with no message format impact.

## Areas of Expertise Touched

- [ ] ASTRO
- [ ] CI/CD
- [ ] COSMOS
- [x] Cybersecurity
- [ ] Docker
- [ ] EDS
- [ ] Git
- [ ] PSPs
- [ ] SBN
- [ ] SMP
- [x] Tables
- [ ] TSN
- [x] Unit Tests
- [ ] Other

---

## Author Checklist

- [x] Linked GitHub issue is referenced above
- [x] Code has been formatted with `.clang-format`
- [ ] Static analysis workflows ran and passed 
- [x] Unit tests (UT Assert) updated/added to cover code changes
- [x] Unit test workflows ran and passed 
- [ ] COSMOS test suite was run; tests updated/added if relevant changes were made 
- [x] Requirements have been reviewed; updated or confirmed still satisfied (see above)
- [x] Testing evidence is included above
- [x] Self-review of the diff completed

### DIFF
--- a/modules/sb/fsw/src/cfe_sb_util.c
+++ b/modules/sb/fsw/src/cfe_sb_util.c
@@ -193,7 +193,12 @@ int32 CFE_SB_MessageStringSet(char       *DestStringPtr,
 {
     int32 Result;
 
-    if (SourceStringPtr == NULL || DestStringPtr == NULL)
+    /*
+     * Error in caller if DestMaxSize == 0.
+     * Cannot terminate the string, since there is no place for the NUL
+     * In this case, do nothing
+     */
+    if (DestMaxSize == 0 || SourceStringPtr == NULL || DestStringPtr == NULL)
     {
         Result = CFE_SB_BAD_ARGUMENT;
     }
@@ -201,7 +206,17 @@ int32 CFE_SB_MessageStringSet(char       *DestStringPtr,
     {
         Result = 0;
 
-        while (SourceMaxSize > 0 && *SourceStringPtr != 0 && DestMaxSize > 0)
+        /*
+         * Stop copying with at least 1 character of space left in the
+         * destination, so the NUL terminator below always has room.
+         * Without this, a source buffer with no NUL character within
+         * the first DestMaxSize bytes (permitted per this function's
+         * documented contract, which does not require the source to
+         * be NUL terminated) would leave DestStringPtr completely
+         * filled with non-NUL data and not NUL terminated anywhere
+         * within its bounds.
+         */
+        while (SourceMaxSize > 0 && *SourceStringPtr != 0 && DestMaxSize > 1)
         {
             *DestStringPtr = *SourceStringPtr;
             ++DestStringPtr;
@@ -213,7 +228,11 @@ int32 CFE_SB_MessageStringSet(char       *DestStringPtr,
 
         /*
          * Pad the remaining space with NUL chars,
-         * but this should NOT be included in the final size
+         * but this should NOT be included in the final size.
+         * Because the loop above always leaves at least 1 byte
+         * of destination space remaining, this unconditionally
+         * NUL terminates the destination, even when the source
+         * data had to be truncated to fit.
          */
         while (DestMaxSize > 0)
         {
--- a/modules/sb/ut-coverage/sb_UT.c
+++ b/modules/sb/ut-coverage/sb_UT.c
@@ -5267,6 +5267,7 @@ void Test_MessageString(void)
                       CFE_SB_BAD_ARGUMENT);
     UtAssert_INT32_EQ(CFE_SB_MessageStringSet(DestString, NULL, sizeof(DestString), sizeof(SrcString)),
                       CFE_SB_BAD_ARGUMENT);
+    UtAssert_INT32_EQ(CFE_SB_MessageStringSet(DestString, SrcString, 0, sizeof(SrcString)), CFE_SB_BAD_ARGUMENT);
     UtAssert_INT32_EQ(CFE_SB_MessageStringGet(NULL, SrcString, DefString, sizeof(DestString), sizeof(SrcString)),
                       CFE_SB_BAD_ARGUMENT);
     UtAssert_INT32_EQ(CFE_SB_MessageStringGet(DestString, SrcString, DefString, 0, sizeof(SrcString)),
@@ -5289,7 +5290,17 @@ void Test_MessageString(void)
     strcpy(SrcString, "abcdefg");
     memset(DestString, 'q', sizeof(DestString));
     CFE_SB_MessageStringSet(DestString, SrcString, 4, sizeof(SrcString));
-    UtAssert_STRINGBUF_EQ(DestString, 4, SrcString, 4);
+    UtAssert_STRINGBUF_EQ(DestString, sizeof(DestString), "abc", UTASSERT_STRINGBUF_NULL_TERM);
+    /* Confirm the destination is left NUL terminated within its 4 byte
+     * allotment, even though the source string had to be truncated to fit */
+    UtAssert_UINT8_EQ(DestString[3], 0);
+
+    /* Test setting string where the source has no NUL character within
+     * the destination size (source is not required to be NUL terminated) */
+    memset(SrcString, 'z', sizeof(SrcString));
+    memset(DestString, 'q', sizeof(DestString));
+    CFE_SB_MessageStringSet(DestString, SrcString, sizeof(DestString), sizeof(SrcString));
+    UtAssert_UINT8_EQ(DestString[sizeof(DestString) - 1], 0);
 
     /* Test getting string where the destination size > source string size */
     strcpy(SrcString, "abcdefg");


dzbaker:
Tagging @ddstewar for awareness.

jphickey (changes requested):
Recommend not merging this.  The original behavior is intended and this was working as designed.

Strings inside fixed-length buffers within messages are not required to be null terminated.

PR #2820, opened 2026-08-29
fix(evs): allow SendEventWithAppID for caller own AppId during startup

### LINKED ISSUE #2663: Unregistered AppId error with CFE_EVS_SendEventWithAppID() during cFE task startup
**Describe the bug**
When the ES and the EVS tasks initializes, they make various CFE_SB API calls (i.e. `CFE_SB_CreatePipe()` and `CFE_SB_Subscribe()`) which in turn call `CFE_EVS_SendEventWithAppID()` passing in the AppId for the CFE_SB.  `CFE_EVS_SendEventWithAppID()` then returns a `CFE_EVS_APP_ILLEGAL_APP_ID` presumably because CFE_SB has not been registered with event services.  This might have gone unnoticed because there is no event outputted nor is the return code set to any variable.  

**To Reproduce**
1. make SIMULATION=native prep
2. make
3. make install
4. Place a breakpoint here - https://github.com/nasa/cFE/blob/main/modules/evs/fsw/src/cfe_evs.c#L185
5. Run the gdb and see it stopped at the breakpoint.  

**System observed on:**
 - OS: Ubuntu 22.04.5 LTS
 - cFE v6.7

**Reporter Info**
Michael Yang - NASA/GSFC

philphauler:
* [x] I reviewed the [Contributing Guide](https://github.com/nasa/cFE/blob/main/CONTRIBUTING.md)
* [ ] I signed and emailed the CLA

**Describe the contribution**

Fixes #2663. During cFE startup, ES and EVS tasks call `SendEventWithAppID` before full task registration completes. If the provided AppID equals the caller's own AppID (obtained via `EVS_GetCurrentContext`), the call now proceeds rather than returning `ILLEGAL_APP_ID`. Prevents startup event loss in the normal startup sequence.

**Testing performed**

Verified against cFE dev branch HEAD. The affected startup path was traced through `cfe_evs_task.c` and `cfe_es_task.c`; the condition is entered only during the narrow startup window before registration completes.

**Expected behavior changes**

Startup events that were silently dropped with `ILLEGAL_APP_ID` during the registration window are now correctly emitted.

**System(s) tested on**

Hardware: PC, OS: Windows 11, Versions: cFE dev branch HEAD

**Contributor Info**

Filip Koscak - phil.phaulre@gmail.com

### DIFF
--- a/modules/evs/fsw/src/cfe_evs.c
+++ b/modules/evs/fsw/src/cfe_evs.c
@@ -185,22 +185,18 @@ CFE_Status_t CFE_EVS_SendEventWithAppID(uint16                   EventID,
     AppDataPtr = EVS_GetAppDataByID(AppID);
     if (AppDataPtr == NULL)
     {
-        /*
-         * Check if the AppId is the caller's own AppId (e.g. during cFE startup
-         * when ES/EVS tasks initialize before full registration). In this case,
-         * allow the event to proceed rather than returning ILLEGAL_APP_ID.
-         */
         CFE_ES_AppId_t callerAppId;
         EVS_AppData_t *callerAppData;
-        int32 status = EVS_GetCurrentContext(&callerAppData, &callerAppId);
-        if (status == CFE_SUCCESS && callerAppData != NULL && CFE_RESOURCEID_TEST_EQUAL\(callerAppId, AppID\))
+        if (EVS_GetCurrentContext(&callerAppData, &callerAppId) == CFE_SUCCESS &&
+            callerAppData != NULL && CFE_RESOURCEID_TEST_EQUAL(callerAppId, AppID))
         {
-            Status = CFE_SUCCESS;
+            AppDataPtr = callerAppData;
         }
-        else
-        {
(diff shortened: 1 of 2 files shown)


dzbaker:
@philphauler Thank you for your contribution. Please resolve the workflow failures.

sylvesterkaczmarek (commented):
There seem to be two issues in the new startup branch. The current source contains literal backslashes in `CFE_RESOURCEID_TEST_EQUAL\(callerAppId, AppID\)`, which will not compile. More importantly, on a caller-ID match it sets `Status = CFE_SUCCESS` but never routes `callerAppData` through the filtering/generation path, so the API reports success without emitting the event. Could the caller context be fed through the normal event path instead?

philphauler:
Onnit will fix 🫡🙌

philphauler:
Closing this one: the fix is vacuous (see my note on #2663 with the actual cause). Thanks @sylvesterkaczmarek for the review, both points were right. Have a nice day!

Issue #2317, opened 2023-05-01
wrong closing parenthesis in CFE_MSG_CMD_HDR_INIT macro


mgbark23:
The CFE_MSG_CMD_HDR_INIT macro's parentheses are not properly balanced.

The CFE_MSG_CMD_HDR_INIT macro contains a ')' instead of a '}' in line 70 as of Latest commit [f9112b4] on Jun 7, 2022
nasa/cFE/modules/msg/option_inc/default_cfe_msg_hdr_priext.h

Simple inspection shows the problem.

Opening and closing parentheses should be balanced.

line 70:  .SystemId = {0, 0) \

Matthias Bark, Boeing

thnkslprpt:
There is also an unbalanced (extra) round bracket/parenthesis:
![Screenshot 2023-05-02 11 07 12](https://user-images.githubusercontent.com/9024662/235558896-c4f483f5-6a90-4121-9bf8-572fdb8ee1e3.png)

PR #2821, opened 2026-08-29
Fix #2745, validate CFE_MSG_Init size and use OR for byte concatenation

### LINKED ISSUE #2745: [Security] cfe/modules/msg — 4 novel findings (HIGH: memset OOB write + broken priext macro; MEDIUM: checksum OOB read driven by wire-controlled length)
## Summary

Four findings in `cfe/modules/msg` — the CCSDS message framing layer used by every cFS application for command and telemetry messages. Two are HIGH (one out-of-bounds write, one mission-build-config-dependent silently-broken header initializer), one MEDIUM (out-of-bounds READ in checksum walk driven by attacker-supplied wire length), one LOW (fragile arithmetic that is benign today but breaks under a foreseeable refactor).

The findings were produced by an LLM-driven source-aware audit of the same files KCode's deterministic pattern library considers clean (KCode skip-verify on `cfe/modules/msg` returned **0 candidates**). None of the four are matchable by regex / AST / structural pattern rules — they require reading the function in context, tracing argument flow from caller, or recognizing a syntactic error inside a macro that no compiler warning catches because the macro itself is not currently exercised in the default build profile.

## Severity

**HIGH** for findings #1 and #2; **MEDIUM** for finding #3; **LOW** for finding #4.

For finding #1, exploitability requires only that some caller of `CFE_MSG_Init` receive an attacker-influenced `Size` argument — a pattern that is easy to introduce in mission-specific applications (table-driven message sizing, command-driven allocations).

For finding #2, exploitability is gated by the mission's choice of header configuration (`MSG_HDR_FILE`). If the priext configuration is ever selected, the table commands declared with `CFE_MSG_CMD_HDR_INIT` may not compile, or — worse — silently emit incorrect headers depending on how the C compiler tolerates the malformed initializer.

For finding #3, exploitability requires an upstream caller that hands an unvalidated wire packet to `CFE_MSG_ValidateChecksum`. The function's comment ("Message already checked, no error case reachable") is itself the defect — the routine inherits trust from its caller without verifying it.

## Affected component(s)

- `cfe/modules/msg/fsw/src/cfe_msg_init.c:36` — `CFE_MSG_Init`
- `cfe/modules/msg/option_inc/default_cfe_msg_hdr_priext.h:57-73` — `CFE_MSG_CMD_HDR_INIT` macro (priext build configuration)
- `cfe/modules/msg/fsw/src/cfe_msg_sechdr_checksum.c:36-50` — `CFE_MSG_ComputeCheckSum`
- `cfe/modules/msg/fsw/src/cfe_msg_ccsdsext.c:224` — `CFE_MSG_GetSystem` (also: `cfe_msg_ccsdspri.c` `CFE_MSG_GetSize`)

### #1 — `cfe/modules/msg/fsw/src/cfe_msg_init.c:36`  Out-of-bounds memset on attacker-influenced `Size` (HIGH, CWE-787)

```c
CFE_Status_t CFE_MSG_Init(CFE_MSG_Message_t *MsgPtr, CFE_SB_MsgId_t MsgId, CFE_MSG_Size_t Size)
{
    int32 status;

    if (MsgPtr == NULL)
    {
        return CFE_MSG_BAD_ARGUMENT;
    }

    /* Clear and set defaults */
    memset(MsgPtr, 0, Size);                /* <-- (1) executes before Size validation */
    CFE_MSG_InitDefaultHdr(MsgPtr);

    /* Set values input */
    /* Allow "CFE_SB_INVALID_MSG_ID" here - in some cases the real msgid is filled in later */
    /* however the size should always be a valid value */
    status = CFE_MSG_SetSize(MsgPtr, Size);  /* <-- (2) Size enforced here, too late */
    if (status != CFE_SUCCESS)
    {
        return status;
    }
    /* ... */
}
```

`Size` is `CFE_MSG_Size_t` (an unsigned type wide enough to express the CCSDS Length field). At point (1) the function has not yet checked that `Size` is within the buffer pointed to by `MsgPtr`. At point (2) `CFE_MSG_SetSize` enforces a maximum of `CFE_MSG_MAX_MESSAGE_SIZE` (= `0xFFFF + 7` = 65 542 bytes) — but by then the `memset` has already executed using the caller's unvalidated value.

**Exploit path.** Any caller that derives `Size` from attacker-influenced input — a table load entry, a command-driven allocation length, a forwarded-from-network value — and passes it into `CFE_MSG_Init` without first clamping against the actual buffer size triggers an out-of-bounds write into adjacent memory. The `memset` zeroes the writes (so this is not a direct content-injection primitive), but the OOB region is fully under the attacker's control via `Size`, and the cleared region likely contains structured data (other components' state, function pointers in BSS / data layout) that is consequential when zeroed.

**Fix.**

```c
if (Size > CFE_MSG_MAX_MESSAGE_SIZE)
{
    return CFE_MSG_BAD_ARGUMENT;
}
/* Clear and set defaults */
memset(MsgPtr, 0, Size);
```

The same maximum is enforced one call down inside `CFE_MSG_SetSize`; hoisting the check above the `memset` resolves the bug.

### #2 — `cfe/modules/msg/option_inc/default_cfe_msg_hdr_priext.h:57-73`  Broken `CFE_MSG_CMD_HDR_INIT` macro under priext build configuration (HIGH, CWE-665 / CWE-682)

```c
#define CFE_MSG_CMD_HDR_INIT(mid, size, fc, cksum)      \
    {                                                   \
    .Msg.CCSDS = \
    { \
        .Pri = \
        { \
            .StreamId = {(((mid)&0x80)<<5, (mid)&0x7F}, \  /* (a) parses as initializer */
            .Sequence = {0xC0, 0}, \
            .Length = {((size)-7)>>8, ((size)-7)&0xFF} \
        }, \
        .Ext = \
        { \
            .Subsystem = {0, ((mid)>>8)&0xFF}, \
            .SystemId = {0, 0) \                           /* (b) ← '{' closed with ')' */
        } \
    }, \
    CFE_MSG_CMD_HDR_SEC_INIT(fc, cksum) \
}
```

The `.SystemId = {0, 0)` initializer (line marked (b)) opens with `{` and closes with `)`. This is not legal C — array initializers must close with `}`. A compiler will either reject the expansion, or — depending on which preceding `(` it pairs the rogue `)` with — silently mis-parse the entire `CFE_MSG_CMD_HDR_INIT` invocation as a different expression, emitting a header whose fields are assigned to different members than the macro author intended.

`mission_build.cmake:13` sets `MSG_HDR_FILE "default_cfe_msg_hdr_priext.h"` for the priext build profile. `CFE_MSG_CMD_HDR_INIT` is referenced from real command tables — `apps/hs/fsw/tables/hs_mat.c:71,78,82,89,96,103,110,118`, `apps/sc/fsw/tables/sc_ats*.c`, and others. Under the priext header profile these tables either fail to build, or produce tables with incorrect CCSDS StreamId / SystemId fields silently in the bytes the flight image ships.

**Exploit path.** Two failure modes:

1. **Build-break footgun.** A mission integrator selecting priext for an extended-header build (e.g. for fields such as `Subsystem` / `SystemId`) discovers the build fails. Time-cost only.
2. **Silent mis-initialization.** If the compiler tolerates the malformed initializer by mis-pairing the stray `)` with a preceding `(`, the produced table commands carry wrong CCSDS field values. On the flight image, commands run with the wrong opcode bytes; on the ground, telemetry matching against opcode-by-table-position will silently disagree with the actual on-wire values. This is the harder failure to debug — the symptom is "commands work in pri builds, mysteriously misbehave in priext builds".

**Fix.** Replace the trailing `)` with `}`:

```diff
-            .SystemId = {0, 0) \
+            .SystemId = {0, 0} \
```

While in the macro, also clean up the `.StreamId` initializer if a comma-operator-style initialization was intended (the current form `{((mid)&0x80)<<5, (mid)&0x7F}` is valid C — a 2-element array initialized to those two byte values — but worth double-checking it was the author's intent rather than a similar typo from the same change).

### #3 — `cfe/modules/msg/fsw/src/cfe_msg_sechdr_checksum.c:36-50`  Out-of-bounds READ in checksum walk driven by wire-controlled length (MEDIUM, CWE-125)

```c
CFE_MSG_Checksum_t CFE_MSG_ComputeCheckSum(const CFE_MSG_Message_t *MsgPtr)
{
    CFE_MSG_Size_t     PktLen  = 0;
    const uint8       *BytePtr = (const uint8 *)MsgPtr;
    CFE_MSG_Checksum_t chksum  = 0xFF;

    /* Message already checked, no error case reachable */
    CFE_MSG_GetSize(MsgPtr, &PktLen);

    while (PktLen--)
    {
        chksum ^= *(BytePtr++);
    }

    return chksum;
}
```

`CFE_MSG_GetSize` reads the CCSDS Length field from the message header (bytes 4-5 of the primary header) and returns `Length + CFE_MSG_SIZE_OFFSET` — a value taken straight from the wire. The function then walks PktLen bytes from MsgPtr, XORing each into the checksum.

There is no comparison between PktLen and the actual buffer size the caller is holding. The comment "Message already checked, no error case reachable" expresses an invariant that the routine does NOT enforce — it inherits it.

**Exploit path.** Any flow that delivers a wire packet to a CI subsystem which then calls `CFE_MSG_ValidateChecksum` (which in turn calls `CFE_MSG_ComputeCheckSum`) on the received buffer is at risk. An attacker sets the CCSDS Length field to declare a packet much larger than the actual receive buffer. The XOR loop walks past the buffer end, reading adjacent memory. The XOR itself is the result, so this is an information-disclosure side channel (via the checksum byte returned to the dispatcher and possibly logged) and a potential trigger for downstream OOB reads when the dispatcher treats the buffer as "PktLen bytes long".

**Fix.** Add an explicit upper bound parameter and check it inside the function:

```c
CFE_MSG_Checksum_t CFE_MSG_ComputeCheckSum(const CFE_MSG_Message_t *MsgPtr, CFE_MSG_Size_t BufLen)
{
    CFE_MSG_Size_t PktLen = 0;
    CFE_MSG_GetSize(MsgPtr, &PktLen);
    if (PktLen > BufLen) PktLen = BufLen;        /* refuse to walk past caller's buffer */
    /* ... */
}
```

A non-breaking alternative is to clamp inside `CFE_MSG_GetSize` against `CFE_MSG_MAX_MESSAGE_SIZE` and require all CFE_MSG callers to provide a buffer at least that large. The breaking-but-correct fix is the explicit `BufLen` argument above.

### #4 — `cfe/modules/msg/fsw/src/cfe_msg_ccsdsext.c:224`  Use of `+` instead of `|` in byte concatenation (LOW, CWE-682)

```c
*System = (MsgPtr->CCSDS.Ext.SystemId[0] << 8) + MsgPtr->CCSDS.Ext.SystemId[1];

/* identical shape in cfe_msg_ccsdspri.c CFE_MSG_GetSize: */
*Size = (MsgPtr->CCSDS.Pri.Length[0] << 8) + MsgPtr->CCSDS.Pri.Length[1] + CFE_MSG_SIZE_OFFSET;
```

`SystemId[0]` and `SystemId[1]` are `uint8`. Integer promotion makes them `int` for the arithmetic. The shifted value `0x0000`–`0xFF00` plus the low byte `0x00`–`0xFF` cannot exceed `0xFFFF`, so today the `+` result equals the `|` result. The code is functionally correct.

The defect is the use of `+` to express what is structurally a bit concatenation. If any future patch widens the byte storage type (e.g. for an extended-Length variant of the CCSDS header), or hand-inlines this calculation for performance and the inlined version sign-extends, the `+` form becomes a real bug while `|` would still be correct.

**Fix.** Spelled correctly:

```c
*System = ((uint16)MsgPtr->CCSDS.Ext.SystemId[0] << 8) | MsgPtr->CCSDS.Ext.SystemId[1];
```

This is a hygiene fix; no immediate security impact in the current code.

## Reproduction notes

Findings #1, #3 are reproducible by reading the cited functions in their current form on the `dev` branch as of commit 4a1c553. Finding #4 is the same kind of read-the-line fact.

Finding #2 requires building the project with the priext header profile selected. To reproduce:

```sh
git clone https://github.com/nasa/cFE.git
cd cFE
cmake -B build -DMSG_HDR_FILE=default_cfe_msg_hdr_priext.h
make -C build apps/hs/CMakeFiles/hs.dir/fsw/tables/hs_mat.c.o
```

The build fails (or, on a tolerant compiler, emits a warning and produces silently incorrect table contents).

## Provenance

Discovered by **Inquisitor**, AstroLexis's autonomous source + binary security agent. The audit pipeline:

1. **KCode** (https://github.com/AstrolexisAI/KCode) — deterministic SAST pattern library. `kcode audit cfe/modules/msg --skip-verify --json` returned **0 candidates** over 19 files. Pattern coverage: 100% of the rules in the library, none matched the four defects above.
2. **Inquisitor SourceHunter / novel-pattern-hunt** — LLM (Claude Opus 4.7) reading the same source bundle, instructed to look for vulnerability classes NOT covered by the deterministic library (cross-function flow, TOCTOU, integer truncation, silently-broken macros, header-derived-length OOB). Returned the four findings above with file:line anchors and quoted proof for each.

Per nasa/cFE SECURITY.md AI policy: **discovery and report drafting used AI assistance.** Discovery: Inquisitor (KCode static pass + LLM novel-pattern pass). Report drafted with AI assistance. All four findings are anchored in concrete source and quoted verbatim; finding #2 is contingent on the priext build configuration being selected by the integrating mission.

Best regards,

**Bruno Aiub** · AstroLexis · Inquisitor
contact@astrolexis.space

philphauler:
* [x] I reviewed the [Contributing Guide](https://github.com/nasa/cFE/blob/main/CONTRIBUTING.md)
* [ ] I signed and emailed the CLA

**Describe the contribution**

Fixes #2745, findings 1 and 4 only (scope narrowed per review):

- `CFE_MSG_Init`: the caller-supplied `Size` is now validated before the `memset`, with the same bound `CFE_MSG_SetSize` uses (`CFE_MSG_SIZE_OFFSET <= Size <= 0xFFFF + CFE_MSG_SIZE_OFFSET`); `CFE_MSG_SIZE_OFFSET` moved to `cfe_msg_priv.h` so both share it.
- `GetSystem`/`GetSize`: `|` instead of `+` for byte concatenation.

Finding 2 (priext initializer) was already fixed upstream in 13ed881 (#2317). Finding 3 (checksum walk) needs a decision on the BufLen API and is left out of this PR.

**Testing performed**

`test_cfe_msg_init.c` adds an oversize case (a 65 KB buffer, `Size` one over the bound) and an underflow case (`Size` below `CFE_MSG_SIZE_OFFSET`), both expecting `CFE_MSG_BAD_ARGUMENT`, plus the exact-bound case. Run locally in the cFS dev bundle with the tests kept and `cfe_msg_init.c` reverted to dev: coverage-msg-ALL dies with a SegFault in Test_MSG_Init (the oversize memset runs past the message). With the fix, coverage-msg-ALL passes.

**Expected behavior changes**

Malformed message inputs that previously caused OOB reads/writes now return an error. Bit-field concatenation in `GetSystem`/`GetSize` is now correct for fields with overlapping bits.

**System(s) tested on**

Hardware: x86_64 (WSL2), OS: Ubuntu 22.04, Versions: cFE dev 2612eb05 with OSAL and PSP from the cFS dev bundle, native build with ENABLE_UNIT_TESTS

**Contributor Info**

Filip Koscak - phil.phaulre@gmail.com

### DIFF
--- a/modules/msg/fsw/src/cfe_msg_ccsdsext.c
+++ b/modules/msg/fsw/src/cfe_msg_ccsdsext.c
@@ -251,7 +251,7 @@ CFE_Status_t CFE_MSG_GetSystem(const CFE_MSG_Message_t *MsgPtr, CFE_MSG_System_t
         return CFE_MSG_BAD_ARGUMENT;
     }
 
-    *System = (MsgPtr->CCSDS.Ext.SystemId[0] << 8) + MsgPtr->CCSDS.Ext.SystemId[1];
+    *System = ((uint16)MsgPtr->CCSDS.Ext.SystemId[0] << 8) | MsgPtr->CCSDS.Ext.SystemId[1];
 
     return CFE_SUCCESS;
 }
--- a/modules/msg/fsw/src/cfe_msg_ccsdspri.c
+++ b/modules/msg/fsw/src/cfe_msg_ccsdspri.c
@@ -26,8 +26,7 @@
 
 #include <assert.h>
 
-/* CCSDS Primary Standard definitions */
-#define CFE_MSG_SIZE_OFFSET    7      /**< \brief CCSDS size offset */
+/* CCSDS Primary Standard definitions (CFE_MSG_SIZE_OFFSET is in cfe_msg_priv.h) */
 #define CFE_MSG_CCSDSVER_MASK  0xE000 /**< \brief CCSDS version mask */
 #define CFE_MSG_CCSDSVER_SHIFT 13     /**< \brief CCSDS version shift */
 #define CFE_MSG_TYPE_MASK      0x1000 /**< \brief CCSDS type mask, command when set */
@@ -378,7 +377,7 @@ CFE_Status_t CFE_MSG_GetSize(const CFE_MSG_Message_t *MsgPtr, CFE_MSG_Size_t *Si
         return CFE_MSG_BAD_ARGUMENT;
     }
 
-    *Size = (MsgPtr->CCSDS.Pri.Length[0] << 8) + MsgPtr->CCSDS.Pri.Length[1] + CFE_MSG_SIZE_OFFSET;
+    *Size = (((uint16)MsgPtr->CCSDS.Pri.Length[0] << 8) | MsgPtr->CCSDS.Pri.Length[1]) + CFE_MSG_SIZE_OFFSET;
 
     return CFE_SUCCESS;
 }
--- a/modules/msg/fsw/src/cfe_msg_init.c
+++ b/modules/msg/fsw/src/cfe_msg_init.c
@@ -41,6 +41,13 @@ CFE_Status_t CFE_MSG_Init(CFE_MSG_Message_t *MsgPtr, CFE_SB_MsgId_t MsgId, CFE_M
         return CFE_MSG_BAD_ARGUMENT;
     }
 
+    /* Reject an invalid size before it is used as the memset length. CFE_MSG_SetSize
+     * enforces the same bounds, but only after the buffer has already been cleared */
+    if (Size < CFE_MSG_SIZE_OFFSET || Size > (0xFFFF + CFE_MSG_SIZE_OFFSET))
+    {
+        return CFE_MSG_BAD_ARGUMENT;
+    }
+
     /* Clear and set defaults */
     memset(MsgPtr, 0, Size);
     CFE_MSG_InitDefaultHdr(MsgPtr);
--- a/modules/msg/fsw/src/cfe_msg_priv.h
+++ b/modules/msg/fsw/src/cfe_msg_priv.h
@@ -32,6 +32,9 @@
 #include "common_types.h"
 #include "cfe_msg_hdr.h"
 
+/* CCSDS size offset, shared by the size accessors and CFE_MSG_Init */
+#define CFE_MSG_SIZE_OFFSET 7 /**< \brief CCSDS size offset */
+
 /*---------------------------------------------------------------------------------------*/
 /**
  * \brief get generic header field (uint8 array[2])
--- a/modules/msg/ut-coverage/test_cfe_msg_init.c
+++ b/modules/msg/ut-coverage/test_cfe_msg_init.c
@@ -31,6 +31,7 @@
 #include "test_cfe_msg_init.h"
 #include "cfe_error.h"
 #include "cfe_msg_defaults.h"
+#include "cfe_msg_priv.h"
 #include <string.h>
 
 #define TEST_DEFAULT_APID_MASK 0x780 /* Bits that can be set by default apid if msgid V2 */
@@ -40,6 +41,12 @@
  */
 void Test_MSG_Init(void)
 {
+    /* Sized to the largest length CFE_MSG_SetSize accepts; static to keep it off the stack */
+    static union
+    {
+        CFE_MSG_Message_t Msg;
+        uint8             Bytes[0xFFFF + CFE_MSG_SIZE_OFFSET];
+    } maxmsg;
     CFE_MSG_CommandHeader_t    cmd;
     CFE_MSG_Size_t             size;
     CFE_SB_MsgId_Atom_t        msgidval_exp;
@@ -57,6 +64,15 @@ void Test_MSG_Init(void)
     UT_SetDefaultReturnValue(UT_KEY(CFE_SB_IsValidMsgId), true);
     UtAssert_INT32_EQ(CFE_MSG_Init(NULL, msgid_act, sizeof(cmd)), CFE_MSG_BAD_ARGUMENT);
     UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgid_act, 0), CFE_MSG_BAD_ARGUMENT);
+    UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgid_act, CFE_MSG_SIZE_OFFSET - 1), CFE_MSG_BAD_ARGUMENT);
+
+    /* An oversize length must be rejected before it is used to clear the buffer: cmd is only a
+     * command header, so clearing 0xFFFF + CFE_MSG_SIZE_OFFSET + 1 bytes would run past it */
+    UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgid_act, 0xFFFF + CFE_MSG_SIZE_OFFSET + 1),
+                      CFE_MSG_BAD_ARGUMENT);
+
+    /* The largest representable size is still accepted */
+    CFE_UtAssert_SUCCESS(CFE_MSG_Init(&maxmsg.Msg, msgid_act, sizeof(maxmsg)));
     UT_SetDefaultReturnValue(UT_KEY(CFE_SB_IsValidMsgId), false);
     UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), msgid_act, sizeof(cmd)), CFE_MSG_BAD_ARGUMENT);
     UtAssert_INT32_EQ(CFE_MSG_Init(CFE_MSG_PTR(cmd), CFE_SB_INVALID_MSG_ID, sizeof(cmd)), CFE_SUCCESS);


dzbaker:
@philphauler Thank you for your contribution. Please resolve the workflow failures.

sylvesterkaczmarek (commented):
The PR is scoped as fixing all four findings from #2745, but the current head still leaves two of them untouched: the malformed `CFE_MSG_CMD_HDR_INIT` initializer in `default_cfe_msg_hdr_priext.h`, and the checksum walk in `cfe_msg_sechdr_checksum.c` (that file only has BOM/whitespace changes here). Could those two fixes be included, or should the PR scope/title be narrowed to the findings actually addressed?

philphauler:
Imma look more into it 🫡⚡️

philphauler:
Hi @sylvesterkaczmarek 👋 good call on the scope. Narrowed to findings 1 and 4 (4e2c757d, squashed to one commit): finding 2 was already fixed upstream in 13ed881 (#2317) and finding 3 needs a decision on the BufLen API that does not belong in this PR. The size bound now mirrors CFE_MSG_SetSize (CFE_MSG_SIZE_OFFSET moved to cfe_msg_priv.h) and the UT adds the 65 KB-over-a-header and underflow cases. coverage-msg-ALL passes. Have a great day!

sylvesterkaczmarek (approved):
The scope is now clear and matches the implementation: findings 1 and 4 are addressed here, with the already-upstream and API-design items left out. The shared size offset and boundary tests make the `CFE_MSG_Init` fix consistent with `CFE_MSG_SetSize`, and the byte concatenation uses the intended unsigned shift/OR. My scope concern is resolved.

philphauler:
Eyyyy let's gooooo ⚡️🚀

philphauler:
Hey @dzbaker 👋 following up on the workflow failures: they were on the earlier head (5e83079a), where `CFE_MSG_Init` rejected anything above `CFE_MISSION_SB_MAX_SB_MSG_SIZE`. The functional test's `CFE_FT_BigMsg` is that max plus the SB header, so Init returned `BAD_ARGUMENT` before `CFE_SB_TransmitBuffer` could say `MSG_TOO_BIG`.

The current head (4e2c757d, the one @sylvesterkaczmarek approved) bounds Init by the CCSDS length field instead (`CFE_MSG_SIZE_OFFSET` up to `0xFFFF + CFE_MSG_SIZE_OFFSET`), leaving the SB buffer limit to TransmitBuffer as the test expects. Ran the functional suite on this head locally the way the workflow does (cFS dev bundle, `core-cpu1` + `cfe_testcase`): `SUMMARY TOTAL::2242 PASS::1867 FAIL::0`, and `sb_sendrecv_test.c:446-453` pass. Coverage for the msg module is 618/618 with `cfe_msg_init.c` at 100% lines and branches, and the five changed files are clean under the cFS dev `.clang-format`.

The nine workflows for 4e2c757d have been sitting in "approve and run" since Sept 5, so the checks shown are still the old head's. Could you kick them off when you get a sec? 🙏⚡


Edit: two more commits landed after this note (18b701dd, then its revert 0c3790cc), the tree at the head is identical to 4e2c757d. The intermediate commit came from a background job of mine that dropped the CCSDS upper bound in Init; I reverted it because that reintroduces the oversize memset before SetSize can reject the length. Sorry for the noise, nothing changed against what was approved.

sylvesterkaczmarek (approved):
Rechecked current `0c3790cc`. Its tree is identical to the previously approved `4e2c757d` (`fe1ebb26579b` in both cases); the intervening change was fully reverted. The scoped #2745 fixes and boundary coverage are therefore unchanged. No blocker from me.

philphauler:
Heyy @sylvesterkaczmarek im actually working here on improving my agentic setup, Lestarch just recently setup a really cool bot himself for checking all the issues in code, in order to offload and mitigate bugs and mistakes, so I got inspired and am re-reviewing my work ✨🙌

philphauler:
Quick update: my Individual CLA for the cFS framework went to GSFC-SoftwareRelease on 2026-09-16, after this PR was opened, so I've ticked that box in the checklist. Happy to adjust anything if needed.

